PR #9, opened 2022-06-07
fix: set a proper image name for the controller

davidmogar:
Without this, infra-deployments will always use controller:latest.


dirgim (approved):
LGTM

PR #10, opened 2022-06-07
feat[HACBS-514]: Add RBAC definitions for managed CRs

dirgim:
Add the required RBAC definitions for all Custom Resource which the Integration controller interacts with.

### DIFF
--- a/config/rbac/application_role_binding.yaml
+++ b/config/rbac/application_role_binding.yaml
@@ -1,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: application-role-binding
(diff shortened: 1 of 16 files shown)


config/rbac/tekton_role.yaml
```diff
@@ -0,0 +1,51 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: tekton-role
```
srivickynesh, line 5: is it okay to change the name to something like tekton-cluster-admin-role like other fies

srivickynesh (approved):
LGTM

jsztuka (approved):
LGTM

PR #22, opened 2022-08-04
fix(HACBS-966): Loop over found bindings to filter based on application

dirgim:
Currently the [findExistingApplicationSnapshotEnvironmentBinding](https://github.com/redhat-appstudio/integration-service/blob/main/controllers/integration/integration_adapter.go#L418) function uses mutliple client.MatchingFields calls in order to find bindings that reference the specified environment as well as the application.

Unfortunately, due to the limitation of the FieldSelector filtering when used with cache-based implementations, this approach does not work correctly, and only the `spec.environment` is taken into account. This change does the following:

* findExistingApplicationSnapshotEnvironmentBinding refactored to use a single MatchingFields
* Use a for loop to filter based on application
* Fix related warnings raised by the IDE

### DIFF
--- a/controllers/integration/integration_adapter.go
+++ b/controllers/integration/integration_adapter.go
@@ -419,7 +419,6 @@ func (a *Adapter) findExistingApplicationSnapshotEnvironmentBinding(environmentN
 	applicationSnapshotEnvironmentBindingList := &appstudioshared.ApplicationSnapshotEnvironmentBindingList{}
 	opts := []client.ListOption{
 		client.InNamespace(a.application.Namespace),
-		client.MatchingFields{"spec.application": a.application.Name},
 		client.MatchingFields{"spec.environment": environmentName},
 	}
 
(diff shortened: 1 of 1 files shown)


johnbieren:
lgtm!

Dannyb48 (approved):
LGTM

controllers/integration/integration_adapter.go
```diff
@@ -489,7 +494,7 @@ func (a *Adapter) EnsureApplicationSnapshotEnvironmentBindingExist() (results.Op
 }
 
 func (a *Adapter) getApplicationSnapshotComponents(components []hasv1alpha1.Component) *[]appstudioshared.BindingComponent {
-	bindingComponents := []appstudioshared.BindingComponent{}
+	var bindingComponents []appstudioshared.BindingComponent
```
jinqi7, line 497: Can you please make clear - any problem with the origin way?  Or, it's recommended to use the changed way in some condition? Thanks

dirgim: I applied the suggested approach after I got a warning from the IDE about `Empty slice declaration using a literal`. After some digging, I've found that this warning can be ignored ([jetbrains ticket here](https://youtrack.jetbrains.com/issue/GO-5317)), so I've decided to revert these changes to use your original approach.

jinqi7: Thanks for your answer!

dirgim: @jinqi7 I reverted the mentioned changes, please take a look

jinqi7: OK.

jinqi7:
LGTM

jinqi7 (approved):

hongweiliu17 (approved):
LGTM

PR #23, opened 2022-08-16
feat[HACBS-806]:Add unit test for tekton functions

jinqi7:

### DIFF
--- a/tekton/predicates_test.go
+++ b/tekton/predicates_test.go
@@ -1,0 +1,145 @@
+package tekton_test
+
+import (
+	"fmt"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	tekton "github.com/redhat-appstudio/integration-service/tekton"
+	tektonv1beta1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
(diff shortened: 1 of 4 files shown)


dirgim (approved):
LGTM

tekton/predicates_test.go
```diff
@@ -0,0 +1,145 @@
...
+				Params: []tektonv1beta1.Param{
+					{
+						Name: "output-image",
+						Value: tektonv1beta1.ArrayOrString{
+							StringVal: "test-image",
+						},
+					},
+				},
+			},
+			Status: tektonv1beta1.PipelineRunStatus{
```
davidmogar, line 46: Maybe I'm wrong, but this should have no effect at all. You cannot set statuses this way.

dirgim: This got me thinking and I tried it out locally - the status does set in the `PipelineRun` object (hence why the unit tests are passing), but it doesn't actually get applied if we wanted to create that resource on the cluster.

davidmogar: My bad. You are absolutely right. Most of our tests create objects so I didn't realize you don't do it here.

Dannyb48 (approved):
LGTM

PR #27, opened 2022-08-22
feat(HACBS-1007): replace ReleaseLink with ReleasePlan

davidmogar:
The release service has replaced ReleaseLinks with ReleasePlans. This change updates every single reference to the old CRD.

**NOTE:** This change depends on https://github.com/redhat-appstudio/release-service/pull/85 so the new CRD is available.


johnbieren (approved):
lgtm. I didn't test it, but it is a 1 for 1 replacement so should be straightforward

davidmogar:
This is still untested but it passes de CI now.

PR #31, opened 2022-08-30
feat(HACBS-1050): Add codecov to integration-service

srivickynesh:
Jira : [HACBS-1050](https://issues.redhat.com/browse/HACBS-1050)

Signed-off-by: Sri Vignesh <sselvan@redhat.com>

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -1,0 +1,17 @@
+# Documentation: https://docs.codecov.io/docs/codecov-yaml
+ignore:
+  - "api/v1alpha1/zz_generated.deepcopy.go" # generated file
+
+coverage:
+    status:
+        # Allows coverage to drop by a 3% when compared against the base commit.
+        project:
+            default:
+                target: auto
+                # this allows a 3% drop from the previous base commit coverage
+                threshold: 3%
+        # Sets the expected status for `codecov/patch` check.
+        # We set this to be only informational hence it does not cause the check to fail.
+        patch:
+            default:
+                informational: true
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -1,0 +1,21 @@
+name: Code Coverage Report
+on:
+  push:
+    branches:
+      - main
+jobs:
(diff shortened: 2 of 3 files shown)


codecov.yml
```diff
@@ -0,0 +1,13 @@
...
+        # Allows coverage to drop by a 3% when compared against the base commit.
+        project:
+            default:
+                target: auto
+                threshold: 3%
+        # Sets the expected status for `codecov/patch` check.
+        # We set this to be only informational hence it does not cause the check to fail.
+        patch:
+            default:
+                informational: true
```
Dannyb48, line 13: Do we not want to block? or maybe I don't know enough about this option?

codecov.yml
```diff
@@ -0,0 +1,13 @@
+# Documentation: https://docs.codecov.io/docs/codecov-yaml
```
dirgim, line 1: Can we also add the setting to ignore the generated `api/v1alpha1/zz_generated.deepcopy.go` file? Example from the release-service: https://github.com/redhat-appstudio/release-service/blob/main/codecov.yml#L2-L3.

srivickynesh: okay, will add.

codecov.yml
dirgim: I think that we do not want to block on codecov because we've been hearing that codecov sometimes behaves weirdly for some PRs.

srivickynesh: Sometimes the difference calculated by codecov is 100% from the previous commit.
Informational only - this way we still have the info but it does not fail if the commit in the same PR has a difference of 100%.
Checking the coverage diff between commits (patch) as it tends to change when rebasing the code.

codecov.yml
dirgim: thanks!

dirgim (approved):
LGTM

codecov.yml
Dannyb48: Ack

Dannyb48 (approved):
LGTM

PR #34, opened 2022-09-05
[HACBS-463]- trigger pipelineruns by integrationservice

jsztuka:
Following changes adds capability for integration-service.

- it registers all integration test scenarios for given application
- checks for last pipeline run for each integration test scenario, if it does not exist, it will create pipelinerun for it


controllers/integration/integration_adapter.go
```diff
@@ -90,10 +119,43 @@ func (a *Adapter) EnsureApplicationSnapshotExists() (results.OperationResult, er
 		return results.RequeueOnErrorOrStop(a.updateStatus())
 	}
 
+	newIntegrationTestScenarios, err := a.getAllIntegrationTestScenariosForApplication(a.application)
```
dirgim, line 122: nit: I think we should rename `newIntegrationTestScenarios` to `allIntegrationTestScenarios` or similar as they are not new scenarios, but existing ones

controllers/snapshot/snapshot_adapter.go
```diff
@@ -60,7 +63,41 @@ func (a *Adapter) EnsureAllIntegrationTestPipelinesExist() (results.OperationRes
 		a.logger.Info("The Snapshot has finished testing.")
 		return results.ContinueProcessing()
 	}
+
 	a.logger.Info("Placeholder for triggering all Integration pipelines (including optional ones)")
```
dirgim, line 67: We can probably remove or modify this line now

controllers/snapshot/snapshot_adapter.go
```diff
@@ -60,7 +63,41 @@ func (a *Adapter) EnsureAllIntegrationTestPipelinesExist() (results.OperationRes
...
+	integrationTestScenarios, err := helpers.GetAllIntegrationTestScenariosForApplication(a.client, a.context, a.application)
+
+	if err != nil {
+		a.logger.Error(err, "Failed to get Integration test scenarios for following application",
+			"Application.Name", a.application.Name,
+			"Application.Namespace", a.application.Namespace)
+	}
+
+	for _, integrationTestScenarios := range *integrationTestScenarios {
```
dirgim, line 76: Maybe we can rename the `integrationTestScenarios` to `integrationTestScenario` singular for the loop variable as it is a single integrationTestScenario.

tekton/integration_pipeline.go
```diff
@@ -0,0 +1,167 @@
...
+	if r.ObjectMeta.Labels == nil {
+		r.ObjectMeta.Labels = map[string]string{}
+	}
+	r.ObjectMeta.Labels[SnapshotNameLabel] = snapshot.Name
+
+	return r
+}
+
+// WithIntegrationLabels adds the type and IntegrationTestScenario name as labels to the Integration PipelineRun.
+func (r *IntegrationPipelineRun) WithIntegrationLabels(scenarioName string) *IntegrationPipelineRun {
```
dirgim, line 146: I think we can have a full `integrationTestScenario` object as the function's parameter here, instead of just the `scenarioName` string, this will match the other `With*` functions.

controllers/snapshot/snapshot_adapter.go
```diff
@@ -286,3 +322,21 @@ func (a *Adapter) getAllApplicationComponents(application *hasv1alpha1.Applicati
 
 	return &applicationComponents.Items, nil
 }
+
+// createIntegrationPipelineRun creates and returns a new integration PipelineRun. The Pipeline information and the parameters to it
+// will be extracted from the given integrationScenario. The integration's ApplicationSnapshot will also be passed to the
+// integration PipelineRun.
+func (a *Adapter) createIntegrationPipelineRun(application *hasv1alpha1.Application, integrationTestScenario *v1alpha1.IntegrationTestScenario, applicationSnapshot *appstudioshared.ApplicationSnapshot) error {
```
Dannyb48, line 329: We need to also append any labels and annotations that were on the build pipelinerun to the test pipelinerun for traceability of commits. So we may need to add a `getAllAnnotationsAndLabelsFromPipelineRun` function to the pipeline_adapter (since it has all the get functions to extract form the build pipeline). We could assign them to the ApplicationSnapshot and then extract them from the Snapshot to assign them to the Integration Pipeline as a simple solution

jsztuka: I'm not sure whether we are getting any labels from build pipelinerun at the moment. In case we do not, does it make sense to have a story regarding that behaviour?

Dannyb48: Yes let's do a story. I will creat it. And add that functionality as part of PR support

Dannyb48 (approved):
LGTM

PR #35, opened 2022-09-06
feat(HACBS-1011):Update the container image in the global list to newly build image

jinqi7:
Update the container image in the global candidate list to newly build image

### DIFF
--- a/controllers/snapshot/snapshot_adapter.go
+++ b/controllers/snapshot/snapshot_adapter.go
@@ -88,6 +88,25 @@ func (a *Adapter) EnsureAllIntegrationTestPipelinesExist() (results.OperationRes
 	return results.ContinueProcessing()
 }
 
+// EnsureGlobalComponentImageUpdated is an operation that ensure the ContainerImage in the Global Candidate List
+// being updated when the ApplicationSnapshot passed all the integration tests
+func (a *Adapter) EnsureGlobalComponentImageUpdated() (results.OperationResult, error) {
+	if (a.component != nil) && gitops.HaveHACBSTestsSucceeded(a.snapshot) {
+		patch := client.MergeFrom(a.component.DeepCopy())
+		for _, component := range a.snapshot.Spec.Components {
+			if component.Name == a.component.Name {
+				a.component.Spec.ContainerImage = component.ContainerImage
+				err := a.client.Patch(a.context, a.component, patch)
+				if err != nil {
+					a.logger.Error(err, "Failed to update Global Candidate for the Component", a.component.Name)
+					return results.RequeueWithError(err)
+				}
+			}
+		}
+	}
+	return results.ContinueProcessing()
+}
+
 // EnsureAllReleasesExist is an operation that will ensure that all pipeline Releases associated
 // to the ApplicationSnapshot and the Application's ReleasePlans exist.
 // Otherwise, it will create new Releases for each ReleasePlan.
--- a/controllers/pipeline/pipeline_adapter.go
+++ b/controllers/pipeline/pipeline_adapter.go
@@ -324,7 +324,7 @@ func (a *Adapter) prepareApplicationSnapshotForPipelineRun(pipelineRun *tektonv1
 
 	var components []appstudioshared.ApplicationSnapshotComponent
 	for _, applicationComponent := range *applicationComponents {
-		pullSpec := applicationComponent.Status.ContainerImage
+		pullSpec := applicationComponent.Spec.ContainerImage
 		if applicationComponent.Name == component.Name {
 			pullSpec, err = a.getImagePullSpecFromPipelineRun(pipelineRun)
 			if err != nil {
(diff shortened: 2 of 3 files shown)


controllers/snapshot/snapshot_adapter.go
```diff
@@ -88,6 +88,26 @@ func (a *Adapter) EnsureAllIntegrationTestPipelinesExist() (results.OperationRes
...
+func (a *Adapter) EnsureGlobalComponentImageUpdated() (results.OperationResult, error) {
+	if gitops.HaveHACBSTestsSucceeded(a.snapshot) {
+		patch := client.MergeFrom(a.component.DeepCopy())
+		for _, component := range a.snapshot.Spec.Components {
+			if component.Name == a.component.Name {
+				a.component.Spec.ContainerImage = component.ContainerImage
+				break
+			}
+		}
+		err := a.client.Patch(a.context, a.component, patch)
```
dirgim, line 102: I believe that in order to patch the `Status` subresource we need to use the `client.Status.Patch`, e.g. in this case:
`err := a.client.Status.Patch(a.context, a.component, patch)`

controllers/snapshot/snapshot_adapter.go
```diff
@@ -88,6 +88,26 @@ func (a *Adapter) EnsureAllIntegrationTestPipelinesExist() (results.OperationRes
...
 }
 
+// EnsureGlobalComponentImageUpdated is an operation that ensure the ContainerImage in the Global Candidate List
+// being updated when the ApplicationSnapshot passed all the integration tests
+func (a *Adapter) EnsureGlobalComponentImageUpdated() (results.OperationResult, error) {
+	if gitops.HaveHACBSTestsSucceeded(a.snapshot) {
+		patch := client.MergeFrom(a.component.DeepCopy())
+		for _, component := range a.snapshot.Spec.Components {
+			if component.Name == a.component.Name {
+				a.component.Spec.ContainerImage = component.ContainerImage
```
dirgim, line 98: I think that we need to update the a.component.Status.[ContainerImage](https://github.com/redhat-appstudio/application-service/blob/main/api/v1alpha1/component_types.go#L122) here instead of the `Spec.ContainerImage`

dirgim: I might be wrong on this, will update after I get more info

dirgim: Disregard these comments, the `Spec.ContainerImage` is the correct field.

controllers/snapshot/snapshot_adapter.go
dirgim: Disregard this, I've gotten confirmation that only the component `Spec` is supposed to be updated.

controllers/snapshot/snapshot_adapter.go
```diff
@@ -88,6 +88,25 @@ func (a *Adapter) EnsureAllIntegrationTestPipelinesExist() (results.OperationRes
 	return results.ContinueProcessing()
 }
 
+// EnsureGlobalComponentImageUpdated is an operation that ensure the ContainerImage in the Global Candidate List
+// being updated when the ApplicationSnapshot passed all the integration tests
+func (a *Adapter) EnsureGlobalComponentImageUpdated() (results.OperationResult, error) {
+	if gitops.HaveHACBSTestsSucceeded(a.snapshot) {
+		patch := client.MergeFrom(a.component.DeepCopy())
+		for _, component := range a.snapshot.Spec.Components {
+			if component.Name == a.component.Name {
```
dirgim, line 97: We will also need to check if `a.component` is `nil` because not all ApplicationSnapshots are associated with a single component.

dirgim: We may do that check even earlier, at [line 94](https://github.com/redhat-appstudio/integration-service/pull/35/files#diff-8e718c6bf06f754db7e9e94c6f74b5c6704941dff95d637289e3c4cc194c95a0R94)

jinqi7: > We will also need to check if `a.component` is `nil` because not all ApplicationSnapshots are associated with a single component.
Do you mean to check "a.snapshot.Spec.Components" which are associated with the snapshot?

dirgim:
@jinqi7 can you also add `feat(HACBS-1011):` prefix to the PR and commit titles?

controllers/snapshot/snapshot_adapter.go
dirgim: We just need to check the `a.component`.

dirgim (approved):
LGTM, nice work!

jsztuka:
The error we are running in within our CI is that for some reason it can't clone repo, it is not connected to the changes.

jinqi7:
> @jinqi7 can you also add `feat(HACBS-1011):` prefix to the PR and commit titles?

Updated

hongweiliu17 (approved):
LGTM

PR #36, opened 2022-09-08
Add kcp workspace support to two controller pipeline and snapshot

hongweiliu17:
Add kcp workspace to controller to pipeline and snapshot
Add samples to run integration service on kcp workspace

JIRA: HACBS-713


config/kcp/samples/apibinding_integration.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: apis.kcp.dev/v1alpha1
+kind: APIBinding
+metadata:
+  name: kubernetes
+spec:
+  reference:
+    workspace:
+      path: root:users:xr:vd:rh-sso-hongweiliu:hongliu-dev-7
```
jsztuka, line 8: I suppose these need to change

hongweiliu17: Yes, updated. Thanks!

Dannyb48 (approved):
LGTM

syncer710.yaml
```diff
@@ -0,0 +1,144 @@
+---
```
dirgim, line 1: Is this file needed here or can it be removed?

hongweiliu17: No, it is not needed, when I use it, I have a namespace creation part above. I will remove it. Thanks.

hongweiliu17: This is an example generated by command `kubectl kcp workload sync mycluster` and I add some CRD into it.

dirgim (approved):
LGTM, just the `bin/controller-gen` needs to be resolved before merging

PR #32, opened 2022-08-30
Log the count of scenarios found

ralphbean:
To help with debugging in staging.

See https://coreos.slack.com/archives/C030SGP2VB9/p1661881854052129

PR #45, opened 2022-09-23
feat: log the count of scenarios found

dirgim:
* Added logging to help debugging in staging

Replaces #32 because of refactoring changes

### DIFF
--- a/controllers/snapshot/snapshot_adapter.go
+++ b/controllers/snapshot/snapshot_adapter.go
@@ -75,6 +75,9 @@ func (a *Adapter) EnsureAllIntegrationTestPipelinesExist() (results.OperationRes
 	}
 
 	if integrationTestScenarios != nil {
+		a.logger.Info("Found IntegrationTestScenarios for application",
+			"Application.Name", a.application.Name,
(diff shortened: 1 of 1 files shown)


controllers/snapshot/snapshot_adapter.go
```diff
@@ -72,6 +72,10 @@ func (a *Adapter) EnsureAllIntegrationTestPipelinesExist() (results.OperationRes
 			"Application.Namespace", a.application.Namespace)
 	}
 
+	a.logger.Info("Found IntegrationTestScenarios for application",
```
srivickynesh, line 75: Also good to add this within https://github.com/redhat-appstudio/integration-service/blob/main/helpers/integration.go#L58.

dirgim: Yes, I think it's better to handle the logging here, within the `snapshot_adapter.go`, as we're able to tailor the message to the use-case.

srivickynesh (approved):
LGTM

hongweiliu17 (approved):
LGTM

PR #54, opened 2022-10-13
feat(hacbs-597): implement integration service PR orchestration strategy

jinqi7:
Signed-off-by: jinqi <jinqi@redhat.com>


controllers/snapshot/snapshot_adapter_test.go
```diff
@@ -32,6 +32,7 @@ var _ = Describe("Snapshot Adapter", Ordered, func() {
 		hasApp                  *applicationapiv1alpha1.Application
 		hasComp                 *applicationapiv1alpha1.Component
 		hasSnapshot             *applicationapiv1alpha1.ApplicationSnapshot
+		hasSnapshot1            *applicationapiv1alpha1.ApplicationSnapshot
```
jsztuka, line 35: maybe renaming this variable to something relevant to the `pull_request` ? 
Something like `hasSnapshotPullRequest` ?

jinqi7: Sure. Updated to  "hasSnapshotPR".

controllers/snapshot/snapshot_adapter.go
```diff
@@ -129,7 +129,8 @@ func (a *Adapter) EnsureAllIntegrationTestPipelinesExist() (results.OperationRes
 // EnsureGlobalComponentImageUpdated is an operation that ensure the ContainerImage in the Global Candidate List
 // being updated when the ApplicationSnapshot passed all the integration tests
 func (a *Adapter) EnsureGlobalComponentImageUpdated() (results.OperationResult, error) {
-	if (a.component != nil) && gitops.HaveHACBSTestsSucceeded(a.snapshot) {
+	if (a.component != nil) && gitops.HaveHACBSTestsSucceeded(a.snapshot) &&
+		a.snapshot.Labels[gitops.PipelineAsCodeEventTypeLabel] == gitops.PipelineAsCodePushType {
```
dirgim, line 133: I think we'll need to also check if the snapshot has the `gitops.PipelineAsCodeEventTypeLabel` label at all. Might be best to create a new function that checks if the label is the `push` type and returns a boolean if it is.

jinqi7: OK. Updated, thanks!

controllers/snapshot/snapshot_adapter.go
```diff
@@ -155,6 +155,10 @@ func (a *Adapter) EnsureAllReleasesExist() (results.OperationResult, error) {
 		return results.ContinueProcessing()
 	}
 
+	if !gitops.IsSnapshotCreatedByPushEvent(a.snapshot) {
+		return results.ContinueProcessing()
```
dirgim, line 159: tiny nit: maybe also add an `a.logger.Info` call informing that the snapshot won't be released because it was not created for a push event.

controllers/snapshot/snapshot_adapter.go
```diff
@@ -187,6 +191,10 @@ func (a *Adapter) EnsureApplicationSnapshotEnvironmentBindingExist() (results.Op
...
```
dirgim, line 195: tiny nit: maybe also add an a.logger.Info call informing that the snapshot won't be released because it was not created for a push event.

controllers/snapshot/snapshot_adapter.go
dirgim: Thanks, nicely implemented!

dirgim:
@jinqi7 @jsztuka I'm thinking if we'll also need to modify the [CompareApplicationSnapshots](https://github.com/redhat-appstudio/integration-service/blob/19b81aed02cbe8bf9524f6f3856cc2b9feef8cce/gitops/snapshot.go#L143) function to also take into account if both snapshots were created by a push event. Wdyt?

controllers/snapshot/snapshot_adapter.go
jinqi7: Added. Thanks!

controllers/snapshot/snapshot_adapter.go
jinqi7: Added. Thanks

jinqi7:
From the current logic, the compareApplicationSnapshots function is called under the condition of   gitops.IsSnapshotCreatedByPushEvent(existingApplicationSnapshot) as true. But, to make the function itself more safe, I added the below compariation logic-
if IsSnapshotCreatedByPushEvent(expectedApplicationSnapshot) !=  IsSnapshotCreatedByPushEvent(foundApplicationSnapshot) {
                return false
        }

How do you think of it?

> @jinqi7 @jsztuka I'm thinking if we'll also need to modify the [CompareApplicationSnapshots](https://github.com/redhat-appstudio/integration-service/blob/19b81aed02cbe8bf9524f6f3856cc2b9feef8cce/gitops/snapshot.go#L143) function to also take into account if both snapshots were created by a push event. Wdyt?

jsztuka (approved):
lgtm

dirgim (approved):
LGTM

PR #55, opened 2022-10-14
feat(HACBS-1150) fix apiexport for infra-deployments

hongweiliu17:
Signed-off-by: Hongwei Liu  hongliu@redhat.com

### DIFF
--- a/config/kcp-rbac/role.yaml
+++ b/config/kcp-rbac/role.yaml
@@ -50,4 +50,4 @@ rules:
   resources:
   - apiexports/content
   verbs:
-  - update
+  - '*'
--- a/config/kcp/kustomization.yaml
+++ b/config/kcp/kustomization.yaml
@@ -1,32 +1,31 @@
 # Adds namespace to all resources.
-namespace: integration-service-system
+# namespace: integration-service-system
 
 # Value of this field is prepended to the
 # names of all resources, e.g. a deployment named
 # "wordpress" becomes "alices-wordpress".
 # Note that it should also match with the prefix (text before '-') of the namespace
 # field above.
-namePrefix: integration-service-
+# namePrefix: integration-service-
 
 # Labels to add to all resources and selectors.
 #commonLabels:
 #  someName: someValue
 
 bases:
-  - ../crd
   - ../manager
-  - ../rbac
+  - ../kcp-rbac
 
 resources:
   - apiexport_integration.yaml
   - apiresourceschema_integration.yaml
 
 patches:
-# Needs to use default service account name right now
+# Add APIExport flag to the operator arguments list
 - patch: |-
-    - op: replace
-      path: /spec/template/spec/serviceAccountName
-      value: default
+    - op: add
+      path: /spec/template/spec/containers/0/args/-
+      value: --api-export-name=integration-service
   target:
     kind: Deployment
-    name: controller-manager
+    name: integration-service-controller-manager
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -18,7 +18,7 @@ rules:
   resources:
   - apiexports/content
   verbs:
-  - update
+  - '*'
 - apiGroups:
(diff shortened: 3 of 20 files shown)


dirgim:
The gitlint check is reporting that the commit is missing a colon in its title, it should be:
```
feat(HACBS-1150): fix apiexport for infra-deployments
```

config/kcp/kustomization.yaml
```diff
@@ -1,32 +1,40 @@
...
-  - ../rbac
+  - ../kcp-rbac
 
 resources:
   - apiexport_integration.yaml
   - apiresourceschema_integration.yaml
 
 patches:
 # Needs to use default service account name right now
+# - patch: |-
```
hongweiliu17, line 26: @dirgim @johnbieren Remove this patch because I guess it can fix the issue in this comment from @Michkov https://github.com/redhat-appstudio/infra-deployments/pull/821#pullrequestreview-1140241007, is it?

config/kcp-rbac/role.yaml
```diff
@@ -50,4 +50,4 @@ rules:
   resources:
   - apiexports/content
   verbs:
-  - update
+  - *
```
Michkov, line 53: must be `'*'` otherwise kustomize failing

hongweiliu17: Fixed. Thanks.

config/kcp/kustomization.yaml
```diff
@@ -1,32 +1,40 @@
...
+#     kind: Deployment
+#     name: controller-manager
+# Add APIExport flag to the operator arguments list
 - patch: |-
-    - op: replace
-      path: /spec/template/spec/serviceAccountName
-      value: default
+    - op: add
+      path: /spec/template/spec/containers/0/args/-
+      value: --api-export-name=release-service
```
Michkov, line 37: `release-service` or `integration-service` ?

hongweiliu17: Ha:)

hongweiliu17: fixed.

config/kcp/kustomization.yaml
dirgim: Yes, it looks that way to me

config/kcp/kustomization.yaml
```diff
@@ -1,32 +1,40 @@
...
-namespace: integration-service-system
+# namespace: integration-service-system
 
 # Value of this field is prepended to the
 # names of all resources, e.g. a deployment named
 # "wordpress" becomes "alices-wordpress".
 # Note that it should also match with the prefix (text before '-') of the namespace
 # field above.
-namePrefix: integration-service-
+# namePrefix: integration-service-
```
Michkov, line 9: could you rename deployment `controller-manager` to `integration-controller-manager`? That would be pretty useful for searching for deployment on physical cluster.

Michkov (approved):
please rename controller otherwise it's good.

dirgim (approved):
LGTM

config/rbac/role.yaml
```diff
@@ -18,7 +18,7 @@ rules:
   resources:
   - apiexports/content
   verbs:
-  - update
```
johnbieren, line 21: The `kcp/kustomization.yaml` file uses `kcp-rbac` dir, not `rbac` dir, so I don't think this change is really needed, but doesn't hurt I guess

johnbieren (commented):
All the changes from default to controller-manager I guess work.. not sure why some of them are needed. Adding a namePrefix to config/manager/kustomization.yaml seems to work just as well for us. But if this works, great

config/rbac/role.yaml
hongweiliu17: Thanks, I agree. It is added when I run `made manifests`.

johnbieren: Ah, makes sense!

PR #57, opened 2022-10-18
fix(HACBS-1256): add fix for unit test errors during execution

srivickynesh:
Signed-off-by: Sri Vignesh <sselvan@redhat.com>

PR #58, opened 2022-10-20
feat(HACBS-1223): add EnvironmentConfiguration to IntegrationTestScenario…

jinqi7:


controllers/snapshot/snapshot_adapter_test.go
```diff
@@ -71,9 +70,13 @@ var _ = Describe("Snapshot Adapter", Ordered, func() {
...
 				Pipeline:    "component-pipeline-pass",
 				Environment: integrationv1alpha1.TestEnvironment{
-					Name:   "envname",
-					Type:   "POC",
-					Params: []string{},
+					Name: "envname",
+					Type: "POC",
+					Configuration: applicationapiv1alpha1.EnvironmentConfiguration{
+						Env: []applicationapiv1alpha1.EnvVarPair{
+							{Name: "", Value: ""},
```
davidmogar, line 77: Is this line needed? Haven't tried myself, but it seems you are just setting nil values.

api/v1alpha1/integrationtestscenario_types.go
```diff
@@ -53,9 +54,9 @@ type PipelineParameter struct {
 
 // TestEnvironment contains the name and values of a Test environment
 type TestEnvironment struct {
-	Name   string   `json:"name"`
-	Type   string   `json:"type"`
-	Params []string `json:"params"`
+	Name          string                                          `json:"name"`
+	Type          string                                          `json:"type"`
```
dirgim, line 58: We could also set the `Type` to be [EnvironmentType](https://github.com/redhat-appstudio/application-api/blob/932f11859a29e8e413e9cd270e9d2cd3fa0a88ca/api/v1alpha1/environment_types.go#L54) instead of `string` to help with future compatibility.

jinqi7: Updated, thanks!

controllers/snapshot/snapshot_adapter_test.go
jinqi7: Updated them to some temp values and will use them in the coming test case later. Thanks

api/v1alpha1/integrationtestscenario_types.go
dirgim: Thanks for the change!

controllers/snapshot/snapshot_controller.go
```diff
@@ -132,9 +132,9 @@ func (r *Reconciler) getComponentFromSnapshot(context context.Context, snapshot
 // AdapterInterface is an interface defining all the operations that should be defined in an Integration adapter.
 type AdapterInterface interface {
 	EnsureAllReleasesExist() (results.OperationResult, error)
+	EnsureAllIntegrationTestPipelinesExist() (results.OperationResult, error)
```
dirgim, line 135: Is this a required change? I know that #57 modified the order when fixing unit test errors, so wondering if this is related to that.

jinqi7: Because I met the "panic: runtime error: invalid memory address or nil pointer dereference" issue and it helps to pass by changing it back although it may not be the root cause.

dirgim: Ok, it seems that it works this way. We'll need to look into the issue some more, but that doesn't affect this PR.

hongweiliu17 (approved):
LGTM

hongweiliu17:
@jinqi7 @dirgim I just realize we need to run https://github.com/redhat-appstudio/integration-service/blob/main/hack/generate-kcp-api.sh to re-generate apiresourceschema https://github.com/redhat-appstudio/integration-service/tree/main/config/kcp for kcp because of the change in this MR, or we may file a ticket to track the task then do it after EPIC https://issues.redhat.com/browse/HACBS-802 is done.

dirgim:
> @jinqi7 @dirgim I just realize we need to run https://github.com/redhat-appstudio/integration-service/blob/main/hack/generate-kcp-api.sh to re-generate apiresourceschema https://github.com/redhat-appstudio/integration-service/tree/main/config/kcp for kcp because of the change in this MR, or we may file a ticket to track the task then do it after EPIC https://issues.redhat.com/browse/HACBS-802 is done.

@hongweiliu17 that's a good point, I will note down that all changes to our API definitions need to also re-generate the `apiresourceschema`. I think it would make sense to do the change immediately.
@jinqi7 can you create a follow-up PR with the regenerated `apiresourceschema` by running the `generate-kcp-api.sh` script?

PR #59, opened 2022-10-20
feat(HACBS-1284): add unit testing for integration functions

srivickynesh:
Signed-off-by: Sri Vignesh <sselvan@redhat.com>

### DIFF
--- a/helpers/integration_test.go
+++ b/helpers/integration_test.go
@@ -1,0 +1,290 @@
+package helpers_test
+
+import (
+	"time"
+
+	"github.com/go-logr/logr"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	tektonv1beta1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+
+	applicationapiv1alpha1 "github.com/redhat-appstudio/application-api/api/v1alpha1"
+	"github.com/redhat-appstudio/integration-service/gitops"
+	"github.com/redhat-appstudio/integration-service/helpers"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+)
+
+var _ = Describe("Pipeline Adapter", Ordered, func() {
+	var (
+		testpipelineRun *tektonv1beta1.PipelineRun
+		hasApp          *applicationapiv1alpha1.Application
+		hasSnapshot     *applicationapiv1alpha1.ApplicationSnapshot
+		logger          logr.Logger
+		sample_image    string
+	)
+
+	BeforeAll(func() {
+		logger = logf.Log.WithName("helpers_test")
+
+		hasApp = &applicationapiv1alpha1.Application{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "application-sample",
+				Namespace: "default",
+			},
+			Spec: applicationapiv1alpha1.ApplicationSpec{
+				DisplayName: "application-sample",
+				Description: "This is an example application",
+			},
+		}
+		Expect(k8sClient.Create(ctx, hasApp)).Should(Succeed())
+	})
+
+	BeforeEach(func() {
+		sample_image = "quay.io/redhat-appstudio/sample-image"
+
+		hasSnapshot = &applicationapiv1alpha1.ApplicationSnapshot{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "snapshot-sample",
+				Namespace: "default",
+				Labels: map[string]string{
+					gitops.ApplicationSnapshotTypeLabel:      "component",
+					gitops.ApplicationSnapshotComponentLabel: "component-sample",
+				},
+			},
+			Spec: applicationapiv1alpha1.ApplicationSnapshotSpec{
+				Application: hasApp.Name,
+				Components: []applicationapiv1alpha1.ApplicationSnapshotComponent{
+					{
+						Name:           "component-sample",
+						ContainerImage: sample_image,
+					},
+				},
+			},
+		}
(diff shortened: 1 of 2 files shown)


helpers/integration_test.go
```diff
@@ -0,0 +1,192 @@
...
+
+		integrationTestScenarios, err := helpers.GetAllIntegrationTestScenariosForApplication(k8sClient, ctx, hasApp)
+		Expect(err == nil && integrationTestScenarios != nil).To(BeTrue())
+
+		for _, integrationTestScenario := range *integrationTestScenarios {
+			integrationTestScenario := integrationTestScenario //G601
+			integrationPipelineRun, err := helpers.GetLatestPipelineRunForApplicationSnapshotAndScenario(k8sClient, ctx, hasApp, hasSnapshot, &integrationTestScenario)
+			Expect(err != nil && integrationPipelineRun == nil)
+		}
+		pipelineRunOutcome, err := helpers.CalculateIntegrationPipelineRunOutcome(logger, testpipelineRun)
```
dirgim, line 154: After giving this some thought, I think we'll need to also have separate unit tests for the `CalculateIntegrationPipelineRunOutcome` function. 
Since it's such a key part of the integration service logic, I think it would be a good idea to have unit tests for three different pipelineRuns:

- A `pipelineRun` with one (or more) tasks which have the `HACBS_TEST_OUTPUT` Tekton result with a `SUCCESS` result
- A `pipelineRun` with at least one task which has the `HACBS_TEST_OUTPUT` Tekton result with a `FAILED` result
- A `pipelineRun` with 0 tasks that have the `HACBS_TEST_OUTPUT` result

helpers/integration_test.go
```diff
@@ -0,0 +1,230 @@
...
+	. "github.com/onsi/gomega"
+	tektonv1beta1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	applicationapiv1alpha1 "github.com/redhat-appstudio/application-api/api/v1alpha1"
+	"github.com/redhat-appstudio/integration-service/gitops"
+	"github.com/redhat-appstudio/integration-service/helpers"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
```
dirgim, line 17: The staticcheck-action is complaining about the `v1 "k8s.io/apimachinery/pkg/apis/meta/v1` being imported twice here: https://github.com/redhat-appstudio/integration-service/actions/runs/3304067425/jobs/5452735358

srivickynesh: Done.

helpers/integration_test.go
srivickynesh: Added the above checks

helpers/integration_test.go
```diff
@@ -0,0 +1,229 @@
...
+						},
+					},
+				},
+			},
+		}
+		Expect(k8sClient.Create(ctx, testpipelineRun)).Should(Succeed())
+
+		testpipelineRun.Status = tektonv1beta1.PipelineRunStatus{
+			PipelineRunStatusFields: tektonv1beta1.PipelineRunStatusFields{
+				TaskRuns: map[string]*tektonv1beta1.PipelineRunTaskRunStatus{
```
dirgim, line 103: The integration test pipeline run is not supposed to contain the `build-container` task with the `IMAGE_DIGEST` Tekton result. It should have a number of tasks that each contains the `HACBS_TEST_OUTPUT` Tekton result with JSON contents like in [these examples](https://docs.google.com/document/d/1FJX7TH5wBTcWdU2VYuecmkY5O69DGpItN52A-JaoCtU/edit#heading=h.ffpp5iwdprkv).
In order to cover the different test cases, we'll need to have multiple pipelineRun examples which contain different combinations of tasks that have the `result` field set to `SUCCESS`, `FAILURE` or `SKIPPED`.

helpers/integration_test.go
dirgim: Thanks for adding the checks, but we'll need to modify the `testPipelineRun` to have `HACBS_TEST_OUTPUT` task results in order to test the decision logic.

helpers/integration_test.go
```diff
@@ -0,0 +1,229 @@
...
+	It("ensures pipelinerun outcome when HACBSTests succeeded", func() {
+		gitops.MarkSnapshotAsPassed(k8sClient, ctx, hasSnapshot, "test passed")
+		Expect(gitops.HaveHACBSTestsSucceeded(hasSnapshot)).To(BeTrue())
+
+		pipelineRunOutcome, err := helpers.CalculateIntegrationPipelineRunOutcome(logger, testpipelineRun)
+		Expect(err == nil && pipelineRunOutcome).To(BeTrue())
+	})
+
+	It("no error from pipelinrun when HACBSTests failed", func() {
+		gitops.MarkSnapshotAsFailed(k8sClient, ctx, hasSnapshot, "test failed")
```
dirgim, line 204: Marking the snapshot as failed doesn't affect the outcome `CalculateIntegrationPipelineRunOutcome` function - it's the opposite, when the `CalculateIntegrationPipelineRunOutcome` returns `false` we then call the `MarkSnapshotAsFailed` function.
The contents of the taskRun results are what affect the outcome of `CalculateIntegrationPipelineRunOutcome`.

helpers/integration_test.go
dirgim: I added comments to the relevant places, resolving this part of the discussion

helpers/integration_test.go
dirgim: Thanks!

helpers/integration_test.go
srivickynesh: Done

helpers/integration_test.go
srivickynesh: Done

dirgim: Thanks!

helpers/integration_test.go
dirgim: Thanks!

dirgim (approved):
LGTM

jinqi7 (approved):

PR #61, opened 2022-10-24
[HACBS-621] Report integration PipelineRun status to GitHub

amisstea:
Opening as a draft to get some initial feedback since I'm still new to golang and this project. Please let me know if I've violated any best practices :smile:!


controllers/pipeline/pipeline_adapter.go
```diff
@@ -195,6 +196,21 @@ func (a *Adapter) EnsureApplicationSnapshotPassedAllTests() (results.OperationRe
 	return results.ContinueProcessing()
 }
 
+// EnsureStatusReported will ensure that integration PipelineRun status is reported to the git provider
+// which (indirectly) triggered its execution.
+func (a *Adapter) EnsureStatusReported() (results.OperationResult, error) {
+	if !tekton.IsIntegrationPipelineRun(a.pipelineRun) {
```
dirgim, line 202: I think we'll also need to check if the pipeline was run for the `pull_request` event type by checking the [PipelineAsCodeEventTypeLabel](pipelinesascode.tekton.dev/event-type) on the pipelineRun.

amisstea: This check is now performed by each `Reporter` implementation (e.g. https://github.com/redhat-appstudio/integration-service/pull/61/files#diff-e3efbea6f0e41cd39d3da695764ffb088284311c7a59a8c5b53a27fffe33d8f1R167). This should make it flexible should we ever wish to also react to PipelineRun events on a provider-by-provider basis.

git/github.go
```diff
@@ -0,0 +1,261 @@
...
+
+// createOrUpdateCheckRun reports state using GitHub's CheckRun API. It can only be used with GitHub Apps.
+func (a *GitHubAdapter) createOrUpdateCheckRun(client *github.Client, appID int64, gtpr *GitTriggeredPipelineRun) (*github.CheckRun, error) {
+	var cr *github.CheckRun
+
+	var status, title, summary, conclusion string
+	var startTime time.Time
+	var completionTime time.Time
+
+	checkName := AppName + " / " + gtpr.Component + " / " + gtpr.TestScenario
```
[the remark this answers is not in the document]
amisstea: It doesn't need to be unique. GitHub assumes the latest check run using a given name should supersede prior runs. It will only display the newer run too.

status/status.go
```diff
@@ -0,0 +1,84 @@
...
+	SHALabel          = "pipelinesascode.tekton.dev/sha"
+	OrgLabel          = "pipelinesascode.tekton.dev/url-org"
+	RepositoryLabel   = "pipelinesascode.tekton.dev/url-repository"
+	TestScenarioLabel = "test.appstudio.openshift.io/scenario"
+	ComponentLabel    = "test.appstudio.openshift.io/component"
+
+	InstallationIDAnnotation = "pipelinesascode.tekton.dev/installation-id"
+
+	GitHubProvider       = "github"
+	PullRequestEventType = "pull_request"
```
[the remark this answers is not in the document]
amisstea: Good point. I consolidated almost all of these constants in the `gitops` package.

git/github.go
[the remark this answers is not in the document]
jinqi7: Sorry, the two commits I meant are two PRs. Each PR in Github should get a pipelinerun result, right?

controllers/pipeline/pipeline_adapter.go
dirgim: Cool, that makes sense

git/github.go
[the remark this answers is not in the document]
amisstea: Sort of. A build `PipelineRun` will be kicked off for the head commit of the PR's source branch which, upon successful completion, may trigger multiple integration `PipelineRuns` (one per `IntegrationTestScenario`). There can be multiple `CheckRuns` for a given commit and so the name is used to group them under distinct workflows. Each `CheckRun` also has a unique `ExternalID` which matches the name of the `PipelineRun`. There should only ever be a single `CheckRun` per `PipelineRun`. Does that help?

config/rbac/role.yaml
```diff
@@ -77,3 +77,9 @@ rules:
   - get
   - patch
   - update
+- apiGroups:
+  - appstudio.redhat.com
```
hongweiliu17, line 81: I feel it should be `- [""]` for `secrets`

status/status.go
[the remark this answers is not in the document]
dirgim: Thanks!

config/rbac/role.yaml
amisstea: I think you're right but it should be `- ""` instead (based on this [example](https://kubernetes.io/docs/reference/access-authn-authz/rbac/#role-example)), right?

git/github.go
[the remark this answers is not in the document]
jinqi7: Yes, it helps to understand. Thanks

config/rbac/role.yaml
hongweiliu17: Yes, without `[]`, thanks!

status/reporters.go
```diff
@@ -0,0 +1,203 @@
...
+		return nil, fmt.Errorf("PipelineRun label not found %q", gitops.PipelineAsCodeURLRepositoryLabel)
+	}
+
+	SHA, found := r.pipelineRun.Labels[gitops.PipelineAsCodeSHALabel]
+	if !found {
+		return nil, fmt.Errorf("PipelineRun label not found %q", gitops.PipelineAsCodeSHALabel)
+	}
+
+	var title, conclusion string
+	succeeded := r.pipelineRun.Status.GetCondition(apis.ConditionSucceeded)
```
dirgim, line 115: The Tekton pipelineRun `Succeeded` condition is misleading in our case, because that only means that the pipeline finished with no unexpected errors.
For Integration pipelineRuns (and all other HACBS ones), all pipelines are expected to succeed, and the decision if the tests succeeded is based on the individual TaskRun results. In our case we get those with `GetHACBSTestResultsFromPipelineRun` and then find out if the Integration test passed with `CalculateIntegrationPipelineRunOutcome`.

status/status.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	for _, opt := range opts {
+		opt(&adapter)
+	}
+
+	return &adapter
+}
+
+// GetReporters returns a list of enabled/supported status reporters for a PipelineRun.
```
dirgim, line 58: It might be a good idea for future reference to add a comment here that all new reporters need to be added in this function in order to be used in the workflow.

tekton/utils_test.go
```diff
@@ -0,0 +1,81 @@
...
+					},
+				},
+			},
+		}
+	})
+
+	It("can get output-image", func() {
+		image, _ := tekton.GetOutputImage(pipelineRun)
+		if image != "test-image" {
+			Fail(fmt.Sprintf("Expected image is test-image, bug got %s", image))
```
jsztuka, line 58: bug -> but?

tekton/utils_test.go
```diff
@@ -0,0 +1,81 @@
...
+		}
+		klog.Infoln("Got expected image")
+	})
+
+	It("can get output-image-digest", func() {
+		image_digest, _ := tekton.GetOutputImageDigest(pipelineRun)
+		if image_digest != "image_digest_value" {
+			Fail(fmt.Sprintf("Expected image_digest is image_digest_value, bug got %s", image_digest))
```
jsztuka, line 66: Same like 58

status/status.go
amisstea: Comment added.

tekton/utils_test.go
amisstea: Fixed.

tekton/utils_test.go
amisstea: Fixed.

status/reporters.go
amisstea: Fixed in fb52df37823ee4f205821cea4dcc85ab7d028e03.

config/rbac/role.yaml
amisstea: This has been fixed in 05567bfbaa69715ecc5dacf23c54b5cf377ac456.

jsztuka (approved):
lgtm

status/reporters.go
dirgim: Thanks, the changes look good

dirgim (approved):
LGTM, nice work

config/rbac/role.yaml
amisstea: @hongweiliu17 please resolve this convo if you agree with the fix. Thanks!

hongweiliu17: Thanks!

PR #69, opened 2022-11-14
fix(HACBS-1385): change prefix for propagated PaC labels & annotations

amisstea:
Signed-off-by: Alex Misstear <amisstea@redhat.com>

### DIFF
--- a/controllers/pipeline/pipeline_adapter.go
+++ b/controllers/pipeline/pipeline_adapter.go
@@ -385,11 +385,13 @@ func (a *Adapter) prepareSnapshotForPipelineRun(pipelineRun *tektonv1beta1.Pipel
 	snapshot.Labels[gitops.SnapshotTypeLabel] = gitops.SnapshotComponentType
 	snapshot.Labels[gitops.SnapshotComponentLabel] = a.component.Name
 
-	// copy PipelineRun PAC annotations/labels from Build to snapshot
+	// Copy PipelineRun PAC annotations/labels from Build to snapshot.
+	// Modify the prefix so the PaC controller won't react to PipelineRuns generated from the snapshot.
 	pipelinerunLabels := pipelineRun.GetLabels()
 	for key, value := range pipelinerunLabels {
-		if strings.Contains(key, "pipelinesascode.tekton.dev") {
-			snapshot.Labels[key] = value
+		if strings.Contains(key, "pipelinesascode.tekton.dev/") {
+			newKey := strings.ReplaceAll(key, "pipelinesascode.tekton.dev/", "test.appstudio.openshift.io/")
+			snapshot.Labels[newKey] = value
 		}
 	}
 
@@ -398,8 +400,9 @@ func (a *Adapter) prepareSnapshotForPipelineRun(pipelineRun *tektonv1beta1.Pipel
 	}
 	pipelinerunAnnotations := pipelineRun.GetAnnotations()
 	for key, value := range pipelinerunAnnotations {
-		if strings.Contains(key, "pipelinesascode.tekton.dev") {
-			snapshot.Annotations[key] = value
+		if strings.Contains(key, "pipelinesascode.tekton.dev/") {
+			newKey := strings.ReplaceAll(key, "pipelinesascode.tekton.dev/", "test.appstudio.openshift.io/")
+			snapshot.Annotations[newKey] = value
 		}
 	}
 
--- a/controllers/pipeline/pipeline_adapter.go
+++ b/controllers/pipeline/pipeline_adapter.go
@@ -390,7 +390,7 @@ func (a *Adapter) prepareSnapshotForPipelineRun(pipelineRun *tektonv1beta1.Pipel
 	pipelinerunLabels := pipelineRun.GetLabels()
 	for key, value := range pipelinerunLabels {
 		if strings.Contains(key, "pipelinesascode.tekton.dev/") {
-			newKey := strings.ReplaceAll(key, "pipelinesascode.tekton.dev/", "test.appstudio.openshift.io/")
+			newKey := strings.ReplaceAll(key, "pipelinesascode.tekton.dev/", "pac.test.appstudio.openshift.io/")
 			snapshot.Labels[newKey] = value
 		}
 	}
@@ -401,7 +401,7 @@ func (a *Adapter) prepareSnapshotForPipelineRun(pipelineRun *tektonv1beta1.Pipel
 	pipelinerunAnnotations := pipelineRun.GetAnnotations()
 	for key, value := range pipelinerunAnnotations {
 		if strings.Contains(key, "pipelinesascode.tekton.dev/") {
-			newKey := strings.ReplaceAll(key, "pipelinesascode.tekton.dev/", "test.appstudio.openshift.io/")
+			newKey := strings.ReplaceAll(key, "pipelinesascode.tekton.dev/", "pac.test.appstudio.openshift.io/")
 			snapshot.Annotations[newKey] = value
 		}
 	}
--- a/controllers/pipeline/pipeline_adapter.go
+++ b/controllers/pipeline/pipeline_adapter.go
@@ -387,24 +387,8 @@ func (a *Adapter) prepareSnapshotForPipelineRun(pipelineRun *tektonv1beta1.Pipel
 
 	// Copy PipelineRun PAC annotations/labels from Build to snapshot.
 	// Modify the prefix so the PaC controller won't react to PipelineRuns generated from the snapshot.
-	pipelinerunLabels := pipelineRun.GetLabels()
-	for key, value := range pipelinerunLabels {
-		if strings.Contains(key, "pipelinesascode.tekton.dev/") {
-			newKey := strings.ReplaceAll(key, "pipelinesascode.tekton.dev/", "pac.test.appstudio.openshift.io/")
-			snapshot.Labels[newKey] = value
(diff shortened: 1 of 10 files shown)


controllers/pipeline/pipeline_adapter.go
```diff
@@ -385,11 +385,13 @@ func (a *Adapter) prepareSnapshotForPipelineRun(pipelineRun *tektonv1beta1.Pipel
...
 
-	// copy PipelineRun PAC annotations/labels from Build to snapshot
+	// Copy PipelineRun PAC annotations/labels from Build to snapshot.
+	// Modify the prefix so the PaC controller won't react to PipelineRuns generated from the snapshot.
 	pipelinerunLabels := pipelineRun.GetLabels()
 	for key, value := range pipelinerunLabels {
-		if strings.Contains(key, "pipelinesascode.tekton.dev") {
-			snapshot.Labels[key] = value
+		if strings.Contains(key, "pipelinesascode.tekton.dev/") {
+			newKey := strings.ReplaceAll(key, "pipelinesascode.tekton.dev/", "test.appstudio.openshift.io/")
```
amisstea, line 393: Should we use `test.appstudio.openshift.io/` like this or further subdivide the namespace to avoid accidentally overriding labels/annotations? For example, if PaC adds a `pipelinesascode.tekton.dev/component` label at some point it's likely to collide with `test.appstudio.openshift.io/component`? Here are a few ideas:

- `pac.test.appstudio.openshift.io/`
- `pipelinesascode.test.appstudio.openshift.io/`
- `pipelinesascode.tekton.dev.test.appstudio.openshift.io/`

Thoughts?

dirgim: We're aiming to reduce the number of `test.*` labels with [HACBS-1331](https://issues.redhat.com/browse/HACBS-1331), but your point still stands about the possibility of collision.
The `pac.test.appstudio.openshift.io/` looks like the best option to me on account of brevity.

amisstea: Done. All the PaC labels/annoations use the `pac.test.appstudio.openshift.io/` prefix now.

dirgim: Thanks!

dirgim (approved):
LGTM
@davidmogar fyi this change will also affect the PaC labels that are copied to the `Release` CRs that the Integration service creates - they will have the `pac.test.appstudio.openshift.io/` prefix instead of `pipelinesascode.tekton.dev/`.

johnbieren:
I'm not sure where it would be best to put, but is it possible to put the label value in some file as a constant so then others can just import and use it? Definitely a nice to have and not required, but it would make for a lot less code change if the value changes again (and remove the requirement for us to all be aligned on this string)

controllers/pipeline/pipeline_adapter_test.go
```diff
@@ -229,6 +231,18 @@ var _ = Describe("Pipeline Adapter", Ordered, func() {
...
+		Expect(err).To(BeNil())
+		Expect(snapshot).ToNot(BeNil())
+		annotation, found := snapshot.GetAnnotations()["pac.test.appstudio.openshift.io/on-target-branch"]
+		Expect(found).To(BeTrue())
+		Expect(annotation).To(Equal("[main,master]"))
+		label, found := snapshot.GetLabels()["pac.test.appstudio.openshift.io/event-type"]
+		Expect(found).To(BeTrue())
+		Expect(label).To(Equal("pull_request"))
+	})
+
```
parasense, line 245: You might consider adding tests for adverse scenarios, for example ensuring keys without the matching prefix are **NOT** copied to the destination maps. As far as I can tell this test provides full code coverage.
You can look it over with `go tool cover -html cover.out -o cover.html` and then reviewing the report in your browser.

(optional)

status/status_test.go
```diff
@@ -26,7 +26,7 @@ var _ = Describe("Status Adapter", func() {
 		pipelineRun = &tektonv1beta1.PipelineRun{
 			ObjectMeta: metav1.ObjectMeta{
 				Labels: map[string]string{
-					"pipelinesascode.tekton.dev/git-provider": "github",
+					"pac.test.appstudio.openshift.io/git-provider": "github",
```
parasense, line 29: Not to seem repetitive, but just imagine how much smaller the diff would have been if these string literals were referencing a string constant declared somewhere central. Something to think about.

amisstea: A constant has been created but I didn't use it in test-related code. This is intentional to make it difficult to accidentally change the prefix without knowingly correcting the expected values in the tests.

controllers/pipeline/pipeline_adapter_test.go
amisstea: A few more tests have been added. Please take a look.

controllers/pipeline/pipeline_adapter.go
```diff
@@ -398,8 +400,9 @@ func (a *Adapter) prepareSnapshotForPipelineRun(pipelineRun *tektonv1beta1.Pipel
 	}
 	pipelinerunAnnotations := pipelineRun.GetAnnotations()
 	for key, value := range pipelinerunAnnotations {
-		if strings.Contains(key, "pipelinesascode.tekton.dev") {
-			snapshot.Annotations[key] = value
+		if strings.Contains(key, "pipelinesascode.tekton.dev/") {
+			newKey := strings.ReplaceAll(key, "pipelinesascode.tekton.dev/", "pac.test.appstudio.openshift.io/")
+			snapshot.Annotations[newKey] = value
```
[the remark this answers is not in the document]
amisstea: Helpers functions have been added to copy labels and annotations. If you have any great ideas for how to combine the two, please share.

controllers/pipeline/pipeline_adapter_test.go
parasense: :+1:

helpers/metadata.go
```diff
@@ -33,3 +38,37 @@ func HasLabelWithValue(object client.Object, label, value string) bool {
 
 	return false
 }
+
+// CopyLabelsByPrefix copies all labels from a source object to a destination object where the key matches the specified prefix.
+// If replacementPrefix is different from prefix, the prefix will be replaced while performing the copy.
+func CopyLabelsByPrefix(src *metav1.ObjectMeta, dest *metav1.ObjectMeta, prefix string, replacementPrefix string) {
```
parasense, line 44: Very nice!

One could potentially simplify the function signature, like this:
`func CopyLabelsByPrefix(src , dest *metav1.ObjectMeta, prefix, replacementPrefix string) {`

helpers/metadata.go
```diff
@@ -33,3 +38,37 @@ func HasLabelWithValue(object client.Object, label, value string) bool {
...
+	if dest.Labels == nil {
+		dest.Labels = make(map[string]string)
```
parasense, line 46: Maybe consider using the object's Get & Set methods rather than going directly.
Something like this...
```
if dest.GetLabels() == nil {
		dest.SetLabels(map[string]string{})
```
Note: please consider this an _**optional**_ suggestion, because it's highly pedantic, and purely subjective. Accessing exported maps is fine, even if the maps have getter's and setter's.

amisstea: I applied your suggestion and added an extra `nil` check to each helper in the process.

helpers/metadata.go
amisstea: Done.

helpers/metadata.go
```diff
@@ -33,3 +38,37 @@ func HasLabelWithValue(object client.Object, label, value string) bool {
...
+	}
+	for key, value := range src.GetLabels() {
+		if strings.HasPrefix(key, prefix) {
+			newKey := key
+			if prefix != replacementPrefix {
+				newKey = strings.Replace(key, prefix, replacementPrefix, 1)
+			}
+			dest.Labels[newKey] = value
+		}
+	}
```
parasense, line 59: Moving this code into a helper function was a good move, it makes the pipeline_adapter.go look clean. :smiley: 
Now consider using another function to do the things common to both **Labels** and **Annotations**.

Something like this:
```
// copyWithNewPrefix copies key/value pairs in the source map adding them into the destination map.
func copyWithNewPrefix(src, dest map[string]string, prefix, replacementPrefix string) {
	for key, value := range src {
		if strings.HasPrefix(key, prefix) {
			newKey := key
			if prefix != replacementPrefix {
				newKey = strings.Replace(key, prefix, replacementPrefix, 1)
			}
			dest[newKey] = value
		}
	}
}
```

Then you can make both these functions a wrapper for the underlying function. Like so...
```
// CopyAnnotationsByPrefix copies all annotations from a source object to a destination object where the key matches the specified prefix.
// If replacementPrefix is different from prefix, the prefix will be replaced while performing the copy.
func CopyAnnotationsByPrefix(src, dest *metav1.ObjectMeta, prefix, replacementPrefix string) {
	if dest.GetAnnotations() == nil {
		dest.SetAnnotations(map[string]string{})
	}
	copyWithNewPrefix(src.GetAnnotations(), dest.GetAnnotations(), prefix, replacementPrefix)
}
```
That way you only have to write the boilerplate code once, test in a single place at the helper function, and the two wrapper function only need to test they created maps if they are nil, and only operate on the given **Label** or **Annotation**.

amisstea: Thanks for the suggestion! Applied.

jinqi7 (approved):

PR #75, opened 2022-11-29
feat: use operator-goodies test package to get external CRDs

davidmogar:
Now that the release-service is removing the test package and moving it to the operator-goodies library, the integration-service has to do the same to avoid broken imports.

In addition to this change, tekton CRDs are fetched using the GetRelativeDependencyPath function as well. This way the dependency version doesn't have to be hard-coded in the code.


dirgim (approved):
LGTM

jsztuka (approved):
lgtm

PR #76, opened 2022-11-29
fix(HACBS-1434): skip snapshot creation if component is unknown

dirgim:
* Add the skip condition and log to EnsureSnapshotExists
* Add more coverage to pipeline_adapter unit tests

### DIFF
--- a/controllers/pipeline/pipeline_adapter.go
+++ b/controllers/pipeline/pipeline_adapter.go
@@ -67,6 +67,11 @@ func (a *Adapter) EnsureSnapshotExists() (results.OperationResult, error) {
 		return results.ContinueProcessing()
 	}
 
+	if a.component == nil {
+		a.logger.Info("The pipelineRun does not have any component associated with it, will not create a new Snapshot.")
+		return results.ContinueProcessing()
+	}
+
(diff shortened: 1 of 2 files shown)


jsztuka (approved):
lgtm

hongweiliu17 (approved):
LGTM

controllers/pipeline/pipeline_adapter.go
```diff
@@ -67,6 +67,11 @@ func (a *Adapter) EnsureSnapshotExists() (results.OperationResult, error) {
 		return results.ContinueProcessing()
 	}
 
+	if a.component == nil {
```
hongweiliu17, line 70: I am just learning to debug it from you, this is to stop the later process because a component is not associated because of some unknown reason, right? And what the user will see for this type of non-associated component?

dirgim: Yes, basically this means that the build pipeline doesn't have a component associated with it (by way of having a label that specifies it). In such cases it is not possible to construct a new Snapshot in this function so we output a log message and ignore that pipeline.

hongweiliu17: I see now, thank you!

PR #84, opened 2022-12-26
feat(HACBS-1407): Set the snapshot upon starting and finishing Integration testing

hongweiliu17:
Set the snapshot integration status condition upon starting and finishing Integration testing

Signed-off-by: Hongwei Liu hongliu@redhat.com

### DIFF
--- a/gitops/snapshot.go
+++ b/gitops/snapshot.go
@@ -76,6 +76,12 @@ const (
 
 	// HACBSIntegrationStatusInvalid is the reason that's set when the HACBS integration gets into an invalid state.
 	HACBSIntegrationStatusInvalid = "Invalid"
+
+	//HACBSIntegrationStatusInProgress is the reason that's set when the HACBS tests gets into an in progress state.
+	HACBSIntegrationStatusInProgress = "InProgress"
+
+	//HACBSIntegrationStatusFinished is the reason that's set when the HACBS tests finish.
+	HACBSIntegrationStatusFinished = "Finished"
 )
 
 // MarkSnapshotAsPassed updates the HACBS Test succeeded condition for the Snapshot to passed.
@@ -88,6 +94,7 @@ func MarkSnapshotAsPassed(adapterClient client.Client, ctx context.Context, snap
 		Reason:  HACBSTestSuceededConditionPassed,
 		Message: message,
 	})
+	SetSnapshotIntegrationStatusAsFinished(snapshot, "Marking snapshot integration status condition as finished since the testing is passed")
 	err := adapterClient.Status().Patch(ctx, snapshot, patch)
 	if err != nil {
 		return nil, err
@@ -105,6 +112,7 @@ func MarkSnapshotAsFailed(adapterClient client.Client, ctx context.Context, snap
 		Reason:  HACBSTestSuceededConditionFailed,
 		Message: message,
 	})
+	SetSnapshotIntegrationStatusAsFinished(snapshot, "Marking snapshot integration status condition as finished since the testing fails")
 	err := adapterClient.Status().Patch(ctx, snapshot, patch)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


gitops/snapshot.go
```diff
@@ -122,6 +128,26 @@ func SetSnapshotIntegrationStatusAsInvalid(snapshot *applicationapiv1alpha1.Snap
...
+		Reason:  HACBSIntegrationStatusInProgress,
+		Message: message,
+	})
+}
+
+// SetSnapshotIntegrationStatusAsFinished sets the HACBS integration status condition for the Snapshot to Finished.
+func SetSnapshotIntegrationStatusAsFinished(snapshot *applicationapiv1alpha1.Snapshot, message string) {
+	meta.SetStatusCondition(&snapshot.Status.Conditions, metav1.Condition{
+		Type:    HACBSIntegrationStatusCondition,
+		Status:  metav1.ConditionFalse,
```
dirgim, line 145: I think we might want to set the status to `metav1.ConditionTrue` here since this is the expected end goal of this condition.

gitops/snapshot.go
```diff
@@ -122,6 +128,26 @@ func SetSnapshotIntegrationStatusAsInvalid(snapshot *applicationapiv1alpha1.Snap
 	})
 }
 
+// SetSnapshotIntegrationStatusAsInProgress sets the HACBS integration status condition for the Snapshot to In Progress.
+func SetSnapshotIntegrationStatusAsInProgress(snapshot *applicationapiv1alpha1.Snapshot, message string) {
```
dirgim, line 132: Since we will be using these functions in a different manner from `SetSnapshotIntegrationStatusAsInvalid`, I believe that we will need to make these functions also patch the resource on the cluster, similar to [MarkSnapshotAsPassed](https://github.com/redhat-appstudio/integration-service/blob/main/gitops/snapshot.go#L84-L94).

gitops/snapshot.go
hongweiliu17: Thanks. So `metav1.ConditionTrue` means it is expected while `metav1.ConditionFalse` means it is unexpected? I searched the `meta` but I didn't find when we should use `True` or `False`. Thanks!

hongweiliu17: @dirgim I found these help comments in `vendor` dir, is it?
```
// These are valid condition statuses. "ConditionTrue" means a resource is in the condition.
// "ConditionFalse" means a resource is not in the condition. "ConditionUnknown" means kubernetes
// can't decide if a resource is in the condition or not. In the future, we could add other
// intermediate conditions, e.g. ConditionDegraded.
```

dirgim: Yes, I think we're on the right track - setting the `ConditionTrue` when the integration service successfully processes the snapshot seem to be in line with the guidance.

gitops/snapshot.go
```diff
@@ -122,6 +128,26 @@ func SetSnapshotIntegrationStatusAsInvalid(snapshot *applicationapiv1alpha1.Snap
...
+	meta.SetStatusCondition(&snapshot.Status.Conditions, metav1.Condition{
+		Type:    HACBSIntegrationStatusCondition,
+		Status:  metav1.ConditionFalse,
```
dirgim, line 135: In light of the [guidance in the meta help](https://github.com/redhat-appstudio/integration-service/pull/84/files#r1060320035), maybe setting this to `metav1.ConditionUnknown` would make more sense?

gitops/snapshot.go
hongweiliu17: Updated. Thanks.

gitops/snapshot.go
dirgim: Thanks for the update here!

gitops/snapshot.go
hongweiliu17: Updated. Thanks.

hongweiliu17:
@dirgim Thanks for the detailed comment. I revised the code, could you review it again? Thanks!

dirgim (approved):
LGTM

jsztuka (approved):
lgtm

jinqi7 (approved):
LGTM

PR #90, opened 2023-01-10
fix(HACBS-1576): fix composite snapshot checking

dirgim:
Follow-on fix for a edge case error that happens if the component can't be found (because of a label mismatch caused by HACBS-1576 or an issue with manual/e2e testing) when the pipeline controller runs the composite snapshot check.

* Check for component existence before running composite logic
* Use the HasLabelWithValue to check Snapshot type


MartinBasti (approved):
LGTM

Josh-Everett (approved):
+1

PR #93, opened 2023-01-18
fix(HACBS-1503): create controller for IntegrationTestScenario

jsztuka:
Signed-off-by: Jiri Sztuka <jsztuka@redhat.com>
Controller for following cases:

- Set owner reference for integrationTestScenarios which have application defined and application exists (if the application does not exist - scenario is marked as INVALID)
- check for existence of environment within IntegrationTestScenario (if it exists, but the environment is in different namespace - mark ITS as INVALID)


controllers/scenario/scenario_adapter.go
```diff
@@ -0,0 +1,146 @@
...
+			return reconciler.RequeueWithError(err)
+		}
+		return reconciler.ContinueProcessing()
+	}
+
+}
+
+// EnsureGlobalScenarioUpdated is an operation that ensure the IntegrationTestScenario in the Global Candidate List
+// being updated when the IntegrationTestScenario(ITS) gets detected, set owner reference only for valid ITS
+func (a *Adapter) EnsureGlobalScenarioUpdated() (reconciler.OperationResult, error) {
```
dirgim, line 82: I'm not sure what this function is supposed to do, the IntegrationTestScenarios are not supposed to have multiple owner references, or to have global candidates.

controllers/scenario/scenario_adapter.go
```diff
@@ -0,0 +1,146 @@
...
+		logger:      logger,
+		client:      client,
+		context:     context,
+	}
+}
+
+// EnsureCreatedScenarioIsValid is an operation that ensures created IntegrationTestScenario is valid
+// in case it is, set its owner reference
+func (a *Adapter) EnsureCreatedScenarioIsValid() (reconciler.OperationResult, error) {
+	if meta.IsStatusConditionTrue(a.scenario.Status.Conditions, gitops.HACBSIntegrationStatusCondition) {
```
dirgim, line 57: I think we first want to check if the `a.scenario` is valid by checking if `a.application` is nil. If it's not nil we should then check if the owner reference is set - if it's not set already, we should try to set the owner reference.
If any of the actions from above fail, we should then check if `a.scenario` has already been marked as invalid (with this check) - if it hasn't been already marked as invalid, only then the update should be made to the scenario status.

Let say the user puts the wrong application name into the ITS. We then mark it as invalid in this function. Once they realize that that they put the wrong application name, they try to fix it. We should then re-check the ITS and either leave the invalid status (by **not** updating the ITS resource, otherwise we'd end up in a loop) or change the status to valid (true).

controllers/scenario/scenario_adapter.go
```diff
@@ -0,0 +1,146 @@
...
+
+	}
+
+	return reconciler.ContinueProcessing()
+}
+
+// EnsureEnvironmentExists checks if ITS has environment object defined
+// in case it does > makes sure it exists within the same namespace as ITS
+// otherwise inform about opposite fact.
+func (a *Adapter) EnsureEnvironmentExists() (reconciler.OperationResult, error) {
```
dirgim, line 104: After thinking about this, I think this function should be merged with `EnsureCreatedScenarioIsValid` since we're dealing with the same scenario and updating the same status which would be problematic to do in parallel - e.g. if one function decides that the scenario is fine, while the other one finds a problem and they both try to update the scenario status at nearly the same time.

controllers/scenario/scenario_adapter.go
jsztuka: Yep this has no role/purpose in this context, removing this one.

controllers/scenario/scenario_adapter.go
jsztuka: Merged with EnsureCreatedScenarioIsValid, dropped this ensure function at the same time.

controllers/scenario/scenario_controller.go
```diff
@@ -0,0 +1,118 @@
...
+// NewScenarioReconciler creates and returns a Reconciler.
+func NewScenarioReconciler(client client.Client, logger *logr.Logger, scheme *runtime.Scheme) *Reconciler {
+	return &Reconciler{
+		Client: client,
+		Log:    logger.WithName("integrationTestScenario"),
+		Scheme: scheme,
+	}
+}
+
+//+kubebuilder:rbac:groups=appstudio.redhat.com,resources=snapshots,verbs=get;list;watch;create;update;patch;delete
```
dirgim, line 50: I think that these resources will need to be change to match the behavior of the controller - I think only integrationtestscenarios, applications and environments need to be specified here.

controllers/scenario/scenario_adapter.go
```diff
@@ -0,0 +1,122 @@
...
+		}
+		err = a.client.Patch(a.context, a.scenario, patch)
+		if err != nil {
+			a.logger.Error(err, "Failed to update Scenario",
+				"Scenario.Name", a.scenario.Name)
+			return reconciler.RequeueWithError(err)
+		}
+		return reconciler.ContinueProcessing()
+		// app exists, owner reference is set, check if the scenario has envirnment defined
+	} else if reflect.ValueOf(a.scenario.Spec.Environment).IsZero() {
```
dirgim, line 86: I think we'll need to decouple these `if` conditions because we want to check the scenario for multiple things in the same reconcile pass.

controllers/scenario/scenario_adapter.go
```diff
@@ -0,0 +1,122 @@
...
+			a.logger.Info("Environment doesn't exist in same namespace as IntegrationTestScenario.",
+				"Scenario.Name:", a.scenario.Name,
+				"Scenario.Namespace:", a.scenario.Namespace,
+				"Environment.Name:", a.scenario.Spec.Environment.Name)
+			patch := client.MergeFrom(a.scenario.DeepCopy())
+			SetScenarioIntegrationStatusAsInvalid(a.scenario, "Environment "+a.scenario.Spec.Environment.Name+" is located in different namespace than scenario.")
+			return reconciler.RequeueOnErrorOrStop(a.client.Status().Patch(a.context, a.scenario, patch))
+		}
+
+	}
```
dirgim, line 110: I think it would be good to update the IntegrationTestScenario status condition to valid here - if all of the above tests passed and if it wasn't already valid.

controllers/scenario/scenario_adapter.go
```diff
@@ -0,0 +1,122 @@
...
+	return reconciler.ContinueProcessing()
+}
+
+// SetSnapshotIntegrationStatusAsInvalid sets the HACBS integration status condition for the Snapshot to invalid.
+func SetScenarioIntegrationStatusAsInvalid(scenario *v1alpha1.IntegrationTestScenario, message string) {
+	meta.SetStatusCondition(&scenario.Status.Conditions, metav1.Condition{
+		Type:    gitops.HACBSIntegrationStatusCondition,
```
dirgim, line 117: I think we'll need to use different constants here since the `IntegrationStatusCondition` could be a bit misleading. Maybe a new constant `IntegrationTestScenarioValid` or something similar to that?

controllers/scenario/scenario_adapter.go
jsztuka: Decoupled if statements.

controllers/scenario/scenario_adapter.go
jsztuka: Added, now the logic corresponds to this:
Valid scenario must meet those conditions:
- Has ENV specified
- ENV is situated within scenario namespace
- Has App defined

As a bonus it has owner reference added.

gitops/snapshot.go
```diff
@@ -65,6 +66,9 @@ const (
 	// HACBSIntegrationStatusCondition is the condition for marking the HACBS integration status of the Snapshot.
 	HACBSIntegrationStatusCondition = "HACBSIntegrationStatus"
 
+	// HACBSIntegrationTestScenarioValid is the condition for marking the HACBS integration status of the Scenario.
+	HACBSIntegrationTestScenarioValid = "HACBSIntegrationTestScenario"
```
dirgim, line 70: nit: The naming convention for [conditions](https://maelvls.dev/kubernetes-conditions/) should read like a boolean, e.g. [PodScheduled](https://github.com/kubernetes/kubernetes/blob/beaf3a2f0/pkg/scheduler/scheduler.go#L435), [ContainersReady](https://github.com/kubernetes/kubernetes/blob/beaf3a2f0/pkg/kubelet/kubelet_pods.go#L1370) etc. We can also skip the `HACBS` prefix since our workflow is no longer HACBS specific.
Something like `IntegrationTestScenarioValid` still makes the most sense to me.

controllers/scenario/scenario_adapter.go
```diff
@@ -0,0 +1,164 @@
...
+			if err != nil {
+				a.logger.Error(err, "Failed to update Scenario",
+					"Scenario.Name", a.scenario.Name)
+				return reconciler.RequeueWithError(err)
+			}
+			return reconciler.ContinueProcessing()
+		}
+
+	}
+	if reflect.ValueOf(a.scenario.Status).IsZero() {
```
dirgim, line 131: This condition only seems to kick-in when the IntegrationTestScenario status is not set. I think we want to check if the `IntegrationTestScenarioValid` condition is set to false instead. Here's an example of how we check the [Snapshot status condition.](https://github.com/redhat-appstudio/integration-service/blob/main/gitops/snapshot.go#L168)

jsztuka: Turned out that valid ITS was not handled properly because of too many ` return reconciler.ContinueProcessing()` statements. It gets properly handled now.

gitops/snapshot.go
jsztuka: Modified variable name together with its value.

controllers/scenario/scenario_adapter.go
```diff
@@ -0,0 +1,164 @@
...
+			return reconciler.RequeueWithError(err)
+		}
+
+	}
+	// Checks if scenario has environment defined
+	if reflect.ValueOf(a.scenario.Spec.Environment).IsZero() {
+		a.logger.Info("IntegrationTestScenario has no environment defined",
+			"Scenario.Name:", a.scenario.Name,
+			"Scenario.Namespace", a.scenario.Namespace)
+		return reconciler.ContinueProcessing()
```
dirgim, line 104: I think that this `ContinueProcessing` call doesn't need to happen since we still want to mark the Snapshot as valid later in the function even if the Environment isn't defined.

controllers/scenario/scenario_adapter.go
```diff
@@ -0,0 +1,164 @@
...
+	// Checks if scenario has ownerReference assigned to it
+	if a.scenario.OwnerReferences == nil {
+		patch := client.MergeFrom(a.scenario.DeepCopy())
+		err := ctrl.SetControllerReference(a.application, a.scenario, a.client.Scheme())
+		if err != nil {
+			a.logger.Error(err, "Error setting owner reference.",
+				"Application.Name", a.application.Name,
+				"Application.Namespace", a.application.Namespace)
+			patch := client.MergeFrom(a.scenario.DeepCopy())
+			SetScenarioIntegrationStatusAsInvalid(a.scenario, "Failed to get application for scenario.")
```
dirgim, line 82: I don't think we want to set the scenario as invalid because we failed when setting the owner reference, that part's our fault, so logging the issue and calling the `RequeueWithError` should be enough.

controllers/scenario/scenario_adapter.go
jsztuka: Agreed.

controllers/scenario/scenario_adapter.go
jsztuka: Agreed.

dirgim (approved):
LGTM

controllers/scenario/scenario_adapter.go
```diff
@@ -0,0 +1,155 @@
...
+		return reconciler.ContinueProcessing()
+	}
+	// Checks if scenario has ownerReference assigned to it
+	if a.scenario.OwnerReferences == nil {
+		patch := client.MergeFrom(a.scenario.DeepCopy())
+		err := ctrl.SetControllerReference(a.application, a.scenario, a.client.Scheme())
+		if err != nil {
+			a.logger.Error(err, "Error setting owner reference.",
+				"Application.Name", a.application.Name,
+				"Application.Namespace", a.application.Namespace)
```
MartinBasti, line 80: Logger messages are inconsistent across this patch. There are two formats: 
a) `"Application.Name", a.application.Name,`
b) `"Application.Name:", a.application.Name,`

I'd prefer unified format, preferably b)

jsztuka: "Unified"

MartinBasti (approved):

PR #95, opened 2023-02-02
Update test coverage

MartinBasti:


controllers/pipeline/pipeline_controller_test.go
```diff
@@ -97,6 +97,7 @@ var _ = Describe("PipelineController", func() {
 					"pipelines.openshift.io/runtime":        "nodejs",
 					"pipelines.openshift.io/strategy":       "s2i",
 					"appstudio.openshift.io/component":      "component-sample",
+					"appstudio.openshift.io/application":    applicationName,
```
MartinBasti, line 100: I don't know if I can add this to the current pipeline, must it be separate pipeline ?

controllers/scenario/scenario_adapter_test.go
```diff
@@ -144,6 +147,35 @@ var _ = Describe("Scenario Adapter", Ordered, func() {
...
+		AfterEach(func() {
+			envNamespace = DefaultNamespace
+		})
+
+		It("ensure the scenario status is invalid", func() {
+			Eventually(func() bool {
+				result, err := adapter.EnsureCreatedScenarioIsValid()
+				return !result.CancelRequest && err == nil
+			}, time.Second*10).Should(BeTrue())
+			Expect(meta.IsStatusConditionFalse(integrationTestScenario.Status.Conditions, gitops.IntegrationTestScenarioValid)).To(BeTrue())
```
MartinBasti, line 174: I cannot figure out, why this test is failing, it's supposed to be Invalid

controllers/pipeline/pipeline_controller_test.go
jsztuka: This interpretation is fine.

controllers/scenario/scenario_adapter_test.go
dirgim: I think it's because the [env](https://github.com/redhat-appstudio/integration-service/pull/95/files#diff-2ea0168978a368c1f4520a42d682ce29f826a4e0014a6dc573a4a2ef1964db4dR98) is created only in the `BeforeAll` function and is not modified or recreated before every test.

MartinBasti: I updated test to utilize JustBeforeEach, PTAL, it's little bit magic, but now it works :D

Josh-Everett (approved):
lgtm

controllers/scenario/scenario_adapter_test.go
dirgim: Yes, this looks like the way to go

MartinBasti:
Coverage here decreased by -4.30%, I'll check what I did wrong and how to get it back (most likely adding application label into pipeline run)
```
controllers/snapshot/snapshot_adapter.go 11 48.68% - -4.30%
```

controllers/pipeline/pipeline_adapter_test.go
```diff
@@ -462,6 +462,27 @@ var _ = Describe("Pipeline Adapter", Ordered, func() {
...
+			snapshot, err := adapter.prepareSnapshotForPipelineRun(adapter.pipelineRun, adapter.component, adapter.application)
+			Expect(err).To(BeNil())
+
+			err = adapter.client.Create(adapter.context, snapshot)
+			Expect(err).To(BeNil())
+		})
+
+		It("ensures snapshot creation is skipped when snapshot already exists", func() {
+			Eventually(func() bool {
+				adapter.EnsureSnapshotExists()
```
hongweiliu17, line 479: I don't understand why this `adapter.EnsureSnapshotExists()` is called twice, could you explain? Thanks!

MartinBasti: my bad, removed. Nice catch!

jsztuka (approved):
lgtm

controllers/pipeline/pipeline_adapter_test.go
hongweiliu17: Thanks!

hongweiliu17:
The code looks good to me, thanks! I think the AC of [HACBS-1535](https://issues.redhat.com/browse/HACBS-1535) is `Increase unit test coverage on the individual directories within integration service to be above 70%`, so I feel we can add some tests in [HACBS-1535](https://issues.redhat.com/browse/HACBS-1535) to make the code coverage of integration-service higher than 70% then the later PRs should make sure they won't lower the code coverage and we don't need to file the similar ticket to increase the code coverage any more. Is it? @jsztuka since you file [HACBS-1535](https://issues.redhat.com/browse/HACBS-1535) and add AC.

MartinBasti:
yes, I'll add more tests. This is just my first try so I wanted early feedback if my approach is correct. I'll open a new PR for more test cases

hongweiliu17 (approved):
LGTM

PR #115, opened 2023-02-21
Prevent out-of-order deployments

MartinBasti:
Fixing race condition where older pipeline could replace newer
    deployment with image from older pipeline, if older pipeline was running
    longer than latest one.


jsztuka (approved):
Well done, LGTM

hongweiliu17 (approved):
LGTM, thanks!

PR #133, opened 2023-03-15
feat(STONEINTG-291): Deny egress connections

MartinBasti:
This is for security hardening, integration-service is not supposed to be communicating with the internet, only kube-api should be allowed.

Note: this also blocks DNS resolving


dirgim (approved):
LGTM

PR #135, opened 2023-03-23
feat(STONEINTG-345): add component git info to Snapshot

hongweiliu17:
* Add component git info to Snapshot built from pipelinerun
* Add component git info to Snapshot built from testedSnapshot


hongweiliu17:
I am trying to use `Revision` in [gitSource](https://github.com/redhat-appstudio/application-api/blob/main/api/v1alpha1/component_types.go#L42) to save 'Commit' in the draft then I can update it after https://github.com/redhat-appstudio/application-api/pull/39 is merged. But the unit test fails now at https://github.com/redhat-appstudio/integration-service/actions/runs/4498338339/jobs/7921639600?pr=135#step:18:44. Function `existingCompositeSnapshot, err := adapter.createCompositeSnapshotsIfConflictExists(hasApp, hasComp, createdSnapshot)` is called twice, the expected result is it creates one compositeSnapshot and the second found the existing compositeSnapshot, but in my code it creates two compositeSnapshots, I compare the two compositeSnapshots, they are almost the same except for `Snapshot.Name`. Don't know why.

controllers/pipeline/pipeline_adapter.go
```diff
@@ -278,6 +301,23 @@ func (a *Adapter) getImagePullSpecFromSnapshotComponent(snapshot *applicationapi
...
+			return &snapshotComponent.Source, nil
+		}
+	}
+	return nil, fmt.Errorf("couldn't find the requested component source info in the given Snapshot")
+}
+
+// getComponentSourceWithLastBuildCommit gets the component source from the given Component with LastBuiltCommit as Revision
+func (a *Adapter) getComponentSourceWithLastBuildCommit(component *applicationapiv1alpha1.Component) *applicationapiv1alpha1.ComponentSource {
+	componentSource := component.Spec.Source.DeepCopy()
+	componentSource.GitSource.Revision = component.Status.LastBuiltCommit
```
dirgim, line 317: There is a possibility that the `component.Status` doesn't have the `LastBuiltCommit` defined. I think we need to handle that case.

hongweiliu17:
> Once this is merged (and released to prod and all that), I should expect that the snapshot components will contain `source.git.revision`, correct?

Yes.

controllers/pipeline/pipeline_adapter.go
hongweiliu17: Thanks. Add code to handle the missing `LastBuiltCommit`.

tekton/utils_test.go
```diff
@@ -63,7 +73,22 @@ var _ = Describe("Utils", func() {
...
+		klog.Infoln("Got expected git_url")
 	})
 
+	It("can get git-commit", func() {
+		commit, _ := tekton.GetComponentSourceGitCommit(pipelineRun)
+		if commit != "a2ba645d50e471d5f084b" {
+			Fail(fmt.Sprintf("Expected commit is a2ba645d50e471d5f084b, but got %s", commit))
+		}
+		klog.Infoln("Got expected commit")
+	})
```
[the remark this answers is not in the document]
hongweiliu17: Added. Thanks.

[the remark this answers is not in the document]
dirgim: Thanks!

controllers/pipeline/pipeline_adapter.go
dirgim: Thanks, I made a follow-up comment on the updated code.

controllers/pipeline/pipeline_adapter.go
```diff
@@ -278,6 +301,26 @@ func (a *Adapter) getImagePullSpecFromSnapshotComponent(snapshot *applicationapi
...
+}
+
+// getComponentSourceWithLastBuildCommit gets the component source from the given Component with LastBuiltCommit as Revision
+func (a *Adapter) getComponentSourceWithLastBuildCommit(component *applicationapiv1alpha1.Component) (*applicationapiv1alpha1.ComponentSource, error) {
+	if component.Status.LastBuiltCommit != "" {
+		componentSource := component.Spec.Source.DeepCopy()
+		componentSource.GitSource.Revision = component.Status.LastBuiltCommit
+		return componentSource, nil
+	}
+	return nil, fmt.Errorf("couldn't find the requested .Status.LastBuiltCommit info in the given SnapshotComponent")
```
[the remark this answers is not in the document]
hongweiliu17: Yes. So let me rename the function name to `getComponentSourceFromComponent`. Thanks.

controllers/pipeline/pipeline_adapter.go
```diff
@@ -278,6 +301,26 @@ func (a *Adapter) getImagePullSpecFromSnapshotComponent(snapshot *applicationapi
...
+}
+
+// getComponentSourceFromComponent gets the component source from the given Component as Revision
+// and set Component.Status.LastBuiltCommit as Component.Source.GitSource.Revision if it is defined.
+func (a *Adapter) getComponentSourceFromComponent(component *applicationapiv1alpha1.Component) *applicationapiv1alpha1.ComponentSource {
+	componentSource := component.Spec.Source.DeepCopy()
+	if component.Status.LastBuiltCommit != "" {
+		componentSource.GitSource.Revision = component.Status.LastBuiltCommit
+	}
+	return componentSource
```
hongweiliu17, line 321: @dirgim Should we return an error here when Revision is not found in componentSource?

dirgim: Not completely sure, the [Revision](https://github.com/redhat-appstudio/application-api/blob/main/api/v1alpha1/component_types.go#L39-L42) field is optional. My gut feeling is that we shouldn't return an error since the `componentSource` is still valid even without that field.

controllers/pipeline/pipeline_adapter.go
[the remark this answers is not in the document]
dirgim: Ack, the change looks good.

dirgim:
> Sorry, I think we need to also implement [STONEINTG-354](https://issues.redhat.com/browse/STONEINTG-354) to update the commit in global candidate list. Right? @dirgim

With this change the source field with `Revision` will already start showing up in application Snapshots, but all components other than the newly built one will not have their latest built commit there. The follow-up STONEINTG-354 will ensure that.
@johnbieren so in conclusion, it depends if this change is enough for your related PR.

dirgim (approved):
LGTM

johnbieren:
> > Sorry, I think we need to also implement [STONEINTG-354](https://issues.redhat.com/browse/STONEINTG-354) to update the commit in global candidate list. Right? @dirgim
> 
> With this change the source field with `Revision` will already start showing up in application Snapshots, but components other than the newly built one will not have their latest built commit there. The follow-up [STONEINTG-354](https://issues.redhat.com//browse/STONEINTG-354) will ensure that. @johnbieren so in conclusion, it depends if this change is enough for your related PR.

Thanks. We treat all components the same, so I will wait for 354. I created myself a ticket to change our parameter default from false to true (thus turning on the use of the revision stuff) and marked it blocked by 354

controllers/pipeline/pipeline_adapter.go
dirgim: Resolving this discussion for now so it doesn't get in the way of the review process.

controllers/pipeline/pipeline_adapter.go
```diff
@@ -350,21 +393,29 @@ func (a *Adapter) getAllPipelineRunsForSnapshot(snapshot *applicationapiv1alpha1
...
-func (a *Adapter) prepareSnapshot(application *applicationapiv1alpha1.Application, component *applicationapiv1alpha1.Component, newContainerImage string) (*applicationapiv1alpha1.Snapshot, error) {
+func (a *Adapter) prepareSnapshot(application *applicationapiv1alpha1.Application, component *applicationapiv1alpha1.Component, newContainerImage string, newComponentSource *applicationapiv1alpha1.ComponentSource) (*applicationapiv1alpha1.Snapshot, error) {
 	applicationComponents, err := a.getAllApplicationComponents(application)
 	if err != nil {
 		return nil, fmt.Errorf("failed to get all Application Components for Application %s", a.application.Name)
 	}
 
 	var snapshotComponents []applicationapiv1alpha1.SnapshotComponent
 	for _, applicationComponent := range *applicationComponents {
+		applicationComponent := applicationComponent
```
MartinBasti, line 404: as go newbie, why is this assignment needed?

controllers/pipeline/pipeline_adapter.go
```diff
@@ -350,21 +393,29 @@ func (a *Adapter) getAllPipelineRunsForSnapshot(snapshot *applicationapiv1alpha1
...
 		containerImage := applicationComponent.Spec.ContainerImage
+
+		var componentSource *applicationapiv1alpha1.ComponentSource
 		if applicationComponent.Name == component.Name {
 			containerImage = newContainerImage
+			componentSource = newComponentSource
+		} else {
+			// Get ComponentSource for the component which is not built in this pipeline
+			componentSource = a.getComponentSourceFromComponent(&applicationComponent)
```
MartinBasti, line 413: I'm puzzled with logic here, why are we passing `newComponentSource` when we can always get Source from component using  getComponentSourceFromComponent?

This also feel like violation of dependency injection pattern, you are passing dependency via param but at the same time for other components you are doing dynamic fething

controllers/pipeline/pipeline_adapter.go
```diff
@@ -385,8 +436,12 @@ func (a *Adapter) prepareSnapshotForPipelineRun(pipelineRun *tektonv1beta1.Pipel
 	if err != nil {
 		return nil, err
 	}
+	componentSource, err := a.getComponentSourceFromPipelineRun(pipelineRun)
```
MartinBasti, line 439: does this mean, that component doesn't always contain sources, or why there are 2 methods, getting source from component and getting source from pipelinerun?

controllers/pipeline/pipeline_adapter.go
MartinBasti: Resolving, I got explanation of component workflow offline

controllers/pipeline/pipeline_adapter.go
MartinBasti: was explained to me offline. Yes we need to fetch data from pipeline, resolving

MartinBasti (approved):

PR #139, opened 2023-03-31
Add new version v1beta1 to integrationtestscenario to support Tekton Resolver

hongweiliu17:
* Add a new api version v1beta1 and make v1alpha1 as storage version because we need update more code to enable v1beta1. After the conversion webhook is implemented, v1beta1 will be set as storage version.

* Add ResolverRef with Resolver(String) and Params([]tektonv1beta1.Param) under IntegrationTestScenario.Spec

The expected IntegrationTestScenario is like below:
```
apiVersion: appstudio.redhat.com/v1beta1
kind: IntegrationTestScenario
metadata:
  name: example-pass
  namespace: default
spec:
  application: application-sample
  resolverRef:
    resolver: git
    params:
      - name: url
        value: https://url
      - name: branch
        value: main
      - name: pathInRepo
        value: pipelines/helloworld.yaml
```


api/v1beta1/integrationtestscenario_types.go
```diff
@@ -0,0 +1,92 @@
...
+
+// IntegrationTestScenarioSpec defines the desired state of IntegrationScenario
+type IntegrationTestScenarioSpec struct {
+	// Application that's associated with the IntegrationTestScenario
+	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
+	// +required
+	Application string `json:"application"`
+	// Tekton Resolver where to store the Tekton resolverRef trigger Tekton pipeline used to refer to a Pipeline or Task in a remote location like a git repo.
+	// +required
+	ResolverRef tektonv1beta1.ResolverRef `json:"resolverRef"`
```
adambkaplan, line 33: General word of advice when it comes to APIs - don't reference other project CRDs (objects and sub-objects). APIs can change in unexpected ways, especially if they are not labeled `v1`. You are better off defining your own object here, using only the fields you care about. This advice also applies to k8s objects.

hongweiliu17: Thanks. Let's use our own object. That works, but it is strange why referencing tekton.ResolverRef will fails in unittest.

controllers/pipeline/pipeline_controller_test.go
```diff
@@ -262,6 +262,7 @@ var _ = Describe("PipelineController", func() {
 					},
 				},
 			}
+
```
jinqi7, line 265: Is it expected?

hongweiliu17: No. Thanks, it is added unintentionally.

api/v1alpha1/integrationtestscenario_types.go
```diff
@@ -69,6 +69,7 @@ type TestContext struct {
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
 // +kubebuilder:printcolumn:name="Application",type=string,JSONPath=`.spec.application`
+// +kubebuilder:storageversion
```
jsztuka, line 72: I have the feeling that we want to store the version that we want to migrate to, I suppose `v1beta1`

hongweiliu17: Yes, `v1beta1` will be the storage version finally, but before we have the conversion to convert them and finish the controller update, we have  to use v1alpha1 as the storage version.

jsztuka (approved):
LGTM, great work.

codecov.yml
```diff
@@ -1,6 +1,7 @@
 # Documentation: https://docs.codecov.io/docs/codecov-yaml
 ignore:
   - "api/v1alpha1/zz_generated.deepcopy.go" # generated file
+  - "api/v2alpha1/zz_generated.deepcopy.go" # generated file
```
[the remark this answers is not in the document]
hongweiliu17: Ah, yes, thanks!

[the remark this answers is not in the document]
dirgim: Thanks!

PR #140, opened 2023-04-03
feat(STONEINTG-202): Create a JSON schema for the standardized test output

Josh-Everett:
Add a schema validation for test results using [jsonschema](https://github.com/santhosh-tekuri/jsonschema).

### DIFF
--- a/helpers/integration.go
+++ b/helpers/integration.go
@@ -12,6 +12,7 @@ import (
 	"github.com/go-logr/logr"
 	applicationapiv1alpha1 "github.com/redhat-appstudio/application-api/api/v1alpha1"
 	"github.com/redhat-appstudio/integration-service/api/v1alpha1"
+	"github.com/santhosh-tekuri/jsonschema/v5"
 	tektonv1beta1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/labels"
@@ -50,6 +51,40 @@ type HACBSTestResult struct {
 	Failures  int    `json:"failures"`
 	Warnings  int    `json:"warnings"`
 }
+
+var testResultSchema = `{
+  "$schema": "http://json-schema.org/draft/2020-12/schema#",
+  "type": "object",
+  "properties": {
+    "result": {
+      "type": "string",
+      "enum": ["SUCCESS", "FAILURE", "WARNING", "SKIPPED", "ERROR"]
+    },
+    "namespace": {
+      "type": "string"
+    },
+    "timestamp": {
+      "type": "string",
+      "pattern": "^[0-9]{10}$"
+    },
+    "successes": {
+      "type": "integer",
+      "minimum": 0
+    },
+    "note": {
+      "type": "string"
+    },
+    "failures": {
+      "type": "integer",
+      "minimum": 0
+    },
+    "warnings": {
+      "type": "integer",
+      "minimum": 0
+    }
+  },
+  "required": ["result", "timestamp", "successes", "failures", "warnings"]
+}`
 
 // TaskRun is an integration specific wrapper around the status of a Tekton TaskRun.
 type TaskRun struct {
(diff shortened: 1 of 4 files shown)


helpers/integration.go
```diff
@@ -51,6 +51,40 @@ type HACBSTestResult struct {
 	Warnings  int    `json:"warnings"`
 }
 
+HACBSTestResult := `{
```
dheerajodha, line 54:
```suggestion
HACBSTestResult = `{
```

Maybe this will solve the following GoSec error that we're seeing (based on [this post](https://stackoverflow.com/a/20508376)): 
```
helpers/integration.go:54:1: syntax error: non-declaration statement outside function body
[line 54 : column 1] - expected declaration, found HACBSTestResult
```

dirgim: Could we store the schema in a separate JSON Schema file and use that as a reference? That way we could also provide it in documentation and to the other teams if they need to do validation on their side.

helpers/integration.go
```diff
@@ -51,6 +51,40 @@ type HACBSTestResult struct {
...
+		},
+	"timestamp": {
+		"type": "string",
+		"pattern": "^[0-9]{10}$"
+		},
+	"successes": {
+		"type": "integer",
+		"minimum": 0
+		},
+	"note": {
```
hongweiliu17, line 73: We may also add `maxLength` to limit the length of `note` then `HACBS_TEST_OUTPUT` won't exceed 4096 bytes?

helpers/test_schema.json
```diff
@@ -0,0 +1,33 @@
...
+    "timestamp": {
+      "type": "string",
+      "pattern": "^[0-9]{10}$"
+    },
+    "successes": {
+      "type": "integer",
+      "minimum": 0
+    },
+    "note": {
+      "type": "string"
```
dirgim, line 21: I think we do need the `maxLength` set here, something like 512 could be a good limit?

helpers/integration.go
Josh-Everett: Fixed declaration and separated the pattern to it's own file, test_schema.json in most recent patch.

helpers/integration.go
```diff
@@ -96,6 +97,11 @@ func (t *TaskRun) GetTestResult() (*HACBSTestResult, error) {
 	if t.testResult != nil {
 		return t.testResult, nil
 	}
+	// load schema for test validation
+	sch, err := jsonschema.Compile("helpers/test_schema.json")
+	if err != nil {
+		return nil, err
```
MartinBasti, line 103: Error chaining should be used here

helpers/integration.go
```diff
@@ -96,6 +97,11 @@ func (t *TaskRun) GetTestResult() (*HACBSTestResult, error) {
 	if t.testResult != nil {
 		return t.testResult, nil
 	}
+	// load schema for test validation
+	sch, err := jsonschema.Compile("helpers/test_schema.json")
```
dirgim, line 101: Hm, I don't think referencing the schema by relative path will work, since functions that call this will be in different places. I'm open to moving the schema back to be a JSON object and we can figure out generating a JSON file for reference in some other way.

helpers/integration.go
Josh-Everett: adjusted maxLength of the note in most recent patch.

helpers/test_schema.json
```diff
@@ -0,0 +1,33 @@
+{
+  "$schema": "http://json-schema.org/draft-07/schema#",
```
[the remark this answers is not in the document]
Josh-Everett: Yes, believe this should work fine. The library is reportedly compatible with `2020-12`but much of the example code had been been using the prior draft.

helpers/test_schema.json
Josh-Everett: `maxLength` for `note` was updated to be 512.

status/format_test.go
Josh-Everett, line 1: `example-task-7` was removed, it was to test for the `UNEXPECTED` result, which as not included in the [API contracts ADR](https://github.com/redhat-appstudio/book/blob/main/ADR/0013-integration-service-api-contracts.md#tekton-result-format)

helpers/integration.go
```diff
@@ -51,6 +52,40 @@ type HACBSTestResult struct {
 	Warnings  int    `json:"warnings"`
 }
 
+var testSchema = `{
```
MartinBasti, line 55: Maybe `testResultSchema` ?

dirgim (approved):
The changes look good to me code-wise. @Josh-Everett have you had a chance to test it on your local environment?

helpers/integration.go
```diff
@@ -97,13 +132,25 @@ func (t *TaskRun) GetTestResult() (*HACBSTestResult, error) {
...
 		if taskRunResult.Name == HACBSTestOutputName {
 			var result HACBSTestResult
+			var v interface{}
 			err := json.Unmarshal([]byte(taskRunResult.Value.StringVal), &result)
 			if err != nil {
-				return nil, err
+				return nil, fmt.Errorf("error while mapping json data from taskRun %s: to HACBSTestResult - %w", taskRunResult.Name, err)
+			}
+			if err := json.Unmarshal([]byte(taskRunResult.Value.StringVal), &v); err != nil {
+				return nil, fmt.Errorf("error while mapping json data from taskRun %s: - %w", taskRunResult.Name, err)
```
MartinBasti, line 150: Why `-` before `%w` ?

Josh-Everett: This was helpful for me visually when I was tracing back some errors but can change back to follow prior formatting convention.

MartinBasti (approved):

helpers/integration.go
Josh-Everett: Updated pr uses error chaining here

helpers/integration.go
```diff
@@ -104,6 +110,9 @@ func (t *TaskRun) GetTestResult() (*HACBSTestResult, error) {
 			if err != nil {
 				return nil, err
 			}
+			if err = sch.Validate(result); err != nil {
+				return nil, err
```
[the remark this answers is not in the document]
Josh-Everett: Updated pr uses error chaining here

helpers/integration.go
Josh-Everett: updated naming to `testResultSchema` in most recent PR

PR #141, opened 2023-04-03
feat(STONEINTG-354): update lastBuiltCommit with Revision field

dheerajodha:
This PR updates the Global Candidate List with the commit sha for the built Component.

### DIFF
--- a/controllers/snapshot/snapshot_adapter.go
+++ b/controllers/snapshot/snapshot_adapter.go
@@ -18,9 +18,10 @@ package snapshot
 
 import (
 	"context"
+	"reflect"
+	"strings"
 
 	ctrl "sigs.k8s.io/controller-runtime"
-	"strings"
 
 	"github.com/go-logr/logr"
 	applicationapiv1alpha1 "github.com/redhat-appstudio/application-api/api/v1alpha1"
@@ -156,9 +157,18 @@ func (a *Adapter) EnsureGlobalCandidateImageUpdated() (reconciler.OperationResul
 				a.component.Spec.ContainerImage = component.ContainerImage
 				err := a.client.Patch(a.context, a.component, patch)
 				if err != nil {
-					a.logger.Error(err, "Failed to update Global Candidate for the Component",
+					a.logger.Error(err, "Failed to update .Spec.ContainerImage of Global Candidate for the Component",
 						"Component.Name", a.component.Name)
 					return reconciler.RequeueWithError(err)
+				}
+				if reflect.ValueOf(component.Source).IsValid() {
+					a.component.Status.LastBuiltCommit = component.Source.GitSource.Revision
+					err = a.client.Status().Patch(a.context, a.component, patch)
+					if err != nil {
+						a.logger.Error(err, "Failed to update .Status.LastBuiltCommit of Global Candidate for the Component",
+							"Component.Name", a.component.Name)
+						return reconciler.RequeueWithError(err)
+					}
 				}
 			}
 		}
--- a/controllers/snapshot/snapshot_adapter_test.go
+++ b/controllers/snapshot/snapshot_adapter_test.go
@@ -2,10 +2,11 @@ package snapshot
 
 import (
 	"bytes"
-	"github.com/go-logr/logr"
-	"github.com/tonglil/buflogr"
 	"reflect"
 	"time"
+
+	"github.com/go-logr/logr"
+	"github.com/tonglil/buflogr"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
@@ -39,6 +40,7 @@ var _ = Describe("Snapshot Adapter", Ordered, func() {
 		integrationTestScenario *integrationv1alpha1.IntegrationTestScenario
 		env                     applicationapiv1alpha1.Environment
 		sample_image            string
+		sample_revision         string
 	)
 	const (
 		SampleRepoLink = "https://github.com/devfile-samples/devfile-sample-java-springboot-basic"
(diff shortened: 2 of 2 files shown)


dheerajodha:
Currently working on adding unit tests for this change.

controllers/snapshot/snapshot_adapter.go
```diff
@@ -154,6 +154,7 @@ func (a *Adapter) EnsureGlobalCandidateImageUpdated() (reconciler.OperationResul
 		for _, component := range a.snapshot.Spec.Components {
 			if component.Name == a.component.Name {
 				a.component.Spec.ContainerImage = component.ContainerImage
+				a.component.Status.LastBuiltCommit = component.Source.GitSource.Revision
```
dirgim, line 157: In order to update the component's status, we will need to also call `a.client.Status.Patch` separately, like e.g. [here](https://github.com/redhat-appstudio/integration-service/blob/main/gitops/snapshot.go#L118).

dheerajodha: Thank you, that makes sense.
Just out of curiosity, wouldn't this already be taken care of by `a.client.Patch()`? It would update both the `spec` and `status` sections of the Component together?

dirgim: I thought so too initially but it turned out It wouldn't unfortunately

dirgim: We'll also need to check if the `component.Source.GitSource` is defined before using it here. This should resolve the [unit test failure](https://github.com/redhat-appstudio/integration-service/actions/runs/4600217515/jobs/8126517849?pr=141).

dirgim (changes requested):

dheerajodha:
Adding the unit test in a few mins.

controllers/snapshot/snapshot_adapter.go
```diff
@@ -154,11 +155,20 @@ func (a *Adapter) EnsureGlobalCandidateImageUpdated() (reconciler.OperationResul
...
+				if reflect.ValueOf(component.Source).IsValid() {
+					a.component.Status.LastBuiltCommit = component.Source.GitSource.Revision
+					err := a.client.Status().Patch(a.context, a.component, patch)
+					if err != nil {
+						a.logger.Error(err, "Failed to update the Status section of the Component while updating Global Candidate for the Component",
+							"Component.Name", a.component.Name)
+						return reconciler.RequeueWithError(err)
+					}
+				}
+				err1 := a.client.Patch(a.context, a.component, patch)
```
dirgim, line 167: I think it would be a good idea to move this patch call before we start checking if the status should be updated (so around line 157) since updating the `component.containerImage` is the more crucial operation of the two.

dheerajodha: Thank you for your help!

controllers/snapshot/snapshot_adapter.go
```diff
@@ -154,11 +155,20 @@ func (a *Adapter) EnsureGlobalCandidateImageUpdated() (reconciler.OperationResul
...
-				if err != nil {
-					a.logger.Error(err, "Failed to update Global Candidate for the Component",
+				err1 := a.client.Patch(a.context, a.component, patch)
+				if err1 != nil {
+					a.logger.Error(err1, "Failed to update Global Candidate for the Component",
 						"Component.Name", a.component.Name)
-					return reconciler.RequeueWithError(err)
+					return reconciler.RequeueWithError(err1)
+				}
+				if reflect.ValueOf(component.Source).IsValid() {
```
hongweiliu17, line 164: I feel we should also check if the value of `component.Source.GitSource.Revision` exist here

controllers/snapshot/snapshot_adapter.go
```diff
@@ -154,11 +155,20 @@ func (a *Adapter) EnsureGlobalCandidateImageUpdated() (reconciler.OperationResul
 		for _, component := range a.snapshot.Spec.Components {
 			if component.Name == a.component.Name {
 				a.component.Spec.ContainerImage = component.ContainerImage
-				err := a.client.Patch(a.context, a.component, patch)
-				if err != nil {
-					a.logger.Error(err, "Failed to update Global Candidate for the Component",
+				err1 := a.client.Patch(a.context, a.component, patch)
+				if err1 != nil {
+					a.logger.Error(err1, "Failed to update Global Candidate for the Component",
```
hongweiliu17, line 160:
```suggestion
					a.logger.Error(err1, "Failed to update .Spec.ContainerImage of Global Candidate for the Component",
```

dheerajodha: Thank you! updated.

controllers/snapshot/snapshot_adapter.go
dheerajodha: Thanks for your help!

controllers/snapshot/snapshot_adapter.go
```diff
@@ -154,11 +155,20 @@ func (a *Adapter) EnsureGlobalCandidateImageUpdated() (reconciler.OperationResul
 		for _, component := range a.snapshot.Spec.Components {
 			if component.Name == a.component.Name {
 				a.component.Spec.ContainerImage = component.ContainerImage
-				err := a.client.Patch(a.context, a.component, patch)
-				if err != nil {
-					a.logger.Error(err, "Failed to update Global Candidate for the Component",
+				err1 := a.client.Patch(a.context, a.component, patch)
```
jinqi7, line 158: Just use "err" instead of "err1" (and the following lines), then change line 166 to "err = a.client.Status().Patch(a.context, a.component, patch)" will be ok.

dheerajodha: Done, thank you!

dirgim (approved):
LGTM

PR #145, opened 2023-04-11
fix: set permissions to update Component.Status

dheerajodha:
### Need for this PR:

Previously merged PR: #141 deals with updating the `Status` field of the component, but currently our service doesn't have the required permissions to do this. So this PR adds those permissions and fixes the [failing e2e-test](https://github.com/redhat-appstudio/infra-deployments/pull/1638#issuecomment-1503315698).

### DIFF
--- a/config/rbac/component_editor_role.yaml
+++ b/config/rbac/component_editor_role.yaml
@@ -25,3 +25,5 @@ rules:
   - components/status
   verbs:
   - get
+  - patch
+  - update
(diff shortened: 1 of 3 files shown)


dirgim (approved):
LGTM

dheerajodha:
The Snapshot controller is now able to patch the `component.Status` section, as you can now see the `lastBuiltCommit` field having the SHA value:
image

hongweiliu17 (approved):
LGTM

PR #146, opened 2023-04-12
feat(STONEINTG-371): Don't add a component to Snapshots if it doesn't have a container image defined

14rcole:
The acceptance criteria stated "Updates for relevant unit/controller tests".  I believe I covered all relevant tests.  Please let me know if there are other places where I should be adding additional checks.

### DIFF
--- a/controllers/pipeline/pipeline_adapter.go
+++ b/controllers/pipeline/pipeline_adapter.go
@@ -412,6 +412,14 @@ func (a *Adapter) prepareSnapshot(application *applicationapiv1alpha1.Applicatio
 			// Get ComponentSource for the component which is not built in this pipeline
 			componentSource = a.getComponentSourceFromComponent(&applicationComponent)
 		}
+
+		// If containerImage is empty, we have run into a race condition in
+		// which multiple components are being built in close succession.
+		// We omit this not-yet-built component from the snapshot rather than
+		// including a component that is incomplete.
+		if containerImage == "" {
+			continue
+		}
 		snapshotComponents = append(snapshotComponents, applicationapiv1alpha1.SnapshotComponent{
 			Name:           applicationComponent.Name,
 			ContainerImage: containerImage,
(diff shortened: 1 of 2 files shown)


controllers/pipeline/pipeline_adapter.go
```diff
@@ -412,6 +412,14 @@ func (a *Adapter) prepareSnapshot(application *applicationapiv1alpha1.Applicatio
 			// Get ComponentSource for the component which is not built in this pipeline
 			componentSource = a.getComponentSourceFromComponent(&applicationComponent)
 		}
+
+		// If containerImage is empty, we have run into a race condition in
+		// which multiple components are being built in close succession.
+		// We omit this not-yet-built component from the snapshot rather than
+		// including a component that is incomplete.
+		if containerImage == "" {
```
Josh-Everett, line 420: Do we need to be concerned by the possibility of container Image being nil?

Josh-Everett:
Looks good, just curious about nil condition. Think this is ready to open up to others as well.

14rcole:
I think the failure in the unit tests is another race condition.  The test in question, `It ensures the global component list is changed and compositeSnapshot should be created`, creates a composite snapshot, waits for the snapshot CR to exist with an `Eventually` block, then attempts to create the snapshot again to ensure that the function correctly identifies the existing snapshot CR and returns it rather than attempting to create a new one.  Most of the time when I run `make test` locally I run into the same error.  Occassionally the error doesn't occur.  I think the best option is to increase the timeout of this eventually block but I'm going to verify during standup today that there's not a better option.

dirgim:
The [gitlint check](https://github.com/redhat-appstudio/integration-service/actions/runs/4689564616/jobs/8314106779?pr=146) is failing because of the title of the commit needing to be compatible with the conventional commit format, so it would need to be changed to something like this: `feat(STONEINTG-371): snapshot components must have container images`

dirgim:
LGTM, this just needs a rebase on the latest master to fix e2e tests

PR #148, opened 2023-04-13
fix: checks Revision field for nil value

dheerajodha:
Fixes the error seen on [this Slack thread](https://redhat-internal.slack.com/archives/C030SGP2VB9/p1681393407389439).


dirgim (approved):
LGTM

PR #150, opened 2023-04-13
test(STONEINTG-385): update unit tests to be more clear about errors

14rcole:
This is outlined more thoroughly in Jira, but basically using `Expect(a == b).To(BeTrue())` isn't idiomatic and hides important information in the case that tests fail.  This MR fixes statements like that throughout our unit tests.

### DIFF
--- a/controllers/pipeline/pipeline_adapter_test.go
+++ b/controllers/pipeline/pipeline_adapter_test.go
@@ -370,41 +370,41 @@ var _ = Describe("Pipeline Adapter", Ordered, func() {
 
 	It("ensures the Application Components can be found ", func() {
 		applicationComponents, err := adapter.getAllApplicationComponents(hasApp)
-		Expect(err == nil).To(BeTrue())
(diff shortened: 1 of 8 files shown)


14rcole:
> 1: UC1 Description should start with lowercase
My commitment to good grammar is killing me here

14rcole:
The Definition of Done document says that changes should be deployed to stage.  I assume that doesn't apply here since only unit tests were change?

dirgim:
> The Definition of Done document says that changes should be deployed to stage. I assume that doesn't apply here since only unit tests were change?

Yes, it should be fine to consider this as done after the PR is merged.

dheerajodha:
do we want to fix the conflicts? or is that not needed?

Josh-Everett (approved):
lgtm +1

14rcole:
@dheerajodha I fixed them.  They were just a couple of lines from #146.  Waiting for tests to pass before merging.

PR #152, opened 2023-04-18
Webhook - STONEINTG-386

jsztuka:
Creating feature branch containing mutating webhooks from v1alpha1 to v1beta1 version.


config/default/kustomization.yaml
```diff
@@ -18,11 +18,11 @@ bases:
 - ../manager
 # [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix including the one in
 # crd/kustomization.yaml
-#- ../webhook
+- ../webhook
 # [CERTMANAGER] To enable cert-manager, uncomment all sections with 'CERTMANAGER'. 'WEBHOOK' components are required.
-#- ../certmanager
+- ../certmanager
 # [PROMETHEUS] To enable prometheus monitor, uncomment all sections with 'PROMETHEUS'.
-- ../prometheus
```
jinqi7, line 25: Why comments this line? If it's not related to the webhook, please leave it as uncomment for the prometheus usage.

config/manager/kustomization.yaml
```diff
@@ -14,5 +14,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/redhat-appstudio/integration-service
-  newTag: next
+  newName: quay.io/jsztuka/integration-service
```
jinqi7, line 17: Is this expected to use an image from a private  account here?

jinqi7 (changes requested):

config/default/kustomization.yaml
jsztuka: Good point, prometheus is not needed.

config/manager/kustomization.yaml
jsztuka: Oh it changes the source image when you run make docker-build docker-push, we want to use redhat-appstudio one.

api/v1alpha1/integrationtestscenario_conversion.go
```diff
@@ -0,0 +1,105 @@
...
+			dst.Spec.Params = append(dst.Spec.Params, v1beta1.PipelineParameter(par))
+		}
+	}
+	if src.Spec.Contexts != nil {
+		for _, par := range src.Spec.Contexts {
+			dst.Spec.Contexts = append(dst.Spec.Contexts, v1beta1.TestContext(par))
+		}
+	}
+	dst.Spec.ResolverRef = v1beta1.ResolverRef{
+		Resolver: "bundle",
```
dirgim, line 53: Looks like this needs to be `bundles` ([docs](https://tekton.dev/docs/pipelines/bundle-resolver/#bundles-resolver)), just double checked it in local testing.

dirgim: Thanks

api/v1alpha1/integrationtestscenario_conversion.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	return nil
+}
+
+func (dst *IntegrationTestScenario) ConvertFrom(srcRaw conversion.Hub) error {
+	src := srcRaw.(*v1beta1.IntegrationTestScenario)
+	dst.ObjectMeta = src.ObjectMeta
+	dst.Spec.Application = src.Spec.Application
+	if !reflect.ValueOf(src.Spec.Environment).IsZero() {
+		src.Spec.Environment = v1beta1.TestEnvironment(dst.Spec.Environment)
```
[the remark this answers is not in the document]
hongweiliu17: I don't think we need to do it or can do it. In `v1beta1`, we introduce some change (bundle in Resolver) which can't be converted to `v1alpha1`,  I feel it is impossible to convert all field from `v1beta1` and `v1alpha1`, or we may give some explanation to users about we can't convert all `v1beta1` fields to `v1alpha1` field and the change is not backward compatible?. Actually I don't think users will create CR in `v1beta1` but try to get CR in `v1beta1`.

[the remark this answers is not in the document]
jsztuka: Made adjustments, take a look.

api/v1alpha1/integrationtestscenario_conversion.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	if src.Spec.Params != nil {
+		src.Spec.Params = append(src.Spec.Params, src.Spec.Params...)
```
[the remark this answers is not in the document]
jsztuka: Was not able to have src at right side.

api/v1alpha1/integrationtestscenario_conversion.go
```diff
@@ -0,0 +1,105 @@
...
+	}
+	if src.Spec.Contexts != nil {
+		src.Spec.Contexts = append(src.Spec.Contexts, src.Spec.Contexts...)
+	}
+	src.Spec.ResolverRef = v1beta1.ResolverRef{
```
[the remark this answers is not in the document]
jsztuka: Added rough value assign on this side, please take a look.

api/v1alpha1/integrationtestscenario_conversion.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+func (dst *IntegrationTestScenario) ConvertFrom(srcRaw conversion.Hub) error {
+	src := srcRaw.(*v1beta1.IntegrationTestScenario)
+	dst.ObjectMeta = src.ObjectMeta
+	dst.Spec.Application = src.Spec.Application
+	if !reflect.ValueOf(dst.Spec.Environment).IsZero() {
+		dst.Spec.Environment = TestEnvironment(src.Spec.Environment)
+	}
+	if dst.Spec.Params != nil {
```
hongweiliu17, line 79:
```suggestion
	if src.Spec.Params != nil {
```

api/v1alpha1/integrationtestscenario_conversion.go
```diff
@@ -0,0 +1,95 @@
...
+		for _, par := range src.Spec.Params {
+			dst.Spec.Params = append(dst.Spec.Params, PipelineParameter(par))
+		}
+	}
+	if dst.Spec.Contexts != nil {
```
hongweiliu17, line 84:
```suggestion
	if src.Spec.Contexts != nil {
```

api/v1alpha1/integrationtestscenario_conversion.go
```diff
@@ -0,0 +1,95 @@
...
+		for _, par := range src.Spec.Contexts {
+			dst.Spec.Contexts = append(dst.Spec.Contexts, TestContext(par))
+		}
+	}
+
+	dst.Spec.Pipeline = src.Spec.ResolverRef.Params[1].Value
```
hongweiliu17, line 90: We do these two line only when `src.Spec.ResolverRef.Resolver == 'bundles'`

api/v1alpha1/integrationtestscenario_conversion.go
jsztuka: updated

api/v1alpha1/integrationtestscenario_conversion.go
jsztuka: updated

api/v1alpha1/integrationtestscenario_conversion.go
jsztuka: There should not be any other type of resolver present from previous version.

jsztuka: Do you have any specific way in mind to allign those attributes?

hongweiliu17: Here `dst` is CR in `v1alpha1` version, `src` is CR in `v1beta1`, This function is to convert CR from `v1beta1` to `v1alpha1` version.

And I just realize we can't use `src.Spec.ResolverRef.Params[1].Value` because the `Params` is a list https://tekton.dev/docs/pipelines/bundle-resolver/#task-resolution so we can't make sure the first key-value pair is for `bundle`. So I think we need to find the their value by its key.

dirgim: Agreed, I think we still need to address this by looping through `src.Spec.ResolverRef.Params` and checking the `name` of the `param` before setting it as either `bundle` or `pipeline`.

jsztuka: Updated.

dirgim: Thanks, I believe that this resolves this conversation

config/manager/kustomization.yaml
dirgim: I just double checked and can confirm that this was resolved.

api/v1alpha1/integrationtestscenario_conversion.go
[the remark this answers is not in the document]
dirgim: Can confirm this was updated, resolving the conversation.

api/v1alpha1/integrationtestscenario_conversion.go
```diff
@@ -0,0 +1,105 @@
...
+	dst.ObjectMeta = src.ObjectMeta
+	dst.Spec.Application = src.Spec.Application
+	if !reflect.ValueOf(src.Spec.Environment).IsZero() {
+		src.Spec.Environment = v1beta1.TestEnvironment(dst.Spec.Environment)
+	}
+	if src.Spec.Params != nil {
+		src.Spec.Params = append(src.Spec.Params, src.Spec.Params...)
+	}
+	if src.Spec.Contexts != nil {
+		src.Spec.Contexts = append(src.Spec.Contexts, src.Spec.Contexts...)
```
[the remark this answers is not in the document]
dirgim: Can confirm this was updated, resolving the conversation.

config/default/kustomization.yaml
dirgim: Prometheus is back in, resolving this conversation.

api/v1alpha1/integrationtestscenario_conversion.go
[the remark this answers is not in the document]
dirgim: This has been modified as part of other discussions, I believe it can be resolved now.

dirgim (approved):
LGTM

api/v1alpha1/integrationtestscenario_conversion.go
```diff
@@ -0,0 +1,103 @@
...
+			}
+			if par.Name == "name" {
+				dst.Spec.Pipeline = par.Value
+			}
+		}
+	}
+	return nil
+}
+
+// TODO(user): EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
```
MartinBasti, line 103: :eyes:

api/v1beta1/integrationtestscenario_types.go
```diff
@@ -1,12 +1,9 @@
 /*
 Copyright 2022.
-
```
MartinBasti, line 3: Why these lines are being removed?

api/v1beta1/integrationtestscenario_webhook.go
```diff
@@ -0,0 +1,29 @@
...
+	ctrl "sigs.k8s.io/controller-runtime"
+)
+
+func (r *IntegrationTestScenario) SetupWebhookWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewWebhookManagedBy(mgr).
+		For(r).
+		Complete()
+}
+
+// TODO(user): EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
```
MartinBasti, line 29: :eyes:

config/samples/appstudio_v1alpha1_integrationtestscenario.yaml
```diff
@@ -1,6 +1,11 @@
...
+  name: integration-service
+---
 apiVersion: appstudio.redhat.com/v1alpha1
 kind: IntegrationTestScenario
 metadata:
   name: integrationtestscenario-sample
 spec:
-# TODO(user): Add fields here
\ No newline at end of file
+# TODO(user): Add fields here
```
MartinBasti, line 11: :eyes:

config/samples/appstudio_v1alpha1_integrationtestscenario.yaml
```diff
@@ -1,6 +1,11 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: integration-service
+---
 apiVersion: appstudio.redhat.com/v1alpha1
 kind: IntegrationTestScenario
 metadata:
   name: integrationtestscenario-sample
 spec:
```
MartinBasti, line 10: Do we miss sample data here?

config/samples/appstudio_v1beta1_integrationtestscenario.yaml
```diff
@@ -0,0 +1,17 @@
...
+metadata:
+  labels:
+    app.kubernetes.io/name: integrationtestscenario
+    app.kubernetes.io/instance: integrationtestscenario-sample
+    app.kubernetes.io/part-of: integration-service
+    app.kubernetes.io/managed-by: kustomize
+    app.kubernetes.io/created-by: integration-service
+  name: integrationtestscenario-sample
+spec:
+  # TODO(user): Add fields here
```
MartinBasti, line 17: :eyes:

api/v1alpha1/integrationtestscenario_conversion.go
```diff
@@ -0,0 +1,103 @@
...
+
+func (r *IntegrationTestScenario) SetupWebhookWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewWebhookManagedBy(mgr).
+		For(r).
+		Complete()
+}
+
+// Hub marks this type as a conversion hub.
+// ConvertTo converts this ITS to the Hub version (v1beta1).
+func (src *IntegrationTestScenario) ConvertTo(dstRaw conversion.Hub) error {
```
14rcole, line 35: Why is this called `ConvertTo` rather than something like `ConvertToHubVersion`?

jsztuka: Im little affraid that its standardized naming, I saw it used in every example of webhooks that mutates to different version.

config/samples/appstudio_v1alpha1_integrationtestscenario.yaml
jsztuka: Yep, missing namespace for the sample, let me add it.

jsztuka: It can work without spec, there can be specified whatever we want, but I would keep it this way.

MartinBasti (approved):

PR #154, opened 2023-04-19
feat(STONEINTG-333): create a controller to trigger integration pipeline

hongweiliu17:
Copied most of code from https://github.com/redhat-appstudio/integration-service/pull/81 and add the code for the new Environments API

* Add binding controller for test pipeline deployments on ephemeral envs
* Add functions that can detect a SEB with a finished deployment
* Ensure creation of test pipelines with associated DT
* Add unit tests for the binding controller
* Add unit tests for updated code in other packages

### DIFF
--- a/controllers/snapshot/cache.go
+++ b/controllers/snapshot/cache.go
@@ -54,8 +54,8 @@ func SetupApplicationCache(mgr ctrl.Manager) error {
 		"spec.environment", applicationIndexFunc)
 }
 
-// SetupEnvironmentCache adds a new index field to be able to search SnapshotEnvironmentBindings by Application.
-func SetupEnvironmentCache(mgr ctrl.Manager) error {
+// SetupSnapshotEnvironmentBindingCache adds a new index field to be able to search SnapshotEnvironmentBindings by Application.
+func SetupSnapshotEnvironmentBindingCache(mgr ctrl.Manager) error {
 	environmentIndexFunc := func(obj client.Object) []string {
 		return []string{obj.(*applicationapiv1alpha1.SnapshotEnvironmentBinding).Spec.Application}
 	}
--- a/tekton/integration_pipeline.go
+++ b/tekton/integration_pipeline.go
@@ -19,11 +19,13 @@ package tekton
 import (
 	"encoding/json"
 	"fmt"
+	"reflect"
 
 	applicationapiv1alpha1 "github.com/redhat-appstudio/application-api/api/v1alpha1"
 	"github.com/redhat-appstudio/integration-service/api/v1alpha1"
 	"github.com/redhat-appstudio/integration-service/helpers"
 	tektonv1beta1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
@@ -53,6 +55,9 @@ var (
 
 	// SnapshotNameLabel is the label of specific the name of the snapshot associated with PipelineRun
 	SnapshotNameLabel = fmt.Sprintf("%s/%s", ResourceLabelSuffix, "snapshot")
+
+	// EnvironmentNameLabel is the label of specific the name of the environment associated with PipelineRun
+	EnvironmentNameLabel = fmt.Sprintf("%s/%s", ResourceLabelSuffix, "environment")
 
 	// ApplicationNameLabel is the label of specific the name of the Application associated with PipelineRun
 	ApplicationNameLabel = fmt.Sprintf("%s/%s", ResourceLabelSuffix, "application")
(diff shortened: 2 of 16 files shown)


controllers/binding/binding_adapter.go
```diff
@@ -0,0 +1,180 @@
...
+		return nil, fmt.Errorf("failed to find deploymentTarget %s: %w", deploymentTarget.Name, err)
+	}
+
+	return deploymentTarget, nil
+}
+
+// GetDeploymentTargetClaimForEnvironment try to find the DeploymentTargetClaim whose name is defined in Environment
+// if not found, an error is returned
+func (a *Adapter) GetDeploymentTargetClaimForEnvironment(environment *applicationapiv1alpha1.Environment) (*applicationapiv1alpha1.DeploymentTargetClaim, error) {
+	dtcName := environment.Spec.Configuration.Target.DeploymentTargetClaim.ClaimName
```
dirgim, line 143: Should we also check first if the `environment.Spec.Configuration.Target` defined to avoid errors here?

controllers/snapshot/cache.go
```diff
@@ -54,8 +54,8 @@ func SetupApplicationCache(mgr ctrl.Manager) error {
 		"spec.environment", applicationIndexFunc)
 }
 
-// SetupEnvironmentCache adds a new index field to be able to search SnapshotEnvironmentBindings by Application.
-func SetupEnvironmentCache(mgr ctrl.Manager) error {
+// SetupSnapshotEnvironmentBindingCache adds a new index field to be able to search SnapshotEnvironmentBindings by Application.
+func SetupSnapshotEnvironmentBindingCache(mgr ctrl.Manager) error {
```
dirgim, line 58: Thanks for catching this!

tekton/integration_pipeline.go
```diff
@@ -171,3 +176,34 @@ func (r *IntegrationPipelineRun) WithApplicationAndComponent(application *applic
 
 	return r
 }
+
+// WithEnvironmentAndDeploymentTarget adds a param containing the DeploymentTarget connection details and Environemtn name
```
dirgim, line 180: typo: `Environemtn` -> `Environment`

controllers/binding/binding_adapter.go
hongweiliu17: Thanks, I concerned it and `dtc.Spec.TargetName`, so I tested them, the result is if the `Target` and `TargetName` are not defined, `environment.Spec.Configuration.Target.DeploymentTargetClaim.ClaimName` and `dtc.Spec.TargetName` will be "" and then will be caught by `if dtName == ""` and `if dtcName == ""`, therefore I didn't check `environment.Spec.Configuration.Target` first. But let me add code to handle it to make code review easier.

tekton/integration_pipeline.go
hongweiliu17: Thanks, I realize I ofter have typo for this `environment`

controllers/binding/binding_adapter.go
```diff
@@ -0,0 +1,180 @@
...
+	err = a.client.Create(a.context, pipelineRun)
+	if err != nil {
+		return err
+	}
+
+	return nil
+
+}
+
+// GetDeploymentTargetForEnvironment gets the DeploymentTarget associated with Environdment, if the DeploymentTarget is not found, an error will be returned
```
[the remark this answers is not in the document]
hongweiliu17: Thanks!

controllers/snapshot/cache.go
hongweiliu17: This is copied from your PR. Thanks for your finishing most of code!

tekton/integration_pipeline.go
dirgim: Yeah, I have the same issue, thanks for correcting it

dheerajodha:
Based on the names of currently-existing controllers I see that they are named after the resources they watch. So I was wondering if we can name it something other than "binding" :thinking: 

`snapshotEnvironmentBinding`? But I see that it is too long, and uses camelcase...

controllers/binding/binding_adapter.go
dirgim: Thanks!

hongweiliu17:
> Based on the names of currently-existing controllers I see that they are named after the resources they watch. So I was wondering if we can name it something other than "binding" 🤔
> 
> `snapshotEnvironmentBinding`? But I see that it is too long, and uses camelcase...
I checked other team's controller and they are nameed in different style but they are all lowercase, so let's use snapshotenvironmentbinding. Thanks.

codecov.yml
```diff
@@ -1,18 +0,0 @@
-# Documentation: https://docs.codecov.io/docs/codecov-yaml
```
dirgim, line 1: I think we want to keep this configuration file

cover.html
```diff
@@ -0,0 +1,5980 @@
+
```
dirgim, line 1: I think we can safely remove this report file

codecov.yml
hongweiliu17: Thanks a lot! This is deleted wrongly after I check the code coverage, I wanted to delete cover.html.

dirgim: I thought that might be it, thanks

dirgim (approved):
LGTM

PR #157, opened 2023-04-19
feat(STONEINTG-339): use v1beta1 IntegrationTestScenarios

dirgim:
PR for the `webhook` feature branch:

* Modify all packages to start using v1beta1 ITS
* Set v1beta1 to be the storage version
* Set v1alpha1 to be deprecated with a warning
* Add resolverRef to integration test pipelineRun creation
* Update all unit tests to have the new required fields


controllers/pipeline/cache.go
```diff
@@ -19,7 +19,7 @@ package pipeline
 import (
 	"context"
 	applicationapiv1alpha1 "github.com/redhat-appstudio/application-api/api/v1alpha1"
-	"github.com/redhat-appstudio/integration-service/api/v1alpha1"
+	"github.com/redhat-appstudio/integration-service/api/v1beta1"
```
hongweiliu17, line 22: This is not relating to the schema change, but I am thinking maybe we can import this api package in the same way in all files, use `"github.com/redhat-appstudio/integration-service/api/v1beta1"` or `integrationv1beta1 "github.com/redhat-appstudio/integration-service/api/v1beta1"`, then it is easier to copy code in different files during the development or update the schema version the future. Thanks!

hongweiliu17 (approved):

controllers/pipeline/cache.go
dirgim: Good point, I updated all references to use `v1beta1` and removed `integrationv1beta1` named imports.

jsztuka (approved):
lgtm

PR #158, opened 2023-04-19
fix: set maxlength for note in schema validation

Josh-Everett:
MaxLength was set and then removed at one point during development, adding it back.


14rcole (approved):

dirgim (approved):
LGTM

MartinBasti:
but wait, why 512?

Josh-Everett:
Dropping after discussion in standup that the tekton message limit of 4096 is sufficient.

PR #159, opened 2023-04-20
feat(STONEINTG-184): Migrate functions to load resources from cluster

14rcole:
- Migrate all relevant functions to `loader/loader.go`
- Set up unit testing for new loader package
- Migrate/add unit tests as applicable.


14rcole:
I see this error in the docker build test:
```
controllers/pipeline/pipeline_adapter.go:31:2: no required module provides package github.com/redhat-appstudio/integration-service/loader; to add it:
	go get github.com/redhat-appstudio/integration-service/loader
The command '/bin/sh -c CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o manager main.go' returned a non-zero code: 1
```
Is this happening because the loader package is new?  Is there something else I need to do to make sure it can build?

dirgim:
> I see this error in the docker build test:
> 
> ```
> controllers/pipeline/pipeline_adapter.go:31:2: no required module provides package github.com/redhat-appstudio/integration-service/loader; to add it:
> 	go get github.com/redhat-appstudio/integration-service/loader
> The command '/bin/sh -c CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -a -o manager main.go' returned a non-zero code: 1
> ```
> 
> Is this happening because the loader package is new? Is there something else I need to do to make sure it can build?

Yep, it the package will need to be copied in the [Dockerfile](https://github.com/redhat-appstudio/integration-service/blob/main/Dockerfile#L23) so the build can work.

dirgim:
Unfortunately, in the meantime more getter/loader functions made it into the codebase - namely the ones in the [snapshotenvironmentbinding](https://github.com/redhat-appstudio/integration-service/blob/main/controllers/binding/snapshotenvironmentbinding_controller.go#L109-L175) controller.

loader/loader_test.go
```diff
@@ -0,0 +1,282 @@
...
+		Expect(app).NotTo(BeNil())
+		Expect(app.ObjectMeta).To(Equal(hasApp.ObjectMeta))
+	})
+
+	It("ensures we can get the Snapshot from a Pipeline Run", func() {
+		snapshot, err := GetSnapshotFromPipelineRun(k8sClient, ctx, testPipelineRun)
+		Expect(err).To(BeNil())
+		Expect(snapshot).NotTo(BeNil())
+		Expect(snapshot.ObjectMeta).To(Equal(hasSnapshot.ObjectMeta))
+	})
```
14rcole, line 281: This is the test that is failing, with the error:
```
• [FAILED] [0.025 seconds]
Loader [It] ensures we can get the Snapshot from a Pipeline Run
/var/home/ryan/Development/RHTAP/integration-service/loader/loader_test.go:276

  [FAILED] Expected
      <*errors.errorString | 0xc000a8e690>: {
          s: "the pipeline has no snapshot associated with it",
      }
  to be nil
  In [It] at: /var/home/ryan/Development/RHTAP/integration-service/loader/loader_test.go:278 @ 04/26/23 15:45:10.287
------------------------------

Summarizing 1 Failure:
  [FAIL] Loader [It] ensures we can get the Snapshot from a Pipeline Run
  /var/home/ryan/Development/RHTAP/integration-service/loader/loader_test.go:278
```

How can I add a snapshot to the pipeline run?

14rcole: That worked, thanks for your help.

Josh-Everett (approved):
Nice patch, lgtm

loader/loader.go
```diff
@@ -0,0 +1,194 @@
+package loader
```
MartinBasti, line 1: Missing licence

loader/cache.go
```diff
@@ -0,0 +1,38 @@
+package loader
```
MartinBasti, line 1: license is missing

and also module description about purpose of the module

loader/loader_suite_test.go
```diff
@@ -0,0 +1,98 @@
+package loader
```
MartinBasti, line 1: license

loader/loader_test.go
```diff
@@ -0,0 +1,284 @@
+package loader
```
MartinBasti, line 1: license

loader/cache.go
14rcole: I've added the license to all files in `loader`.  Where would the module description go?  I don't see something like that in any of the other packages in the integration service.

Josh-Everett (approved):

loader/cache.go
MartinBasti: something like this from https://go.dev/blog/godoc
```
// Package sort provides primitives for sorting slices and user-defined
// collections.
package sort
```

We don't have it, but we should start adding them at least into new modules

loader/cache.go
```diff
@@ -0,0 +1,54 @@
+/*
+Copyright 2022.
```
MartinBasti, line 2: 2023

PR #161, opened 2023-04-23
feat(STONEINTG-334): delete ephemeral Env CRs after int testing completes

hongweiliu17:
(This pr copies most of code in https://github.com/redhat-appstudio/integration-service/pull/81)

After the integration testing pipeline completes(succeed, skipped), we need to delete the CRs created for it.

* delete the ephemeral env
* delete deploymentTarget
* delete deploymentTargetClaim
* delete snapshotenvironmentbinding

This PR reference the code added in https://github.com/redhat-appstudio/integration-service/pull/154 and so should be rebased after merging https://github.com/redhat-appstudio/integration-service/pull/154.

### DIFF
--- a/controllers/pipeline/pipeline_adapter.go
+++ b/controllers/pipeline/pipeline_adapter.go
@@ -237,6 +237,103 @@ func (a *Adapter) EnsureStatusReported() (reconciler.OperationResult, error) {
 	return reconciler.ContinueProcessing()
 }
 
+// EnsureEphemeralEnvironmentsCleanedUp will ensure that ephemeral environment(s) associated with the
+// integration PipelineRun are cleaned up.
+func (a *Adapter) EnsureEphemeralEnvironmentsCleanedUp() (reconciler.OperationResult, error) {
+	if !tekton.IsIntegrationPipelineRun(a.pipelineRun) || !helpers.HasPipelineRunFinished(a.pipelineRun) {
+		return reconciler.ContinueProcessing()
+	}
+
+	testEnvironment, err := a.getEnvironmentFromIntegrationPipelineRun(a.context, a.pipelineRun)
+	if err != nil {
+		a.logger.Error(err, "Failed to find the environment for the pipelineRun")
+		return reconciler.RequeueWithError(err)
+	} else if testEnvironment == nil {
+		a.logger.Info("The pipelineRun does not have any test Environments associated with it, skipping cleanup.")
+		return reconciler.ContinueProcessing()
+	}
+
+	isEphemeral := false
+	for _, tag := range testEnvironment.Spec.Tags {
+		if tag == "ephemeral" {
+			isEphemeral = true
+			break
+		}
+	}
+
+	if isEphemeral {
+		dtc, err := gitops.GetDeploymentTargetClaimForEnvironment(a.client, a.context, testEnvironment)
+		if err != nil || dtc == nil {
+			a.logger.Error(err, "Failed to find deploymentTargetClaim defined in environment", "environment.Name", testEnvironment.Name)
+			return reconciler.RequeueWithError(err)
+		}
+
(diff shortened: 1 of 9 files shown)


controllers/pipeline/pipeline_adapter.go
```diff
@@ -237,6 +237,103 @@ func (a *Adapter) EnsureStatusReported() (reconciler.OperationResult, error) {
 	return reconciler.ContinueProcessing()
 }
 
+// EnsureEphemeralEnvironmentsCleanedUp will ensure that ephemeral environment(s) associated with the
+// integration PipelineRun are cleaned up.
+func (a *Adapter) EnsureEphemeralEnvironmentsCleanedUp() (reconciler.OperationResult, error) {
+	if !tekton.IsIntegrationPipelineRun(a.pipelineRun) || !helpers.HasPipelineRunSucceeded(a.pipelineRun) {
```
dirgim, line 243: Hm, looking at this now I think that the existing PoC wasn't quite complete in this regard. 
We will need to also clean up the environments in the case when the pipeline fails, which means that we will need to modify the  [IntegrationOrBuildPipelineRunSucceededPredicate](https://github.com/redhat-appstudio/integration-service/blob/main/tekton/predicates.go#L28-L46) predicate to check if the pipelineRun finished instead of succeeded.

hongweiliu17: Thanks, I thought about to delete the env for the failing test. But later I think we might need to rerun the failing test and reuse the env. Should be recreate the env for hte rerun test?

dirgim: Yes, we will recreate the env for the time being. There will be a future feature for re-running tests which will refine that logic.
For now we want to make sure that there is no resource leakage by deleting all environments when a pipeline finishes, no matter its outcome (fail or pass).

hongweiliu17: Thanks very much! Let me revise it.

dirgim: Thanks!

controllers/pipeline/pipeline_adapter_test.go
```diff
@@ -825,5 +864,94 @@ var _ = Describe("Pipeline Adapter", Ordered, func() {
...
+			adapter = NewAdapter(testpipelineRunComponent, hasComp, hasApp, ctrl.Log, k8sClient, ctx)
+			Expect(reflect.TypeOf(adapter)).To(Equal(reflect.TypeOf(&Adapter{})))
+
+			Eventually(func() bool {
+				result, err := adapter.EnsureEphemeralEnvironmentsCleanedUp()
+				return !result.CancelRequest && err == nil
+			}, time.Second*20).Should(BeTrue())
+
+			Eventually(func() bool {
+				err := k8sClient.Get(ctx, types.NamespacedName{
```
dirgim, line 928: While this was the way this unit test was set up in the PoC, in retrospect this looks like a fragile way to check if the environment was deleted. 
Maybe getting all environments (like it's done e.g. [here](https://github.com/redhat-appstudio/integration-service/blob/main/controllers/snapshot/snapshot_adapter.go#L430-L434)) and checking that `len(environmentList.Items) == 0` would be better here?

dirgim: Alternatively, we could use this approach of [checking for apierr.IsNotFound](https://github.com/redhat-appstudio/managed-gitops/blob/main/appstudio-controller/controllers/appstudio.redhat.com/deploymenttargetbinder_controller_test.go#L111-L112) used in managed-gitops unit tests

hongweiliu17: Yes, thanks very much!

hongweiliu17: Updated.

dirgim: Thanks!

dirgim (approved):
LGTM

PR #162, opened 2023-04-27
feat(STONEINTG-161): remove HACBS keyword from integration-service

Josh-Everett:
Includes renaming of test output from HACBS_TEST_OUTPUT to TEST_OUTPUT.

### DIFF
--- a/controllers/pipeline/pipeline_controller_test.go
+++ b/controllers/pipeline/pipeline_controller_test.go
@@ -110,7 +110,7 @@ var _ = Describe("PipelineController", func() {
 				CompletionTime: &metav1.Time{Time: now.Add(5 * time.Minute)},
 				TaskRunResults: []tektonv1beta1.TaskRunResult{
 					{
-						Name: "HACBS_TEST_OUTPUT",
+						Name: "TEST_OUTPUT",
 						Value: *tektonv1beta1.NewArrayOrString(`{
 											"result": "SUCCESS",
 											"timestamp": "1665405318",
--- a/gitops/snapshot.go
+++ b/gitops/snapshot.go
@@ -73,32 +73,38 @@ const (
 	// PipelineAsCodeGitHubProviderType is the git provider type for a GitHub event which triggered the pipelinerun in build service.
 	PipelineAsCodeGitHubProviderType = "github"
 
-	//HACBSTestSuceededCondition is the condition for marking if the HACBS Tests succeeded for the Snapshot.
-	HACBSTestSuceededCondition = "HACBSTestSucceeded"
+	//AppStudioTestSuceededCondition is the condition for marking if the AppStudio Tests succeeded for the Snapshot.
+	AppStudioTestSuceededCondition = "AppStudioTestSucceeded"
 
-	// HACBSIntegrationStatusCondition is the condition for marking the HACBS integration status of the Snapshot.
-	HACBSIntegrationStatusCondition = "HACBSIntegrationStatus"
+	//LegacyTestSuceededCondition is the condition for marking if the AppStudio Tests succeeded for the Snapshot.
+	LegacyTestSuceededCondition = "HACBSStudioTestSucceeded"
 
-	// IntegrationTestScenarioValid is the condition for marking the HACBS integration status of the Scenario.
+	// AppStudioIntegrationStatusCondition is the condition for marking the AppStudio integration status of the Snapshot.
+	AppStudioIntegrationStatusCondition = "AppStudioIntegrationStatus"
+
+	// LegacyIntegrationStatusCondition is the condition for marking the AppStudio integration status of the Snapshot.
+	LegacyIntegrationStatusCondition = "HACBSIntegrationStatus"
+
+	// IntegrationTestScenarioValid is the condition for marking the AppStudio integration status of the Scenario.
 	IntegrationTestScenarioValid = "IntegrationTestScenarioValid"
 
-	// HACBSTestSuceededConditionPassed is the reason that's set when the HACBS tests succeed.
-	HACBSTestSuceededConditionPassed = "Passed"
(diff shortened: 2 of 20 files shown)


controllers/pipeline/pipeline_controller_test.go
```diff
@@ -110,7 +110,7 @@ var _ = Describe("PipelineController", func() {
 				CompletionTime: &metav1.Time{Time: now.Add(5 * time.Minute)},
 				TaskRunResults: []tektonv1beta1.TaskRunResult{
 					{
-						Name: "HACBS_TEST_OUTPUT",
+						Name: "TEST_OUTPUT",
```
MartinBasti, line 113: I saw a discussion about having it named as `APPSTUDIO_TEST_OUTPUT`

gitops/snapshot.go
```diff
@@ -194,51 +194,51 @@ func PrepareToRegisterIntegrationPipelineRun(snapshot *applicationapiv1alpha1.Sn
...
 	meta.SetStatusCondition(&snapshot.Status.Conditions, condition)
 }
 
-// IsSnapshotNotStarted checks if the HACBS Integration Status condition is not in progress status.
+// IsSnapshotNotStarted checks if the RHTAP Integration Status condition is not in progress status.
 func IsSnapshotNotStarted(snapshot *applicationapiv1alpha1.Snapshot) bool {
-	condition := meta.FindStatusCondition(snapshot.Status.Conditions, HACBSIntegrationStatusCondition)
-	if condition == nil || condition.Reason != HACBSIntegrationStatusInProgress {
+	condition := meta.FindStatusCondition(snapshot.Status.Conditions, RHTAPIntegrationStatusCondition)
+	if condition == nil || condition.Reason != RHTAPIntegrationStatusInProgress {
```
MartinBasti, line 211: may this incompatible change cause issues to ongoing requests because name of key changed?

```diff
-	// HACBSIntegrationStatusCondition is the condition for marking the HACBS integration status of the Snapshot.
-	HACBSIntegrationStatusCondition = "HACBSIntegrationStatus"
+	// RHTAPIntegrationStatusCondition is the condition for marking the RHTAP integration status of the Snapshot.
+	RHTAPIntegrationStatusCondition = "RHTAPIntegrationStatus
```

controllers/pipeline/pipeline_controller_test.go
dirgim: The final decision was to have it as just `TEST_OUTPUT` (see [slack announcement](https://redhat-internal.slack.com/archives/C054SA6DWHK/p1682523721822739))

helpers/integration.go
```diff
@@ -22,27 +22,31 @@ import (
 )
 
 const (
-	//HACBSTestOutputName is the name of the standardized HACBS Test output Tekton task result
-	HACBSTestOutputName = "HACBS_TEST_OUTPUT"
```
dirgim, line 26: We'll need to keep this constant so we can run in compatibility mode at least until after service preview.

dirgim: Never mind, totally missed the legacy name. Will have to put on my reading glasses :)

gitops/snapshot.go
dirgim: I think we'll need to support both statuses, the similar way to how we support the legacy `HACBS_TEST_OUTPUT` name.

dirgim:
As an update to requirements, the `HACBSTestSucceeded` status condition will need to be renamed to `AppStudioTestSucceeded` and `HACBSIntegrationStatus` will need to be renamed to `AppStudioIntegrationStatus`.

gitops/snapshot.go
```diff
@@ -194,51 +197,57 @@ func PrepareToRegisterIntegrationPipelineRun(snapshot *applicationapiv1alpha1.Sn
...
 	}
 	return false
 }
 
-// HaveHACBSTestsFinished checks if the HACBS tests have finished by checking if the HACBS Test Succeeded condition is set.
-func HaveHACBSTestsFinished(snapshot *applicationapiv1alpha1.Snapshot) bool {
-	statusCondition := meta.FindStatusCondition(snapshot.Status.Conditions, HACBSTestSuceededCondition)
+// HaveAppStudioTestsFinished checks if the AppStudio tests have finished by checking if the AppStudio Test Succeeded condition is set.
+func HaveAppStudioTestsFinished(snapshot *applicationapiv1alpha1.Snapshot) bool {
+	statusCondition := meta.FindStatusCondition(snapshot.Status.Conditions, AppStudioTestSuceededCondition)
```
Josh-Everett, line 237: Do we need to do a similar thing for `HaveAppStudioTestsFinished` and  `HaveAppStudioTestsSucceeded` to check for legacy conditions?

dirgim: Yes, I think we need to have the same thing for those functions.

status/status.go
```diff
@@ -11,7 +11,7 @@ import (
 )
 
 // NamePrefix is a common name prefix for this service.
-const NamePrefix = "HACBS Test"
+const NamePrefix = "AppStudio Test"
```
dirgim, line 14: I think that this will need to be the exception to the `AppStudio` usage based on [RHTAP-335](https://issues.redhat.com/browse/RHTAP-335) guidelines (and the [RHTAP-511](https://issues.redhat.com/browse/RHTAP-511) decision) and should be `Red Hat Trusted App`.

gitops/snapshot.go
dirgim: That seems to have done the trick

status/status.go
dirgim: Looks good, thanks

dirgim (approved):
LGTM

MartinBasti (approved):

PR #163, opened 2023-05-04
fix(STONEINTG-416): account for edge cases with namespace provisioning

dirgim:
* Change kubebuilder directives to use lowercase for CRs
* Update the role.yaml accordingly
* If environment doesn't get created, delete DTC
* Check if ITS environment config is defined

### DIFF
--- a/controllers/snapshot/snapshot_adapter.go
+++ b/controllers/snapshot/snapshot_adapter.go
@@ -522,15 +522,13 @@ func (a *Adapter) createCopyOfExistingEnvironment(existingEnvironment *applicati
 			"deploymentTargetClass.Name", deploymentTargetClass.Name)
 		return nil, fmt.Errorf("failed to create deploymentTargetClaim with deploymentTargetClass %s: %w", deploymentTargetClass.Name, err)
 	}
-	a.logger.Info("DeploymentTargetClaim is created for environment:",
-		"deploymentTargetClaim.NameSpace", dtc.Namespace,
-		"deploymentTargetClaim.Name", dtc.Name)
+	a.logger.LogAuditEvent("DeploymentTargetClaim is created for environment", dtc, h.LogActionAdd,
+		"integrationTestScenario.Name", integrationTestScenario.Name)
 
 	environment := gitops.NewCopyOfExistingEnvironment(existingEnvironment, namespace, integrationTestScenario, dtc.Name).
 		WithIntegrationLabels(integrationTestScenario).
 		WithSnapshot(snapshot).
 		AsEnvironment()
-
 	ref := ctrl.SetControllerReference(application, environment, a.client.Scheme())
 	if ref != nil {
 		a.logger.Error(ref, "Failed to set controller reference for Environment!",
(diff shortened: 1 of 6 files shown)


controllers/snapshot/snapshot_adapter.go
```diff
@@ -566,6 +566,11 @@ func (a *Adapter) createCopyOfExistingEnvironment(existingEnvironment *applicati
 
 	err = a.client.Create(a.context, environment)
 	if err != nil {
+		// We don't want to leave the new DTC on the cluster without the matching environment
+		dtcErr := a.client.Delete(a.context, dtc)
+		if dtcErr != nil {
+			return nil, fmt.Errorf("failed to delete dtc %s: %w", dtc.Name, err)
```
MartinBasti, line 572: This hides real error why env creation failed.

also this creation/deletion should have AuditLog IMHO

dirgim: Yep, all good points. I'm still testing out if this resolves the edge cases in the best way so didn't invest too much into refining it - will do that after confirming.

dheerajodha (approved):
Thanks for the PR!

controllers/snapshot/snapshot_adapter.go
```diff
@@ -566,8 +564,18 @@ func (a *Adapter) createCopyOfExistingEnvironment(existingEnvironment *applicati
 
 	err = a.client.Create(a.context, environment)
 	if err != nil {
-		return nil, fmt.Errorf("failed to create environment %s: %w", environment.Name, err)
+		// We don't want to leave the new DTC on the cluster without the matching environment
+		dtcErr := a.client.Delete(a.context, dtc)
+		if dtcErr != nil {
+			a.logger.Error(err, "failed to create ephemeral environment", "environment.name", environment.Name)
```
MartinBasti, line 570: This block puzzles me. It's about DTC deletion failure, but logger is logging failure of env creation. Shouldn't it before `if dtcErr != nil {` ?

controllers/snapshot/snapshot_adapter.go
```diff
@@ -566,8 +564,18 @@ func (a *Adapter) createCopyOfExistingEnvironment(existingEnvironment *applicati
...
+			return nil, fmt.Errorf("failed to delete dtc after creation of environment failed %s: %w", dtc.Name, dtcErr)
```
MartinBasti, line 571: Can we just log it as error and return only env creation error?


Or provide both errors
`failed to delete dtc %s: %w; failed to create env %s: %w", dtc.Name, dtcErr, env.Name, err`

MartinBasti: it should be also `DTC `

jsztuka:
lgtm, Martin has good points, I suppose those would be minor changes

controllers/snapshot/snapshot_adapter.go
dirgim: I did that to prevent both logging the message and returning the error, but I agree it's better to combine the errors into the same message, updated according to the other comment.

controllers/snapshot/snapshot_adapter.go
dirgim: Yes, good point, updated.

hongweiliu17 (approved):
Thanks for the PR! LGTM

jinqi7 (approved):
LGTM

MartinBasti (approved):

PR #164, opened 2023-05-04
Always log application name/namespace and request object name in logs

MartinBasti:

PR #165, opened 2023-05-08
fix(STONEINTG-414): don't add component without digest to snaphot

hongweiliu17:


hongweiliu17:
I would like to investigate to use mock for the unit test of newly added method but any review is appreciated.

gitops/snapshot.go
```diff
@@ -235,6 +237,17 @@ func IsSnapshotValid(snapshot *applicationapiv1alpha1.Snapshot) bool {
 	return false
 }
 
+// IsImageDigestValid checks if image url contains digest and it is valid image, return false if check fails
+func IsImageDigestValid(imageUrl string) bool {
+	ref, err := name.NewDigest(imageUrl)
+	if err != nil {
+		return false
+	}
+	_, err = remote.Get(ref)
```
dirgim, line 246: It looks like `remote.Get()` queries the registry, which is fine for public repositories, but would it cause issues when accessing private ones?

hongweiliu17: Good point. I get this page https://github.com/google/go-containerregistry/blob/main/pkg/authn/README.md and let me update the code to use `img, err := remote.Get(ref, remote.WithAuthFromKeychain(authn.DefaultKeychain))` and return `nil` if fails. I think it is supposed that `DOCKER_CONFIG` exist in the deployed environment.

gitops/snapshot.go
```diff
@@ -235,6 +237,17 @@ func IsSnapshotValid(snapshot *applicationapiv1alpha1.Snapshot) bool {
 	return false
 }
 
+// IsImageDigestValid checks if image url contains digest and it is valid image, return false if check fails
+func IsImageDigestValid(imageUrl string) bool {
```
MartinBasti, line 241: we are hiding what's wrong here, dropping useful information.

I suggest to change this bool function into validation function that returns `nil` when everything is ok and error message with details when validation failed. This error can by used by caller and logged in error log or passed forward

```suggestion
func ValidateImageDigest(imageUrl string) error {
```

controllers/pipeline/pipeline_adapter_test.go
```diff
@@ -82,12 +84,23 @@ var _ = Describe("Pipeline Adapter", Ordered, func() {
...
 		SampleRepoLink = "https://github.com/devfile-samples/devfile-sample-java-springboot-basic"
 		SampleCommit   = "a2ba645d50e471d5f084b"
 	)
 
+	SampleImageWithoutDigest = "quay.io/redhat-appstudio/sample-image"
+	imageRef, _ := name.ParseReference("quay.io/redhat-appstudio/sample-image" + ":test-index-pass")
+	imageDesc, _ := remote.Get(imageRef)
+	SampleDigest = imageDesc.Digest.String()
+	SampleImage = SampleImageWithoutDigest + "@" + SampleDigest
+
```
MartinBasti, line 103: Shall we have this in `const`?

gitops/snapshot.go
hongweiliu17: @dirgim If `DOCKER_CONFIG` or `${XDG_RUNTIME_DIR}` is not defined in the environment, should we require it to be provided in log, especially for the external users? Or we just verify the digest of containerImage is a vaid sha string?

gitops/snapshot.go
hongweiliu17: Actually at first I make this function return `error` when it is not a valid digest, but later I think the requirement says `only adds Components that have digests in their spec.containerImage field to new Snapshots.` and we just `continue` when `containerImage == ""` previously, so I change to just `continue` when the digest is not valid. I think it depends on the customer requirement, do they expect the error is thrown and stop creating snapshot when containerImage is not valid ("" or without Digest)? @dirgim

dirgim: I don't think we should stop creating the Snapshot, but I think we should log (warning/info) why we didn't add the component's image to the new Snapshot.

hongweiliu17: Ok, thank you and @MartinBasti,  let me log some warning for it.

controllers/pipeline/pipeline_adapter_test.go
hongweiliu17: After the discussion above, since we don't need to check if the image is valid, I think I can add it to `const` and use a valid sha string here. Thank you!

controllers/pipeline/pipeline_adapter.go
```diff
@@ -475,7 +475,9 @@ func (a *Adapter) prepareSnapshot(application *applicationapiv1alpha1.Applicatio
 		// which multiple components are being built in close succession.
 		// We omit this not-yet-built component from the snapshot rather than
 		// including a component that is incomplete.
-		if containerImage == "" {
+		// if the containerImage donesn't have a valid digest or is invalid, the
+		// the conponent will not be added to snapshot
+		if containerImage == "" || !gitops.IsImageDigestValid(containerImage) {
```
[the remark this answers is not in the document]
hongweiliu17: Thanks. Will update it.

gitops/snapshot.go
hongweiliu17: Updated.

gitops/snapshot.go
hongweiliu17: Updated.

controllers/pipeline/pipeline_adapter.go
```diff
@@ -476,6 +476,14 @@ func (a *Adapter) prepareSnapshot(application *applicationapiv1alpha1.Applicatio
 		// We omit this not-yet-built component from the snapshot rather than
 		// including a component that is incomplete.
 		if containerImage == "" {
+			logger.Error(nil, "component cannot be added to snapshot for application due to missing containerImage", "component.Name", applicationComponent.Name, "application.Name", application.Name)
+			continue
+		}
+		// if the containerImage donesn't have a valid digest, the conponent
```
dirgim, line 482: typo: `conponent` -> `component`

controllers/pipeline/pipeline_adapter.go
```diff
@@ -476,6 +476,14 @@ func (a *Adapter) prepareSnapshot(application *applicationapiv1alpha1.Applicatio
 		// We omit this not-yet-built component from the snapshot rather than
 		// including a component that is incomplete.
 		if containerImage == "" {
+			logger.Error(nil, "component cannot be added to snapshot for application due to missing containerImage", "component.Name", applicationComponent.Name, "application.Name", application.Name)
```
dirgim, line 479: I think we don't need to log the `application.Name` since #164 now automatically adds it to logs.

helpers/logs.go
```diff
@@ -29,7 +29,7 @@ const (
 	LogActionView   LogAction = iota + 1 // VIEW
 	LogActionAdd                         // ADD
 	LogActionUpdate                      // UPDATE
-	LogActionDelete                      // DELETE
+	LogActionDelete
```
dirgim, line 32: I'm guessing this is an accidental change.

gitops/snapshot.go
dirgim: Thanks, looks good!

controllers/pipeline/pipeline_adapter.go
hongweiliu17: Thanks!

controllers/pipeline/pipeline_adapter.go
hongweiliu17: Yes, thanks. But I need to add one line `logger := a.logger.WithApp(*application)` before calling `logger.Error()`.
Updated.

controllers/pipeline/pipeline_adapter.go
```diff
@@ -476,6 +476,14 @@ func (a *Adapter) prepareSnapshot(application *applicationapiv1alpha1.Applicatio
...
+			continue
+		}
+		// if the containerImage donesn't have a valid digest, the conponent
+		// will not be added to snapshot
+		err := gitops.ValidateImageDigest(containerImage)
+		if err != nil {
+			logger.Error(err, "component cannot added to snapshot for application due to invalid digest in containerImage", "component.Name", applicationComponent.Name, "application.Name", application.Name)
```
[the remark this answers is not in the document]
hongweiliu17: Updated.

helpers/logs.go
hongweiliu17: Yes, thanks.

hongweiliu17: Updated.

dirgim (approved):
LGTM

dheerajodha:
@hongweiliu17 if you like you can also test this PR by running the [build-via-appstudio.sh](https://github.com/redhat-appstudio/infra-deployments/blob/fb91c7021f53083615fe13bffef4bf1edccdf05f/hack/build/build-via-appstudio.sh) script as described in the "use case (2)" of [this slack thread](https://redhat-internal.slack.com/archives/C030SGP2VB9/p1683125228410729?thread_ts=1681393407.389439&cid=C030SGP2VB9) and confirm the expected behavior.

It's totally optional because you've already tested this in your tests here.

controllers/pipeline/pipeline_adapter.go
```diff
@@ -452,6 +452,7 @@ func (a *Adapter) getAllPipelineRunsForSnapshot(snapshot *applicationapiv1alpha1
 // prepareSnapshot prepares the Snapshot for a given application and the updated component (if any).
 // In case the Snapshot can't be created, an error will be returned.
 func (a *Adapter) prepareSnapshot(application *applicationapiv1alpha1.Application, component *applicationapiv1alpha1.Component, newContainerImage string, newComponentSource *applicationapiv1alpha1.ComponentSource) (*applicationapiv1alpha1.Snapshot, error) {
+	logger := a.logger.WithApp(*application)
```
MartinBasti, line 455: https://github.com/redhat-appstudio/integration-service/blob/e86f164b6828d7d29bfd87378f1f8ff74ed83463/controllers/pipeline/pipeline_controller.go#L115

application has been already added to logger, why is this needed here?

hongweiliu17: Thanks finding this. When I check the logs in my unittest, I found the application is not logged, so I add that line. You remind me, the controller called `.WithApp`, to see the log in my unittest, I should add that line to pipeline_adapter_test.go https://github.com/redhat-appstudio/integration-service/blob/ae921ce4e43b231117e29477f7f3e1164ea0b9b7/controllers/pipeline/pipeline_adapter_test.go#L138

hongweiliu17:
> @hongweiliu17 if you like you can also test this PR by running the [build-via-appstudio.sh](https://github.com/redhat-appstudio/infra-deployments/blob/fb91c7021f53083615fe13bffef4bf1edccdf05f/hack/build/build-via-appstudio.sh) script as described in the "use case (2)" of [this slack thread](https://redhat-internal.slack.com/archives/C030SGP2VB9/p1683125228410729?thread_ts=1681393407.389439&cid=C030SGP2VB9) and confirm the expected behavior.
> 
> It's totally optional because you've already tested this in your tests here.

Totally and strongly agree to deploy and test it. Thanks!

MartinBasti (approved):

PR #166, opened 2023-05-08
fix(STONEINTG-408): snapshot status get updated when pipelineRun failed

jinqi7:


helpers/integration.go
```diff
@@ -236,6 +236,8 @@ func CalculateIntegrationPipelineRunOutcome(adapterClient client.Client, ctx con
 		if err != nil {
 			return false, fmt.Errorf("error while getting test results from pipelineRun %s: %w", pipelineRun.Name, err)
 		}
+	} else if HasPipelineRunFinished(pipelineRun) && !HasPipelineRunSucceeded(pipelineRun) {
+		return false, fmt.Errorf("PipelineRun %s in namespace %s failed for %s", pipelineRun.Name, pipelineRun.Namespace, GetPipelineRunFailedReason(pipelineRun))
```
dirgim, line 240: I think that if we return an error here, it will break the loop [here](https://github.com/redhat-appstudio/integration-service/blob/e86f164b6828d7d29bfd87378f1f8ff74ed83463/controllers/pipeline/pipeline_adapter.go#L391-L394) and ultimately get picked up by [this check](https://github.com/redhat-appstudio/integration-service/blob/e86f164b6828d7d29bfd87378f1f8ff74ed83463/controllers/pipeline/pipeline_adapter.go#L165-L169), ending the Reconciliation loop and not updating the Snapshot status as a result.

Maybe we can log the output of `GetPipelineRunFailedReason(pipelineRun)` before returning `false, nil` here instead?

jinqi7: Updated. Thanks!

helpers/integration.go
```diff
@@ -236,6 +236,9 @@ func CalculateIntegrationPipelineRunOutcome(adapterClient client.Client, ctx con
 		if err != nil {
 			return false, fmt.Errorf("error while getting test results from pipelineRun %s: %w", pipelineRun.Name, err)
 		}
+	} else if HasPipelineRunFinished(pipelineRun) && !HasPipelineRunSucceeded(pipelineRun) {
```
dirgim, line 239: Thinking about this some more, having this condition in the `else if` block means that a failed pipeline with one passing task would be considered as passed overall. 
I think it might be better to first check if the pipeline succeeded and then check the task child references (meaning swapping [this condition](https://github.com/redhat-appstudio/integration-service/pull/166/files#diff-ea27f7cada77e890e9c4d991d8edc8f9c6bfbbf2a830cf74df4ba606341240caR239) with  [this one](https://github.com/redhat-appstudio/integration-service/pull/166/files#diff-ea27f7cada77e890e9c4d991d8edc8f9c6bfbbf2a830cf74df4ba606341240caR233).

jinqi7: Thinking about this some more, having this condition in the else if block means that a failed pipeline with one passing task would be considered as passed overall.                        => This situation was not considered before and

dirgim (approved):
LGTM

helpers/integration.go
dirgim: Thanks!

MartinBasti (approved):

jsztuka (approved):
lgtm

hongweiliu17:
Should we also update `!h.HasPipelineRunSucceeded(a.pipelineRun)` in https://github.com/redhat-appstudio/integration-service/blob/main/controllers/pipeline/pipeline_adapter.go#LL121C56-L121C96 to `!h.HasPipelineRunFinished(a.pipelineRun)` to handle the failed pipelinerun as well? Like https://github.com/redhat-appstudio/integration-service/blob/main/controllers/pipeline/pipeline_adapter.go#L239

PR #167, opened 2023-05-09
fix(STONEINTG-419): Release pipeline run fails on missing attestation

MartinBasti:
PipelineRun are signed with tekton chains after they finish. We are
hitting race condition than sometimes we are creating release without
signed pipeline and cause error mentioned in subject.
Reconcile only build pipelines which have been signed
by tekton chains to avoid this issue.

Signed pipelineRuns are marked with annotation chains.tekton.dev/signed

### DIFF
--- a/tekton/predicates.go
+++ b/tekton/predicates.go
@@ -5,9 +5,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 )
 
-// IntegrationPipelineRunStartedPredicate returns a predicate which filters out all objects except
-// integration PipelineRuns that have just started.
-func IntegrationPipelineRunStartedPredicate() predicate.Predicate {
+// IntegrationPipelineRunPredicate returns a predicate which filters out all objects except
+// integration PipelineRuns that have just started or finished.
+func IntegrationPipelineRunPredicate() predicate.Predicate {
 	return predicate.Funcs{
 		CreateFunc: func(createEvent event.CreateEvent) bool {
 			return false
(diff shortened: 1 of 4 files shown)


MartinBasti:
A poor man proof that signed pipeline run passed through filter into reconcilation
```
{"level":"info","ts":"2023-05-10T12:57:32Z","logger":"controllers.pipeline","caller":"pipeline/pipeline_adapter.go:73","msg":"The pipelineRun does not have any component associated with it, will not create a new Snapshot.","pipelineRun":"default/fbc-builder-run-5nz8g","application":"/"}
{"level":"info","ts":"2023-05-10T12:59:24Z","logger":"controllers.pipeline","caller":"pipeline/pipeline_adapter.go:73","msg":"The pipelineRun does not have any component associated with it, will not create a new Snapshot.","pipelineRun":"default/java-builder-run-kqkbv","application":"/"}
```

dirgim (approved):
LGTM

tekton/predicates.go
```diff
@@ -39,8 +39,8 @@ func IntegrationOrBuildPipelineRunFinishedPredicate() predicate.Predicate {
 			return false
 		},
 		UpdateFunc: func(e event.UpdateEvent) bool {
-			return (IsIntegrationPipelineRun(e.ObjectNew) || IsBuildPipelineRun(e.ObjectNew)) &&
-				hasPipelineRunStateChangedToFinished(e.ObjectOld, e.ObjectNew)
+			return (IsBuildPipelineRun(e.ObjectNew)) &&
+				hasPipelineRunBeenChangedToSigned(e.ObjectOld, e.ObjectNew) // this signes only finished pipelines
```
dheerajodha, line 43:
```suggestion
				hasPipelineRunBeenChangedToSigned(e.ObjectOld, e.ObjectNew) // this signs only finished pipelines
```
minor typo

dheerajodha, line 43: Would this capture the PLRs which have finished?
The methods `IsBuildPipelineRun()` and `hasPipelineRunBeenChangedToSigned()` don't check if the PLR is finished, what am I missing?

MartinBasti: maybe I should rephrase the comment, while I'm reading it now, looks little bit braindead

Josh-Everett:
Other than minor typo LGTM. I have no issues with comment as is.

tekton/predicates.go
dheerajodha: no im braindead, the comment makes sense now but can also be changed to something like `// only finished pipelines are signed` because im not sure what "this" refers to in the comment.

tekton/predicates.go
```diff
@@ -39,8 +39,8 @@ func IntegrationOrBuildPipelineRunFinishedPredicate() predicate.Predicate {
 			return false
 		},
 		UpdateFunc: func(e event.UpdateEvent) bool {
-			return (IsIntegrationPipelineRun(e.ObjectNew) || IsBuildPipelineRun(e.ObjectNew)) &&
```
jsztuka, line 42: The reason behind removing IsIntegrationPipelineRun is due to predicate??

MartinBasti: I separated predicates for Build and Integration pipeline pipelines, so now it's part of IntegrationPipelineRunPredicate()

tekton/predicates.go
MartinBasti: updated

dheerajodha (approved):
lgtm

PR #199, opened 2023-05-31
Drop cert manager dependency from webhook.

jsztuka:
Following changes inject CA bundle directly to the webhook so its not needed to be provided by third party operator.

### DIFF
--- a/config/webhook/manifests.yaml
+++ b/config/webhook/manifests.yaml
@@ -1,0 +1,79 @@
+
+---
+apiVersion: admissionregistration.k8s.io/v1
+kind: MutatingWebhookConfiguration
+metadata:
+  creationTimestamp: null
+  name: mutating-webhook-configuration
+webhooks:
+- admissionReviewVersions:
+  - v1alpha1
+  - v1beta1
+  clientConfig:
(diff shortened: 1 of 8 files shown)


config/webhook/manifests.yaml
```diff
@@ -0,0 +1,79 @@
+
```
MartinBasti, line 1: I'm puzzled how it this change related to injecting CA cert, can you please provide me more insights?

jsztuka: It should specify how a webhook should be served, I need to confirm this information however. Will post it here.

dirgim: Maybe you can have this change in a separate commit in order to differentiate from removing dependency on the CA cert?

MartinBasti (approved):

config/webhook/manifests.yaml
jsztuka: Commit added.

dirgim: Thanks!

dirgim (approved):
LGTM

PR #200, opened 2023-05-31
chore(STONEINTG-379): use go 1.19

MartinBasti:
Go 1.18 is quite old and our dependecies already requires go 1.19 as min version. Moving to 1.19 will allow us to use latest versions of libraries.


dirgim (approved):
LGTM

dheerajodha (approved):
/lgtm

PR #202, opened 2023-06-01
fix(STONEINTG-434): allow for snapshots without valid component defined in label

hongweiliu17:

### DIFF
--- a/controllers/snapshot/snapshot_controller.go
+++ b/controllers/snapshot/snapshot_controller.go
@@ -86,8 +86,10 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	logger = logger.WithApp(*application)
 
 	component, err := loader.GetComponentFromSnapshot(r.Client, ctx, snapshot)
-	if err != nil {
-		logger.Error(err, "Failed to get Component from the Snapshot")
+	if err != nil && errors.IsNotFound(err) {
+		logger.Info("Failed to get Component defined in snapshot label appstudio.openshift.io/component, use nil value")
+	} else if err != nil && !errors.IsNotFound(err) {
+		logger.Error(err, "Error when getting Component from the Snapshot")
 		return ctrl.Result{}, err
 	}
 
--- a/controllers/snapshot/snapshot_controller_test.go
+++ b/controllers/snapshot/snapshot_controller_test.go
@@ -23,13 +23,16 @@ import (
 
 var _ = Describe("SnapshotController", func() {
 	var (
-		manager            ctrl.Manager
-		snapshotReconciler *Reconciler
-		scheme             runtime.Scheme
-		req                ctrl.Request
-		hasApp             *applicationapiv1alpha1.Application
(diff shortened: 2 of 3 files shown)


hongweiliu17:
Look at GetComponentFromSnapshot() called in [snapshot_controller.go](https://github.com/redhat-appstudio/integration-service/blob/9ae5f506922ba9a1a9f47b0edb135b3bbc6b8b53/controllers/snapshot/snapshot_controller.go#L86-L90)
```
func GetComponentFromSnapshot(c client.Client, ctx context.Context, snapshot *applicationapiv1alpha1.Snapshot) (*applicationapiv1alpha1.Component, error) {
    if componentLabel, ok := snapshot.Labels[gitops.SnapshotComponentLabel]; ok {
        component := &applicationapiv1alpha1.Component{}
        err := c.Get(ctx, types.NamespacedName{
            Namespace: snapshot.Namespace,
            Name:      componentLabel,
        }, component)        
        if err != nil {
            return nil, err
        }
        return component, nil
    } else {
        return nil, nil
    }
}
```

If label `Labels[gitops.SnapshotComponentLabel]` exists but that component doesn't exist at all in cluster,  `nil, err` will be returned to [snapshot_controller.go](https://github.com/redhat-appstudio/integration-service/blob/9ae5f506922ba9a1a9f47b0edb135b3bbc6b8b53/controllers/snapshot/snapshot_controller.go#L86-L90), we will log error `Failed to get Component from the Snapshot` in [code](https://github.com/redhat-appstudio/integration-service/blob/9ae5f506922ba9a1a9f47b0edb135b3bbc6b8b53/controllers/snapshot/snapshot_controller.go#L88),  pr https://github.com/redhat-appstudio/integration-service/pull/202 is committed to allow this nonexisting component. But is  this a valid scenario that component defined in label `Labels[gitops.SnapshotComponentLabel]` doesn't exist? @dirgim 

If label `Labels[gitops.SnapshotComponentLabel]` doesn't exist, `nil, nil` will be returned to [snapshot_controller.go](https://github.com/redhat-appstudio/integration-service/blob/9ae5f506922ba9a1a9f47b0edb135b3bbc6b8b53/controllers/snapshot/snapshot_controller.go#L86-L90), I this this is an expected result, for composite snapshot.

hongweiliu17:
Hi @dirgim , second question. If no valid component defined in snapshot label `Labels[gitops.SnapshotComponentLabel]`, we can always get component in [snapshot_adapter.go function GetAllApplicationComponents()](https://github.com/redhat-appstudio/integration-service/blob/main/controllers/snapshot/snapshot_adapter.go#L213) for this snapshot when creating binding, right?  I am confused. Thank you!

dirgim:
@hongweiliu17 you are correct,  if the label is present on the Snapshot, the component should exist. 
The Integration service already acts correctly by reporting an error if the label is there and the component doesn't exist, so this bugfix is not needed.
The SnapshotEnvironmentBinding can always be created because it does not depend on the component label.
Thank you for analyzing the issue and raising the questions.

hongweiliu17:
> 

Thank you Kruno! Let me close this PR.

PR #205, opened 2023-06-06
feat(STONEINTG-458): build pipelineRuns are annotated with Snapshots

dirgim:
* Build pipelineRuns are annotated with created Snapshots
* Expand build pipelineRun predicate to ignore annotated pipelineRuns
* Only require that build pipelineRun is signed by chains operator
* Require that the build pipelineRun is succeeded

### DIFF
--- a/controllers/pipeline/pipeline_adapter.go
+++ b/controllers/pipeline/pipeline_adapter.go
@@ -100,6 +100,15 @@ func (a *Adapter) EnsureSnapshotExists() (reconciler.OperationResult, error) {
 		a.logger.Info("Found existing Snapshot",
 			"snapshot.Name", existingSnapshot.Name,
 			"snapshot.Spec.Components", existingSnapshot.Spec.Components)
+		// Annotate the build pipelineRun with the existing Snapshot if it hasn't been already annotated
+		if _, found := a.pipelineRun.ObjectMeta.Annotations[tekton.SnapshotNameLabel]; !found {
+			a.pipelineRun, err = a.annotateBuildPipelineRunWithSnapshot(a.pipelineRun, existingSnapshot)
+			if err != nil {
+				a.logger.Error(err, "Failed to update the build pipelineRun with new annotations",
+					"pipelineRun.Name", a.pipelineRun.Name)
+				return reconciler.RequeueWithError(err)
+			}
+		}
 		return reconciler.ContinueProcessing()
 	}
 
@@ -113,16 +122,11 @@ func (a *Adapter) EnsureSnapshotExists() (reconciler.OperationResult, error) {
 		"snapshot.Name", expectedSnapshot.Name,
 		"snapshot.Spec.Components", expectedSnapshot.Spec.Components)
 
-	patch := client.MergeFrom(a.pipelineRun.DeepCopy())
-	h.AddAnnotation(&a.pipelineRun.ObjectMeta, tekton.SnapshotNameLabel, expectedSnapshot.Name)
-	err = a.client.Patch(a.context, a.pipelineRun, patch)
+	a.pipelineRun, err = a.annotateBuildPipelineRunWithSnapshot(a.pipelineRun, expectedSnapshot)
 	if err != nil {
 		a.logger.Error(err, "Failed to update the build pipelineRun with new annotations",
 			"pipelineRun.Name", a.pipelineRun.Name)
-		return reconciler.RequeueWithError(err)
 	}
-	a.logger.LogAuditEvent("Updated build pipelineRun", a.pipelineRun, h.LogActionUpdate,
-		"snapshot.Name", expectedSnapshot.Name)
 
 	return reconciler.ContinueProcessing()
 }
@@ -659,4 +663,16 @@ func (a *Adapter) findMatchingSnapshot(application *applicationapiv1alpha1.Appli
 		}
 	}
 	return nil, nil
+}
+
+func (a *Adapter) annotateBuildPipelineRunWithSnapshot(pipelineRun *tektonv1beta1.PipelineRun, snapshot *applicationapiv1alpha1.Snapshot) (*tektonv1beta1.PipelineRun, error) {
+	patch := client.MergeFrom(pipelineRun.DeepCopy())
+	h.AddAnnotation(&pipelineRun.ObjectMeta, tekton.SnapshotNameLabel, snapshot.Name)
+	err := a.client.Patch(a.context, pipelineRun, patch)
+	if err != nil {
+		return pipelineRun, err
+	}
+	a.logger.LogAuditEvent("Updated build pipelineRun", pipelineRun, h.LogActionUpdate,
+		"snapshot.Name", snapshot.Name)
+	return pipelineRun, nil
 }
--- a/controllers/pipeline/pipeline_adapter.go
+++ b/controllers/pipeline/pipeline_adapter.go
@@ -113,6 +113,17 @@ func (a *Adapter) EnsureSnapshotExists() (reconciler.OperationResult, error) {
 		"snapshot.Name", expectedSnapshot.Name,
 		"snapshot.Spec.Components", expectedSnapshot.Spec.Components)
 
+	patch := client.MergeFrom(a.pipelineRun.DeepCopy())
+	h.AddAnnotation(&a.pipelineRun.ObjectMeta, tekton.SnapshotNameLabel, expectedSnapshot.Name)
+	err = a.client.Patch(a.context, a.pipelineRun, patch)
+	if err != nil {
+		a.logger.Error(err, "Failed to update the build pipelineRun with new annotations",
+			"pipelineRun.Name", a.pipelineRun.Name)
+		return reconciler.RequeueWithError(err)
+	}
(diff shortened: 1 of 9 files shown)


tekton/predicates_test.go
```diff
@@ -96,8 +95,23 @@ var _ = Describe("Predicates", func() {
 			Expect(instance.Update(contextEvent)).To(BeFalse())
 
 			newPipelineRun.Annotations["chains.tekton.dev/signed"] = "true"
-			Expect(instance.Update(contextEvent)).To(BeTrue())
-			contextEvent.ObjectNew = &tektonv1beta1.TaskRun{}
+			Expect(instance.Update(contextEvent)).To(BeFalse())
+		})
+
+		It("should return false when an updated event is received for a failed PipelineRun and no signed", func() {
+			// also failed pipelines are signed by tekton chains, test it
```
dheerajodha, line 102:
```suggestion
		It("should return false when an updated event is received for a succeeded PipelineRun and no signed", func() {
```

tekton/predicates.go
```diff
@@ -39,8 +40,7 @@ func BuildPipelineRunFinishedPredicate() predicate.Predicate {
 			return false
 		},
 		UpdateFunc: func(e event.UpdateEvent) bool {
-			return (IsBuildPipelineRun(e.ObjectNew)) &&
-				hasPipelineRunBeenChangedToSigned(e.ObjectOld, e.ObjectNew) // only finished pipelines are signed
+			return IsBuildPipelineRun(e.ObjectNew) && isPipelineRunSigned(e.ObjectNew) && helpers.HasPipelineRunSucceeded(e.ObjectNew)
```
dheerajodha, line 43: Also, can we change the name of this predicate to something like `BuildPipelineRunSignedAndSucceededPredicate()`

controllers/pipeline/pipeline_adapter.go
```diff
@@ -113,6 +113,18 @@ func (a *Adapter) EnsureSnapshotExists() (reconciler.OperationResult, error) {
...
 
+	patch := client.MergeFrom(a.pipelineRun.DeepCopy())
+	h.AddAnnotation(&a.pipelineRun.ObjectMeta, tekton.SnapshotNameLabel, expectedSnapshot.Name)
+	err = a.client.Patch(a.context, a.component, patch)
+	if err != nil {
+		a.logger.Error(err, "Failed to update the build pipelineRun with new annotations",
+			"pipelineRun.Name", a.pipelineRun.Name)
+		return reconciler.RequeueWithError(err)
+	}
+	a.logger.LogAuditEvent("Updated build pipelineRun", a.pipelineRun, h.LogActionUpdate,
```
hongweiliu17, line 124: Do we need this `"pipelineRun.Name", a.pipelineRun.Name,` in the a.logger.LogAuditEvent()? Others look good to me.

hongweiliu17 (approved):
LGTM

tekton/predicates_test.go
dirgim: Ack, updated, thanks for the catch

controllers/pipeline/pipeline_adapter.go
dirgim: You are right, that line is not needed because the resource name gets logged automatically. I removed it now, thanks.

tekton/predicates_test.go
dheerajodha: Should we also change `failed PipelineRun` to `succeeded PipelineRun`?

dirgim: Yes, totally missed it on the first pass, updated

dheerajodha: Thank you!

loader/loader.go
```diff
@@ -434,3 +435,23 @@ func (l *loader) GetAllSnapshots(c client.Client, ctx context.Context, applicati
...
+func (l *loader) GetSnapshotFromPipelineRunAnnotation(c client.Client, ctx context.Context, pipelineRun *tektonv1beta1.PipelineRun) (*applicationapiv1alpha1.Snapshot, error) {
+	if snapshotName, found := pipelineRun.ObjectMeta.Annotations[tekton.SnapshotNameLabel]; found {
+		snapshot := &applicationapiv1alpha1.Snapshot{}
+		err := c.Get(ctx, types.NamespacedName{
+			Namespace: pipelineRun.Namespace,
+			Name:      snapshotName,
+		}, snapshot)
+
+		if err != nil {
+			return nil, err
```
MartinBasti, line 450: Missing error chaining?

MartinBasti:
I still don't understand how this code should work.

It will get request for every change in buildpipeline, after pipeline is fnished. It will have configured snapshot, so it will go to `return reconciler.ContinueProcessing()` but then it will just execute next steps in adapter. We will be spoiling logs with many events and doing more actions, which are unwanted. 
I don't see how this ensure that code is executed only once.

Also presence of code for handling case that snapshot is already defined is worrying me, can this case happen now?

tekton/predicates.go
dheerajodha: So now that we're only processing the signed & succeeded build PLRs, should the test logic be updated [here](https://github.com/redhat-appstudio/e2e-tests/blob/f25ebe3fbac695690975539054361aeeb11e5a6f/tests/integration-service/integration.go#L185) to `assertBuildPipelineRunSignedAndSucceeded()`?

dirgim:
> I still don't understand how this code should work.
> 
> It will get request for every change in buildpipeline, after pipeline is fnished. It will have configured snapshot, so it will go to `return reconciler.ContinueProcessing()` but then it will just execute next steps in adapter. We will be spoiling logs with many events and doing more actions, which are unwanted. I don't see how this ensure that code is executed only once.
> 
> Also presence of code for handling case that snapshot is already defined is worrying me, can this case happen now?

@MartinBasti you are right, I've come to the same conclusion while testing the changes today. 
While this code does work, it creates too much churn in the controller. I've been working on moving the check for the Snapshot annotation into the predicate so we skip reconciling a pipelineRun which was annotated with the Snapshot.
I updated the code with the changes, please take another look.

loader/loader.go
dirgim: Removed this function in the newest version

MartinBasti:
yup, now it looks better, but I'll take time tomorrow to think more about potential unexpected consequences

controllers/pipeline/pipeline_adapter.go
hongweiliu17: Thanks.

hongweiliu17 (approved):
LGTM

tekton/predicates.go
dirgim: Yes, I think that would be a good idea.

MartinBasti (approved):

jinqi7 (approved):
LGTM

controllers/pipeline/pipeline_adapter.go
```diff
@@ -100,6 +100,15 @@ func (a *Adapter) EnsureSnapshotExists() (reconciler.OperationResult, error) {
 		a.logger.Info("Found existing Snapshot",
 			"snapshot.Name", existingSnapshot.Name,
 			"snapshot.Spec.Components", existingSnapshot.Spec.Components)
+		// Annotate the build pipelineRun with the existing Snapshot if it hasn't been already annotated
+		if _, found := a.pipelineRun.ObjectMeta.Annotations[tekton.SnapshotNameLabel]; !found {
+			a.pipelineRun, err = a.annotateBuildPipelineRunWithSnapshot(a.pipelineRun, existingSnapshot)
+			if err != nil {
+				a.logger.Error(err, "Failed to update the build pipelineRun with new annotations",
+					"pipelineRun.Name", a.pipelineRun.Name)
+				return reconciler.RequeueWithError(err)
```
dheerajodha, line 109: 2 questions:

1. So now we give the controller **one more** chance to attach this annotation to the build PLR before we requeue the reconciliation loop with an error, right?
2. Consider the scenario when for a given PLR a snapshot was created but there was an issue during adding the annotation. So, when this PLR gets reconciled the next time, let's say that it is now not the most recently succeeded PLR so it will exit this function on line 87 above. This would result in some build PLRs not having this annotation, would that be a problem?

tekton/predicates.go
dheerajodha: Okay, I'll create a separate task for this.

controllers/pipeline/pipeline_adapter.go
dirgim: I'd say that the chances of that are getting quite slim at that point - requeueing with error means that the controller will attempt another reconcile for the object within seconds (or less), so there would need to be a seriously quick succession of builds of the same component for that to happen.
In any case, if there is indeed a newer build for the component, then the Snapshot that we're trying to reconcile would get superseded by the new build anyways.

dheerajodha (approved):
/lgtm

controllers/pipeline/pipeline_adapter.go
dheerajodha: > then the pipelineRun/Snapshot combination that we're trying to reconcile would get superseded by the new build anyways.

yup, and the previous build PLR would still not have the annotation on it, but that's not a problem I think.

PR #206, opened 2023-06-07
fix(STONEINTG-185): remove prometheus service monitor

MartinBasti:
Prometheus service monitor config is now part infra-deployments

https://github.com/redhat-appstudio/infra-deployments/pull/1920


dirgim (approved):
LGTM

PR #209, opened 2023-06-08
STONEINTG-401 - merge feature branch into main

jsztuka:
This PR will enrich integration-service capabilities regarding integration tests.
It adds github resolvers to the integrationTestScenarios together with webhook, that will migrate CRDs from version 
v1alpha1 to v1beta1.


config/crd/patches/webhook_in_integrationscenarios.yaml
```diff
@@ -7,10 +7,9 @@ spec:
   conversion:
     strategy: Webhook
     webhook:
+      conversionReviewVersions: ["v1alpha1", "v1beta1"]
```
jinqi7, line 10: This file is similar with webhook_in_integrationtestscenarios.yaml. What is the reason to keep them both?

tekton/integration_pipeline_test.go
```diff
@@ -56,11 +60,26 @@ var _ = Describe("Integration pipeline", func() {
...
 				Application: "application-sample",
-				Bundle:      "quay.io/kpavic/test-bundle:component-pipeline-pass",
-				Pipeline:    "component-pipeline-pass",
-				Environment: v1alpha1.TestEnvironment{
+				ResolverRef: v1beta1.ResolverRef{
+					Resolver: "git",
+					Params: []v1beta1.ResolverParameter{
+						{
+							Name:  "url",
+							Value: "https://github.com/redhat-appstudio/integration-examples.git",
```
dheerajodha, line 70: Optional: What do you think about a new ITS using Git resolver which points to `integration_pipelinerun_fail.yaml`? to test the negative scenario in unit tests

config/default/kustomization.yaml
```diff
@@ -36,39 +36,39 @@ patchesStrategicMerge:
...
+#     kind: Service
+#     version: v1
+#     name: webhook-service
+#   fieldref:
+#     fieldpath: metadata.namespace
+# - name: SERVICE_NAME
+#   objref:
+#     kind: Service
+#     version: v1
+#     name: webhook-service
```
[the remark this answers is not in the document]
dirgim: The default generated file has these lines commented out, so those are generally kept in the file. I think we can remove the extra spaces that were added in this PR and leave them as-is.

config/crd/patches/webhook_in_integrationscenarios.yaml
dirgim: Yes, I think it should be possible to remove this extra file without affecting anything.

tekton/integration_pipeline_test.go
dirgim: Since the integration Tekton pipelines are not actually executed in unit tests (only their CRs are created), I don't think that it would make a difference which pipeline definition it points to, apart from having some visibility into which pipelines exist in our examples repo.

config/crd/patches/webhook_in_integrationscenarios.yaml
dirgim: Removed the unused files, please take a look

config/default/kustomization.yaml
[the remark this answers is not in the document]
dirgim: Made the change, please take a look

controllers/binding/snapshotenvironmentbinding_controller_test.go
```diff
@@ -157,12 +158,27 @@ var _ = Describe("BindingController", func() {
...
+							Value: "main",
+						},
+						{
+							Name:  "pathInRepo",
+							Value: "pipelineruns/integration_pipelinerun_pass.yaml",
+						},
+					},
+				},
+				Environment: v1beta1.TestEnvironment{
+					Name: "envname-int",
```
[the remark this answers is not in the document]
dirgim: Changed back to `hasEnv.name`, please take a look

tekton/integration_pipeline_test.go
```diff
@@ -185,10 +290,28 @@ var _ = Describe("Integration pipeline", func() {
...
 		}
 		Expect(k8sClient.Create(ctx, hasComp)).Should(Succeed())
+
+		newIntegrationBundlePipelineRun = tekton.NewIntegrationPipelineRun(prefix, namespace, *integrationTestScenarioBundle).
+			WithIntegrationLabels(integrationTestScenarioBundle).
+			WithSnapshot(hasSnapshot).
+			WithApplicationAndComponent(hasApp, hasComp)
+		Expect(k8sClient.Create(ctx, newIntegrationBundlePipelineRun.AsPipelineRun())).Should(Succeed())
+
+		enterpriseContractPipelineRun = tekton.NewIntegrationPipelineRun(prefix, namespace, *integrationTestScenarioGit).
```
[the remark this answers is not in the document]
dirgim: Ack, updated

tekton/integration_pipeline_test.go
dirgim: I think we can skip adding this for now.

config/default/kustomization.yaml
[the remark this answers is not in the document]
dheerajodha: thanks!

controllers/binding/snapshotenvironmentbinding_controller_test.go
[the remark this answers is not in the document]
dheerajodha: thanks!

dirgim (approved):
LGTM

sonam1412 (approved):
LGTM

config/crd/patches/webhook_in_integrationscenarios.yaml
jsztuka: No reason at all, we can drop it, this file is not used.

PR #210, opened 2023-06-08
fix(STONEINTG-185): remove cluster-monitring label

MartinBasti:
This label is not supposed to be included in manager config as it may be monitored by user workload prometheus instance


MartinBasti:
described here https://github.com/redhat-appstudio/infra-deployments/blob/main/components/monitoring/README.md#2-service-monitors

dirgim (approved):
LGTM

jinqi7 (approved):

PR #214, opened 2023-06-13
feat: enable podman support in Makefile

14rcole:
Our Makefile only supports docker.  Since podman is developed at Red Hat cri-o is the default OpenShift runtime, we should also make sure we support development using podman.  This is just a quick change to the Makefile and documentation to enable users to override the default docker container runtime via the 'CONT_ENGINE' environment variable.

Note: I grepped for references to `docker-build` and `docker-push`.  I saw some pipelineRefs in `.tekton` named `docker-build`.  Should those be renamed too?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,6 +54,9 @@ IMG ?= $(IMAGE_TAG_BASE):$(TAG_NAME)
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION = 1.23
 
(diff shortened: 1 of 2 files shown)


MartinBasti (approved):
<3

MartinBasti:
> I saw some pipelineRefs in .tekton named docker-build. Should those be renamed too?

Not in this PR. It would require to change the task definition (somewhere else) and make new bundle, and also keep compatibility with other repos.

dirgim (approved):
LGTM

PR #218, opened 2023-06-14
fix(STONEINTG-471): add scenario label to binding created for ephemer…

hongweiliu17:
…al environment

### DIFF
--- a/controllers/snapshot/snapshot_adapter.go
+++ b/controllers/snapshot/snapshot_adapter.go
@@ -172,7 +172,12 @@ func (a *Adapter) EnsureCreationOfEnvironment() (reconciler.OperationResult, err
 	allEnvironments, err := a.loader.GetAllEnvironments(a.client, a.context, a.application)
 	if err != nil {
 		a.logger.Error(err, "Failed to get all environments.")
-		return reconciler.RequeueOnErrorOrStop(err)
+		return reconciler.RequeueWithError(err)
+	}
+	components, err := a.loader.GetAllApplicationComponents(a.client, a.context, a.application)
+	if err != nil {
+		a.logger.Error(err, "Failed to get all components.")
+		return reconciler.RequeueWithError(err)
 	}
 
 TestScenarioLoop:
@@ -188,6 +193,47 @@ TestScenarioLoop:
 				a.logger.Info("Environment already exists and contains snapshot and scenario:",
 					"environment.Name", environment.Name,
 					"integrationScenario.Name", integrationTestScenario.Name)
+
+				//check if the environmentSnapshotBinding exists for this existing environment, create it if it doesn't exist
+				var binding = &applicationapiv1alpha1.SnapshotEnvironmentBinding{}
+				binding, err = a.loader.FindExistingSnapshotEnvironmentBinding(a.client, a.context, a.application, &environment)
+				if err != nil {
+					a.logger.Error(err, "Failed to find snapshotEnvironmentBinding associated with environment", "environment.Name", environment.Name)
+					return reconciler.RequeueWithError(err)
+				}
+				if binding == nil {
+					binding, err = a.createSnapshotEnvironmentBindingForSnapshot(a.application, &environment, a.snapshot, components)
+					if err != nil {
+						a.logger.Error(err, "Failed to create snapshotEnvironmentbinding for snapshot",
+							"snapshot", a.snapshot.Name,
+							"environment.Name", environment.Name,
+							"snapshot.Spec.Components", a.snapshot.Spec.Components)
+						return reconciler.RequeueWithError(err)
+					}
+					a.logger.LogAuditEvent("A snapshotEnvironmentbinding is created", binding, h.LogActionAdd,
+						"integrationTestScenario.Name", integrationTestScenario.Name)
+				} else {
+					a.logger.Info("SnapshotEnvironmentBinding already exists for environment",
+						"binding.Name", binding.Name,
+						"environment.Name", environment.Name)
+				}
+
+				//check if the scenario label is added to the binding, if not, add it
+				if !h.HasLabelWithValue(binding, gitops.SnapshotTestScenarioLabel, integrationTestScenario.Name) {
+					err = a.labelBindingWithScenario(binding, integrationTestScenario.Name)
+					if err != nil {
+						a.logger.Error(err, "Failed to update snapshotEnvironmentbinding label",
+							"binding.Name", binding.Name,
+							gitops.SnapshotTestScenarioLabel, integrationTestScenario.Name)
+						return reconciler.RequeueWithError(err)
+					}
+					a.logger.LogAuditEvent("Updated snapshotEnvironmentbinding label", binding, h.LogActionUpdate,
+						gitops.SnapshotTestScenarioLabel, integrationTestScenario.Name)
+				} else {
+					a.logger.Info("SnapshotEnvironmentBinding already contains label",
+						"binding.Name", binding.Name,
+						"integrationScenario.Name", integrationTestScenario.Name)
+				}
 				continue TestScenarioLoop
 			}
 		}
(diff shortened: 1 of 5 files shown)


controllers/snapshot/snapshot_adapter.go
```diff
@@ -222,10 +222,20 @@ TestScenarioLoop:
 			a.logger.Error(err, "Failed to create snapshotEnvironmentbinding for snapshot.",
 				"binding", binding,
 				"snapshot.Spec.Components", a.snapshot.Spec.Components)
+			return reconciler.RequeueOnErrorOrStop(err)
 		}
 		a.logger.LogAuditEvent("A snapshotEnvironmentbinding is created", binding, h.LogActionAdd,
 			"environment.Name", copyEnv.Name,
 			"integrationTestScenario.Name", integrationTestScenario.Name)
+		err = a.labelBindingWithScenario(binding, gitops.SnapshotTestScenarioLabel, integrationTestScenario.Name)
```
dheerajodha, line 230: Since this is not a generic function, in order to keep the function signature concise, what do you think about removing `gitops.SnapshotTestScenarioLabel` from here and using it directly in the function body?

controllers/snapshot/snapshot_adapter_test.go
```diff
@@ -640,17 +639,42 @@ var _ = Describe("Snapshot Adapter", Ordered, func() {
...
+			env := &applicationapiv1alpha1.Environment{}
+			err := adapter.client.Get(adapter.context, types.NamespacedName{
+				Name:      binding.Spec.Environment,
+				Namespace: binding.Namespace,
+			}, env)
+			Expect(err).To(BeNil())
+
+			dtc := &applicationapiv1alpha1.DeploymentTargetClaim{}
+			err = adapter.client.Get(adapter.context, types.NamespacedName{
+				Name:      env.Spec.Configuration.Target.DeploymentTargetClaim.ClaimName,
```
dheerajodha, line 674: I think this might cause panic when the `env` variable is nil.

controllers/snapshot/snapshot_adapter.go
hongweiliu17: Updated. Thanks.

controllers/snapshot/snapshot_adapter.go
```diff
@@ -222,10 +222,20 @@ TestScenarioLoop:
 			a.logger.Error(err, "Failed to create snapshotEnvironmentbinding for snapshot.",
 				"binding", binding,
 				"snapshot.Spec.Components", a.snapshot.Spec.Components)
+			return reconciler.RequeueOnErrorOrStop(err)
 		}
 		a.logger.LogAuditEvent("A snapshotEnvironmentbinding is created", binding, h.LogActionAdd,
 			"environment.Name", copyEnv.Name,
 			"integrationTestScenario.Name", integrationTestScenario.Name)
+		err = a.labelBindingWithScenario(binding, integrationTestScenario.Name)
```
dirgim, line 230: Hm, thinking some more about this - if we fail on adding the label to the binding here and requeue the reconcile loop for this, we will find the existing environment [here](https://github.com/redhat-appstudio/integration-service/pull/218/files#diff-8e718c6bf06f754db7e9e94c6f74b5c6704941dff95d637289e3c4cc194c95a0R187) and will not get to this line again.
Should we also check if the appropriate SnapshotEnvironmentBinding exists with the correct labels below [that line](https://github.com/redhat-appstudio/integration-service/blob/main/controllers/snapshot/snapshot_adapter.go#L187)?

hongweiliu17: Good point. Thanks!
So I will check if the binding exists and label exists after [this line](https://github.com/redhat-appstudio/integration-service/blob/main/controllers/snapshot/snapshot_adapter.go#L187)

hongweiliu17: Updated. Thanks!

hongweiliu17: I found the udpated code is not shown. Let me check again.

hongweiliu17: It can be reviewed now.

controllers/snapshot/snapshot_adapter.go
```diff
@@ -172,7 +172,12 @@ func (a *Adapter) EnsureCreationOfEnvironment() (reconciler.OperationResult, err
 	allEnvironments, err := a.loader.GetAllEnvironments(a.client, a.context, a.application)
 	if err != nil {
 		a.logger.Error(err, "Failed to get all environments.")
-		return reconciler.RequeueOnErrorOrStop(err)
+		return reconciler.RequeueWithError(err)
```
hongweiliu17, line 175: Should we call `RequeueWithError` or `RequeueOnErrorOrStop` when error is met for `a.loader.GetAllEnvironments` and the following `a.loader.GetAllApplicationComponents`

dirgim: I think that `RequeueWithError` makes more sense here as it can be an intermittent error when getting Environments or Components from the api server, so it makes sense to retry.

hongweiliu17: Thanks!

I searched release-service code, `RequeueOnErrorOrStop` is used in two places in https://github.com/redhat-appstudio/release-service/blob/main/controllers/release/adapter.go#L259, other places use `RequeueWithError`. 

And I also searched the [four adapter files](https://github.com/redhat-appstudio/integration-service/tree/main/controllers) in integration-service, and personally feel most of time, we should use `RequeueWithError`, not `RequeueOnErrorOrStop` to make the object processed again.

jinqi7: Yes. I agree. One thing we may need to consider is the delay time of requeue - when the request is requeued after the failure, we may need to set  the delay time to avoid affecting the new requests in the controller.

jinqi7 (approved):

controllers/snapshot/snapshot_adapter.go
dirgim: I agree on both points, I think we can file a follow-up story to review our `RequeueOnErrorOrStop` and `RequeueWithError` usage as well as introducing delay time.

hongweiliu17: Ok. Thank you Kruno and Jing, let me file a ticket.

controllers/snapshot/snapshot_adapter.go
dirgim: Thanks!

controllers/snapshot/snapshot_adapter.go
```diff
@@ -188,44 +193,92 @@ TestScenarioLoop:
...
+				"snapshot", a.snapshot.Name,
+				"environment.Name", copyEnv.Name,
 				"snapshot.Spec.Components", a.snapshot.Spec.Components)
+			return reconciler.RequeueWithError(err)
 		}
 		a.logger.LogAuditEvent("A snapshotEnvironmentbinding is created", binding, h.LogActionAdd,
 			"environment.Name", copyEnv.Name,
 			"integrationTestScenario.Name", integrationTestScenario.Name)
+
+		err = a.labelBindingWithScenario(binding, integrationTestScenario.Name)
```
dirgim, line 273: The code as it is now does work, and the possible errors are handled well.

I'm thinking this might be simplified further to just add the label as part of creating the SnapshotEnvironmentBinding. It would require modifying the `createSnapshotEnvironmentBindingForSnapshot` function to take the optional label name and value (or something similar to that) and add it before creating the resource on the cluster.
@hongweiliu17 wdyt about it?  I'm fine with keeping it this way too if you think it's not worth the hassle to implement it that way.

hongweiliu17: Thank you. I agree. I thought adding it to `createSnapshotEnvironmentBindingForSnapshot` is simpler, but  I found `createSnapshotEnvironmentBindingForSnapshot` is called [here](https://github.com/redhat-appstudio/integration-service/blob/main/controllers/snapshot/snapshot_adapter.go#L348) and it doesn't need this label. You are right, we may add optional label name and value params.

Let me update it.

controllers/snapshot/snapshot_adapter.go
hongweiliu17: [STONEINTG-483](https://issues.redhat.com/browse/STONEINTG-483) is filed.

controllers/snapshot/snapshot_adapter.go

controllers/snapshot/snapshot_adapter.go
```diff
@@ -466,7 +493,7 @@ func (a *Adapter) createIntegrationPipelineRun(application *applicationapiv1alph
 // If it's not possible to create it and set the application as the owner, an error will be returned
 func (a *Adapter) createSnapshotEnvironmentBindingForSnapshot(application *applicationapiv1alpha1.Application,
 	environment *applicationapiv1alpha1.Environment, snapshot *applicationapiv1alpha1.Snapshot,
-	components *[]applicationapiv1alpha1.Component) (*applicationapiv1alpha1.SnapshotEnvironmentBinding, error) {
+	components *[]applicationapiv1alpha1.Component, optionalLabelKeyAndName ...string) (*applicationapiv1alpha1.SnapshotEnvironmentBinding, error) {
```
dirgim, line 496: Wdyt about having the labels parameter as a `map[string]string{}` and putting the labels in that way?
That would remove the need for the [checks below](https://github.com/redhat-appstudio/integration-service/pull/218/files#diff-8e718c6bf06f754db7e9e94c6f74b5c6704941dff95d637289e3c4cc194c95a0R509-R510) and would simply add what's in that map to the labels.

hongweiliu17: Yes, let me update it.

dirgim (approved):
LGTM, thanks!

controllers/snapshot/snapshot_adapter.go
dirgim: Looks good!

dheerajodha (approved):
LGTM, thanks!

controllers/snapshot/snapshot_adapter.go
hongweiliu17: Thanks for the suggestions. Learned them.

PR #228, opened 2023-06-22
feat(STONEINTG-433): cleanup ephemeral envs for some errors

Josh-Everett:
When an issue happens with deployment of the Snapshot during Integration testing on ephemeral environments, the Integration service needs to delete the associated Environment and DTC CRs in order to clean up the provisioned namespace.

### DIFF
--- a/controllers/binding/snapshotenvironmentbinding_adapter.go
+++ b/controllers/binding/snapshotenvironmentbinding_adapter.go
@@ -18,6 +18,7 @@ import (
 	"fmt"
 
 	"github.com/redhat-appstudio/operator-goodies/reconciler"
+	"k8s.io/apimachinery/pkg/api/meta"
 	ctrl "sigs.k8s.io/controller-runtime"
 
 	applicationapiv1alpha1 "github.com/redhat-appstudio/application-api/api/v1alpha1"
@@ -69,6 +70,11 @@ func (a *Adapter) EnsureIntegrationTestPipelineForScenarioExists() (reconciler.O
 		return reconciler.ContinueProcessing()
 	}
 
+	if gitops.HaveBindingsFailed(a.snapshotEnvironmentBinding) {
+		a.logger.Info("The SnapshotEnvrionmentBinding has failed to deploy on ephemeral envrionment and will be deleted later.", "snapshotEnvironmentBinding.Name", a.snapshotEnvironmentBinding.Name)
+		return reconciler.ContinueProcessing()
+	}
+
 	if a.integrationTestScenario != nil {
 		integrationPipelineRun, err := loader.GetLatestPipelineRunForSnapshotAndScenario(a.client, a.context, a.loader, a.snapshot, a.integrationTestScenario)
 		if err != nil {
@@ -97,6 +103,39 @@ func (a *Adapter) EnsureIntegrationTestPipelineForScenarioExists() (reconciler.O
 	}
 
 	return reconciler.ContinueProcessing()
+}
+
+// EnsureEphemeralEnvironmentsCleanedUp will ensure that ephemeral environment(s) associated with the
+// SnapshotEnvironmentBinding are cleaned up.
+func (a *Adapter) EnsureEphemeralEnvironmentsCleanedUp() (reconciler.OperationResult, error) {
+
+	if !gitops.HaveBindingsFailed(a.snapshotEnvironmentBinding) {
+		return reconciler.ContinueProcessing()
+	}
+
+	// mark snapshot as failed
+	snapshotErrorMessage := "Encountered issue deploying snapshot on ephemeral environments: " +
+		meta.FindStatusCondition(a.snapshotEnvironmentBinding.Status.ComponentDeploymentConditions, "ErrorOccurred").Message
+	_, err := gitops.MarkSnapshotAsFailed(a.client, a.context, a.snapshot, snapshotErrorMessage)
+	if err != nil {
+		a.logger.Error(err, "Failed to Update Snapshot status")
+		return reconciler.RequeueOnErrorOrStop(err)
+	}
+
+	deploymentTargetClaim, err := a.loader.GetDeploymentTargetClaimForEnvironment(a.client, a.context, a.environment)
+	if err != nil {
+		a.logger.Error(err, "failed to find deploymentTargetClaim defined in environment %s", a.environment.Name)
+		return reconciler.RequeueOnErrorOrStop(err)
+	}
+
+	err = h.CleanUpEphemeralEnvironments(a.client, &a.logger, a.context, a.environment, deploymentTargetClaim)
+	if err != nil {
+		a.logger.Error(err, "Failed to delete the Ephemeral Environment")
+		return reconciler.RequeueOnErrorOrStop(err)
+	}
+
+	return reconciler.ContinueProcessing()
+
 }
 
 // createIntegrationPipelineRunWithEnvironment creates new integration PipelineRun. The Pipeline information and the parameters to it
--- a/controllers/binding/snapshotenvironmentbinding_controller.go
+++ b/controllers/binding/snapshotenvironmentbinding_controller.go
@@ -30,6 +30,7 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 )
 
 // Reconciler reconciles a SnapshotEnvironmentBinding object
@@ -115,6 +116,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 	return reconciler.ReconcileHandler([]reconciler.ReconcileOperation{
 		adapter.EnsureIntegrationTestPipelineForScenarioExists,
+		adapter.EnsureEphemeralEnvironmentsCleanedUp,
 	})
 }
 
@@ -199,6 +201,9 @@ func SetupController(manager ctrl.Manager, log *logr.Logger) error {
 // setupControllerWithManager sets up the controller with the Manager which monitors new SnapshotEnvironmentBindings
 func setupControllerWithManager(manager ctrl.Manager, reconciler *Reconciler) error {
 	return ctrl.NewControllerManagedBy(manager).
-		For(&applicationapiv1alpha1.SnapshotEnvironmentBinding{}, builder.WithPredicates(predicates.GenerationUnchangedOnUpdatePredicate{}, gitops.DeploymentSucceededForIntegrationBindingPredicate())).
+		For(&applicationapiv1alpha1.SnapshotEnvironmentBinding{},
+			builder.WithPredicates(predicates.GenerationUnchangedOnUpdatePredicate{})).
+		WithEventFilter(predicate.Or(
+			gitops.DeploymentSucceededForIntegrationBindingPredicate(), gitops.DeploymentFailedForIntegrationBindingPredicate())).
 		Complete(reconciler)
 }
--- a/gitops/binding.go
+++ b/gitops/binding.go
@@ -27,6 +27,10 @@ const (
 	// BindingDeploymentStatusConditionType is the condition type to retrieve from the ComponentDeploymentConditions
 	// in the SnapshotEnvironmentBinding's status to copy into the Release status
 	BindingDeploymentStatusConditionType string = "AllComponentsDeployed"
+
+	// BindingErrorOccurredStatusConditionType is the condition to check for failures within the
+	// SnapshotEnvironmentBindingConditions status
+	BindingErrorOccurredStatusConditionType string = "ErrorOccurred"
 )
 
 // NewSnapshotEnvironmentBinding creates a new SnapshotEnvironmentBinding using the provided info.
@@ -61,6 +65,14 @@ func NewBindingComponents(components []applicationapiv1alpha1.Component) *[]appl
 	return &bindingComponents
 }
 
+func HaveBindingsFailed(snapshotEnvironmentBinding *applicationapiv1alpha1.SnapshotEnvironmentBinding) bool {
+	bindingStatus := meta.FindStatusCondition(snapshotEnvironmentBinding.Status.ComponentDeploymentConditions, BindingErrorOccurredStatusConditionType)
+	if bindingStatus == nil {
+		return false
+	}
+	return bindingStatus.Status == metav1.ConditionTrue
+}
+
 // hasDeploymentSucceeded returns a boolean that is only true if the first passed object
 // is a SnapshotEnvironmentBinding with the componentDeployment status anything other than True and
 // the second passed object is a SnapshotEnvironmentBinding with the componentDeployment status True.
@@ -77,5 +89,22 @@ func hasDeploymentSucceeded(objectOld, objectNew client.Object) bool {
 		}
 	}
 
+	return (oldCondition == nil || oldCondition.Status != metav1.ConditionTrue) && newCondition.Status == metav1.ConditionTrue
+}
+
+// hasDeploymentFailed returns a boolean that is only true if the first passed object
+// is a SnapshotEnvironmentBinding with the BindingConditions status anything other than True and
+// the second passed object is a SnapshotEnvironmentBinding with the BindingErrorOccurred status True.
+func hasDeploymentFailed(objectOld, objectNew client.Object) bool {
+	var oldCondition, newCondition *metav1.Condition
+	if oldBinding, ok := objectOld.(*applicationapiv1alpha1.SnapshotEnvironmentBinding); ok {
+		oldCondition = meta.FindStatusCondition(oldBinding.Status.BindingConditions, BindingErrorOccurredStatusConditionType)
+	}
+	if newBinding, ok := objectNew.(*applicationapiv1alpha1.SnapshotEnvironmentBinding); ok {
+		newCondition = meta.FindStatusCondition(newBinding.Status.BindingConditions, BindingErrorOccurredStatusConditionType)
+		if newCondition == nil {
+			return false
+		}
+	}
 	return (oldCondition == nil || oldCondition.Status != metav1.ConditionTrue) && newCondition.Status == metav1.ConditionTrue
 }
--- a/gitops/predicates_test.go
+++ b/gitops/predicates_test.go
@@ -15,6 +15,7 @@ package gitops_test
 
 import (
 	"context"
+
 	"github.com/redhat-appstudio/integration-service/gitops"
 
 	. "github.com/onsi/ginkgo/v2"
@@ -36,7 +37,7 @@ var _ = Describe("Predicates", Ordered, func() {
 		snapshotName    = "test-snapshot"
 	)
 
-	var bindingMissingStatus, bindingFalseStatus, bindingTrueStatus *applicationapiv1alpha1.SnapshotEnvironmentBinding
+	var bindingMissingStatus, bindingFalseStatus, bindingTrueStatus, bindingDeploymentFailedStatus *applicationapiv1alpha1.SnapshotEnvironmentBinding
 
 	BeforeAll(func() {
 		bindingMissingStatus = &applicationapiv1alpha1.SnapshotEnvironmentBinding{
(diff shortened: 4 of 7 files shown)


controllers/binding/snapshotenvironmentbinding_controller.go
```diff
@@ -102,6 +110,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 	adapter := NewAdapter(snapshotEnvironmentBinding, snapshot, environment, application, integrationTestScenario, logger, loader, r.Client, ctx)
 
+	bindingConditionsFailed := meta.FindStatusCondition(snapshotEnvironmentBinding.Status.BindingConditions, AppStudioSnapshotEnvironmentBindingConditionsFailed)
+	if bindingConditionsFailed != nil {
+		result, err := adapter.EnsureEphemeralEnvironmentsCleanedUp()
```
dirgim, line 115: The `EnsureEphemeralEnvironmentsCleanedUp` is part of the pipeline adapter and is specific to the reconciliation of Integration `pipelineRuns`. We would need to create a new function that handles this within the [snapshotenvironmentbinding adapter]([snapshotenvironmentbinding_adapter.go](https://github.com/redhat-appstudio/integration-service/blob/main/controllers/binding/snapshotenvironmentbinding_adapter.go)).

dirgim: Duplication of code is a good point. We could create a new function in [gitops/environment.go](https://github.com/redhat-appstudio/integration-service/blob/main/gitops/environment.go) which handles deletion of an environment in a generic manner, and that function could then be used by both adapter functions when needing to clean up an environment.

controllers/binding/snapshotenvironmentbinding_controller.go
```diff
@@ -32,6 +34,12 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
+	// AppStudioSnapshotEnvironmentBindingConditionsFailed is the condition to check for failures within the
+	// Snapshot Environment BindingConditions status
+	AppStudioSnapshotEnvironmentBindingConditionsFailed string = "False"
```
dirgim, line 40: Apologies, the example that I shared in the story had the value set as `False` when it should have been `True`

controllers/binding/snapshotenvironmentbinding_controller.go
```diff
@@ -102,6 +110,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 	adapter := NewAdapter(snapshotEnvironmentBinding, snapshot, environment, application, integrationTestScenario, logger, loader, r.Client, ctx)
 
+	bindingConditionsFailed := meta.FindStatusCondition(snapshotEnvironmentBinding.Status.BindingConditions, AppStudioSnapshotEnvironmentBindingConditionsFailed)
```
dirgim, line 113: I think this condition would need to look more like this:
```
const (
BindingErrorOccurredStatusConditionType string = "ErrorOccurred"
)
(...)
bindingConditionsErrorOccured := meta.FindStatusCondition(snapshotEnvironmentBinding.Status.BindingConditions, BindingErrorOccurredStatusConditionType)

if bindingConditionsErrorOccured.Status == metav1.ConditionTrue {
(...)
}
```

controllers/binding/snapshotenvironmentbinding_controller.go
Josh-Everett: ack, updated in most recent patch.

controllers/binding/snapshotenvironmentbinding_adapter.go
```diff
@@ -141,3 +141,72 @@ func (a *Adapter) getDeploymentTargetForEnvironment(environment *applicationapiv
...
+		binding, err := a.loader.FindExistingSnapshotEnvironmentBinding(a.client, a.context, a.application, testEnvironment)
+		if err != nil || binding == nil {
+			a.logger.Error(err, "Failed to find snapshotEnvironmentBinding associated with environment", "environment.Name", testEnvironment.Name)
+			return reconciler.RequeueWithError(err)
+		}
+
+		a.logger.Info("Deleting deploymentTarget", "deploymentTarget.Name", dt.Name)
+		err = a.client.Delete(a.context, dt)
+		if err != nil {
+			a.logger.Error(err, "Failed to delete the deploymentTarget")
```
jsztuka, line 179: Suggestion:
We can include `"deploymentTarget.Name", dt.Name` in error msg also, so its easier to debug.
Same applies for rest of the error msgs.

controllers/binding/snapshotenvironmentbinding_controller_test.go
```diff
@@ -48,8 +48,8 @@ var _ = Describe("BindingController", func() {
 		hasSnapshot             *applicationapiv1alpha1.Snapshot
 		hasEnv                  *applicationapiv1alpha1.Environment
 		hasBinding              *applicationapiv1alpha1.SnapshotEnvironmentBinding
-		deploymentTargetClaim   *applicationapiv1alpha1.DeploymentTargetClaim
 		deploymentTarget        *applicationapiv1alpha1.DeploymentTarget
+		deploymentTargetClaim   *applicationapiv1alpha1.DeploymentTargetClaim
```
jsztuka, line 52: I suppose this can be set to its original state.

controllers/binding/snapshotenvironmentbinding_adapter.go
```diff
@@ -141,3 +141,72 @@ func (a *Adapter) getDeploymentTargetForEnvironment(environment *applicationapiv
...
+
+		a.logger.Info("Deleting environment", "environment.Name", testEnvironment.Name)
+		err = a.client.Delete(a.context, testEnvironment)
+		if err != nil {
+			a.logger.Error(err, "Failed to delete the test ephemeral environment")
+			return reconciler.RequeueWithError(err)
+		}
+		a.logger.LogAuditEvent("Ephemeral environment deleted", testEnvironment, h.LogActionDelete)
+
+		a.logger.Info("Deleting snapshotEnvironmentBinding", "binding.Name", binding.Name)
```
dirgim, line 200: Note that deleting snapshotEnvironmentBinding should not be done after #236 is merged as the owner reference from the environment will ensure that the SEB is also deleted. It could be a good idea to create a function that handles cleanup of the environment as we talked about earlier so we don't duplicate the logic and miss things like this in the future.

controllers/binding/snapshotenvironmentbinding_controller.go
```diff
@@ -102,6 +110,22 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 	adapter := NewAdapter(snapshotEnvironmentBinding, snapshot, environment, application, integrationTestScenario, logger, loader, r.Client, ctx)
 
+	// check for failed bindingconditions and attempt to clean up Ephemeral environments
+	// see STONEINTG-433
+	bindingConditionsFailed := meta.FindStatusCondition(snapshotEnvironmentBinding.Status.BindingConditions, BindingErrorOccurredStatusConditionType)
+	if bindingConditionsFailed.Status == metav1.ConditionTrue {
```
dirgim, line 116: I think that we shouldn't have the business logic directly within the controller - both for testing and logic separation reasons.
I propose moving this condition to the beginning of the `EnsureEphemeralEnvironmentsCleanedUp` within the adapter and handle it similarly to how it's done within the other similar functions, e.g. [here](https://github.com/redhat-appstudio/integration-service/blob/c3828f452f3d80f65806265ee80cf291f504272b/controllers/binding/snapshotenvironmentbinding_adapter.go#L65-L68).

controllers/binding/snapshotenvironmentbinding_controller_test.go
```diff
@@ -339,4 +339,15 @@ var _ = Describe("BindingController", func() {
 		}()
 	})
 
+	It("cleans up ephemeral environment if provisioning fails", func() {
+		hasEnv.Spec.Tags[0] = "ephemeral"
```
dirgim, line 343: This kind of test should probably be moved to the adapter unit tests (see my [other comment](https://github.com/redhat-appstudio/integration-service/pull/228/files#r1252182301)).

Also, since the hasEnv.Spec.Tags is an empty list at this point, setting the value of (nonexistent) first index causes this error:
```
runtime error: invalid memory address or nil pointer dereference
```
Using something like `hasEnv.Spec.Tags = append(hasEnv.Spec.Tags, "ephemeral")` should work.

controllers/binding/snapshotenvironmentbinding_controller.go
```diff
@@ -107,6 +113,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 	adapter := NewAdapter(snapshotEnvironmentBinding, snapshot, environment, application, integrationTestScenario, logger, loader, r.Client, ctx)
 
+	// check for failed bindingconditions and attempt to clean up Ephemeral environments
+	// see STONEINTG-433
+	_, err = adapter.checkForBindingConditionsFailure()
```
dirgim, line 118: I think this would be better served by being an `Ensure*` function and called as part of the ReconcileHandler [here](https://github.com/redhat-appstudio/integration-service/pull/228/files#diff-5b178fa36b2385712f39d2575244cfca705af946a979640c73ed86f02e3ff0e0L111).

controllers/binding/snapshotenvironmentbinding_adapter.go
```diff
@@ -141,3 +144,82 @@ func (a *Adapter) getDeploymentTargetForEnvironment(environment *applicationapiv
 
 	return deploymentTarget, nil
 }
+
+// checkForBindingConditionsFailure will ensure that ephemeral environment(s) are cleaned up.
+func (a *Adapter) checkForBindingConditionsFailure() (reconciler.OperationResult, error) {
+	testEnvironment := a.environment
+
+	// get SnapshotEnvironmentBinding
+	binding, err := a.loader.FindExistingSnapshotEnvironmentBinding(a.client, a.context, a.application, testEnvironment)
```
dirgim, line 153: I don't think getting the binding here is necessary - this controller reacts to SnapshotEnvironmentBinding changes, so the `a.snapshotEnvironmentBinding`  will already contain the correct SnapshotEnvironmentBinding.

Josh-Everett:
This PR needs some additional test coverage for the predicates, would appreciate input in the meantime.

controllers/binding/snapshotenvironmentbinding_controller.go
```diff
@@ -199,6 +199,6 @@ func SetupController(manager ctrl.Manager, log *logr.Logger) error {
 // setupControllerWithManager sets up the controller with the Manager which monitors new SnapshotEnvironmentBindings
 func setupControllerWithManager(manager ctrl.Manager, reconciler *Reconciler) error {
 	return ctrl.NewControllerManagedBy(manager).
-		For(&applicationapiv1alpha1.SnapshotEnvironmentBinding{}, builder.WithPredicates(predicates.GenerationUnchangedOnUpdatePredicate{}, gitops.DeploymentSucceededForIntegrationBindingPredicate())).
+		For(&applicationapiv1alpha1.SnapshotEnvironmentBinding{}).WithEventFilter(predicate.Or(predicates.GenerationUnchangedOnUpdatePredicate{}, gitops.DeploymentSucceededForIntegrationBindingPredicate(), gitops.DeploymentFailedForIntegrationBindingPredicate())).
```
dirgim, line 202: I think that the `predicates.GenerationUnchangedOnUpdatePredicate{}` shouldn't be part of the `Or` clause here because that would then catch all update events to the resource and make the other two predicates redundant if I'm not mistaken. @hongweiliu17 can you also take a look at this?

hongweiliu17: Yes. I found this [Or](https://github.com/onsi/gomega/blob/master/matchers.go#L609-L617)

controllers/binding/snapshotenvironmentbinding_adapter.go
```diff
@@ -99,6 +104,30 @@ func (a *Adapter) EnsureIntegrationTestPipelineForScenarioExists() (reconciler.O
...
+
+	if !gitops.HaveBindingsFailed(a.snapshotEnvironmentBinding) {
+		return reconciler.ContinueProcessing()
+	}
+
+	deploymentTargetClaim, err := a.loader.GetDeploymentTargetClaimForEnvironment(a.client, a.context, a.environment)
+	if err != nil {
+		a.logger.Error(err, "failed to find deploymentTargetClaim defined in environment %s", a.environment.Name)
+		return reconciler.RequeueOnErrorOrStop(err)
+	}
```
dheerajodha, line 119: Totally optional (I'd still approve even without it): but would it be too weird if we move this method call to `controllers/binding/snapshotenvironmentbinding_controller.go` file, right after we are done extracting the environment there, so [somewhere here](https://github.com/redhat-appstudio/integration-service/pull/228/files#diff-5b178fa36b2385712f39d2575244cfca705af946a979640c73ed86f02e3ff0e0R107)?

controllers/binding/snapshotenvironmentbinding_controller.go
```diff
@@ -114,7 +114,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	adapter := NewAdapter(snapshotEnvironmentBinding, snapshot, environment, application, component, integrationTestScenario, logger, loader, r.Client, ctx)
 
 	return reconciler.ReconcileHandler([]reconciler.ReconcileOperation{
-		adapter.EnsureIntegrationTestPipelineForScenarioExists,
+		adapter.EnsureIntegrationTestPipelineForScenarioExists, adapter.EnsureEphemeralEnvironmentsCleanedUp,
```
dheerajodha, line 117: If you don't mind, can we please add this latest adapter method in a new line, so it stays consistent with how other adapters are defined within their controller?

dirgim:
This is looking good to me after updates, only the unit tests for predicates and their related functions to go now.

jinqi7 (commented):
Is the A/C item of " the controller marks the Snapshot as failed with the error reason" still  need to be included in the PR/task?

Josh-Everett:
> Is the A/C item of " the controller marks the Snapshot as failed with the error reason" still need to be included in the PR/task?

@jinqi7 yes thank you. It has been updated in [a9437ab](https://github.com/redhat-appstudio/integration-service/pull/228/commits/a9437aba9900e166568d360b64307fc06d3beed2). The update to the snapshot is [here](https://github.com/redhat-appstudio/integration-service/blob/a9437aba9900e166568d360b64307fc06d3beed2/controllers/binding/snapshotenvironmentbinding_adapter.go#L116) and the test [here](https://github.com/redhat-appstudio/integration-service/blob/a9437aba9900e166568d360b64307fc06d3beed2/controllers/binding/snapshotenvironmentbinding_adapter_test.go#L474)

controllers/binding/snapshotenvironmentbinding_adapter.go
```diff
@@ -99,6 +104,37 @@ func (a *Adapter) EnsureIntegrationTestPipelineForScenarioExists() (reconciler.O
 	return reconciler.ContinueProcessing()
 }
 
+// EnsureEphemeralEnvironmentsCleanedUp will ensure that ephemeral environment(s) associated with the
+// SnapshotEnvironmentBinding are cleaned up.
+func (a *Adapter) EnsureEphemeralEnvironmentsCleanedUp() (reconciler.OperationResult, error) {
```
14rcole, line 109: Why do we use a reconciliation loop for this rather than initiating cleanup in one of the functions that determines that the tests have completed?  I imagine that would use fewer resources since we're not running another function over and over again.

controllers/binding/snapshotenvironmentbinding_adapter.go
```diff
@@ -99,6 +104,37 @@ func (a *Adapter) EnsureIntegrationTestPipelineForScenarioExists() (reconciler.O
...
+
+	if !gitops.HaveBindingsFailed(a.snapshotEnvironmentBinding) {
+		return reconciler.ContinueProcessing()
+	}
+
+	// mark snapshot as failed
+	_, err := gitops.MarkSnapshotAsFailed(a.client, a.context, a.snapshot, "Encountered issue deploying snapshot on ephemeral environments")
```
dheerajodha, line 116: Instead of hardcoding the failure reason of the Snapshot, what do we think about copying over the content of the `message` field from the `BindingErrorOccurredStatusConditionType` of SEB?

gitops/binding.go
```diff
@@ -27,6 +27,10 @@ const (
 	// BindingDeploymentStatusConditionType is the condition type to retrieve from the ComponentDeploymentConditions
 	// in the SnapshotEnvironmentBinding's status to copy into the Release status
 	BindingDeploymentStatusConditionType string = "AllComponentsDeployed"
+
+	// AppStudioSnapshotEnvironmentBindingConditionsFailed is the condition to check for failures within the
```
dirgim, line 31: typo: `AppStudioSnapshotEnvironmentBindingConditionsFailed` -> `BindingErrorOccurredStatusConditionType`

controllers/binding/snapshotenvironmentbinding_adapter.go
```diff
@@ -69,6 +69,11 @@ func (a *Adapter) EnsureIntegrationTestPipelineForScenarioExists() (reconciler.O
 		return reconciler.ContinueProcessing()
 	}
 
+	if gitops.HaveBindingsFailed(a.snapshotEnvironmentBinding) {
+		a.logger.Info("The SnapshotEnvrionmentBinding has failed to deploy on ephemeral envrionment.")
+		return reconciler.ContinueProcessing()
+	}
```
dheerajodha, line 75: Do we need to mark the Snapshot as failed here as well?

gitops/predicates_test.go
```diff
@@ -128,4 +149,22 @@ var _ = Describe("Predicates", Ordered, func() {
...
+		instance := gitops.DeploymentFailedForIntegrationBindingPredicate()
+		It("returns true when the .Status.ComponentDeploymentConditions field of old SEB is set to false and that of new SEB is set to true", func() {
+			contextEvent := event.UpdateEvent{
+				ObjectOld: bindingFalseStatus,
+				ObjectNew: bindingDeploymentFailedStatus,
+			}
+			Expect(instance.Update(contextEvent)).To(BeTrue())
+		})
+
+		It("returns true when the .Status.ComponentDeploymentConditions field of old SEB is unset and that of new SEB is set to true", func() {
```
dirgim, line 162: typo: `ComponentDeploymentConditions` -> `BindingConditions.ErrorOccured`

controllers/binding/snapshotenvironmentbinding_adapter.go
dirgim: This loop is there for when the provisioning fails with an unforseen error (infra or similar) so in that scenario the tests would not have even started.

controllers/binding/snapshotenvironmentbinding_adapter.go
dirgim: I think we can skip marking it as failed here because this function is only supposed to handle successful snapshots. The newly introduced `EnsureEphemeralEnvironmentsCleanedUp()` should handle marking the snapshot as failed.

controllers/binding/snapshotenvironmentbinding_controller.go
```diff
@@ -199,6 +201,6 @@ func SetupController(manager ctrl.Manager, log *logr.Logger) error {
 // setupControllerWithManager sets up the controller with the Manager which monitors new SnapshotEnvironmentBindings
 func setupControllerWithManager(manager ctrl.Manager, reconciler *Reconciler) error {
 	return ctrl.NewControllerManagedBy(manager).
-		For(&applicationapiv1alpha1.SnapshotEnvironmentBinding{}, builder.WithPredicates(predicates.GenerationUnchangedOnUpdatePredicate{}, gitops.DeploymentSucceededForIntegrationBindingPredicate())).
+		For(&applicationapiv1alpha1.SnapshotEnvironmentBinding{}, builder.WithPredicates(predicates.GenerationUnchangedOnUpdatePredicate{})).WithEventFilter(predicate.Or(gitops.DeploymentSucceededForIntegrationBindingPredicate(), gitops.DeploymentFailedForIntegrationBindingPredicate())).
```
dheerajodha, line 204: This is a long line, hard to read without scrolling too much
if possible... :peeposhy:

hongweiliu17 (approved):
Only one log change suggestion. Others look good to me.

controllers/binding/snapshotenvironmentbinding_adapter.go
```diff
@@ -69,6 +70,11 @@ func (a *Adapter) EnsureIntegrationTestPipelineForScenarioExists() (reconciler.O
 		return reconciler.ContinueProcessing()
 	}
 
+	if gitops.HaveBindingsFailed(a.snapshotEnvironmentBinding) {
+		a.logger.Info("The SnapshotEnvrionmentBinding has failed to deploy on ephemeral envrionment  and will be deleted later.", "snapshotEnvironmentBinding.Name", a.snapshotEnvironmentBinding.Name)
```
dirgim, line 74: nit: typo: `envrionment  and` -> `environment and`

dheerajodha (approved):
/lgtm

dirgim (approved):
LGTM

PR #234, opened 2023-06-29
feat(STONEINTG-450): generate RBAC using kubebuilder directives

jsztuka:
Dropping RBAC roles and including them via directives in the controllers.


Josh-Everett (approved):

dirgim (approved):
LGTM

PR #237, opened 2023-07-04
fix(STONEINTG-498 & STONEINTG-504) Modify predicate logic of binding controller & also add labels to Test PipelineRuns

dheerajodha:


dirgim:
Looks good overall - I just have an unit testing suggestion:
Because the GitOps service doesn't set the `Unknown` status for the `AllComponentsDeployed` condition anymore, it would probably be a good idea to change the [bindingUnkownStatus](https://github.com/redhat-appstudio/integration-service/blob/main/gitops/predicates_test.go#L75-L80) to not have the `AllComponentsDeployed`  status condition at all.

dheerajodha:
> Looks good overall - I just have an unit testing suggestion:
Because the GitOps service doesn't set the Unknown status for the AllComponentsDeployed condition anymore, it would probably be a good idea to change the [bindingUnkownStatus](https://github.com/redhat-appstudio/integration-service/blob/main/gitops/predicates_test.go#L75-L80) to not have the AllComponentsDeployed status condition at all.

I've now removed the checks related to the `Unknown` status in the unit test.

gitops/predicates_test.go
```diff
@@ -68,14 +68,14 @@ var _ = Describe("Predicates", Ordered, func() {
...
 		ctx := context.Background()
 
-		Expect(k8sClient.Create(ctx, bindingUnknownStatus)).Should(Succeed())
+		Expect(k8sClient.Create(ctx, bindingFalseStatus)).Should(Succeed())
 		Expect(k8sClient.Create(ctx, bindingTrueStatus)).Should(Succeed())
 
 		// Set the binding statuses after they are created
-		bindingUnknownStatus.Status.ComponentDeploymentConditions = []metav1.Condition{
+		bindingFalseStatus.Status.ComponentDeploymentConditions = []metav1.Condition{
 			{
```
jinqi7, line 76: I think the condition can be empty after it's created. Then the BindingDeploymentStatusConditionType condition is set after the snapshot is deployed.

dheerajodha: I agree with you. But if we keep it empty, we won't be able to test the case when a SEB's BindingDeploymentStatusConditionType condition changes from False to True.

Josh-Everett (approved):
lgtm

gitops/predicates_test.go
jinqi7: From Kruno's comment, if I understand correctly, the real situation should be from an empty status to  AllComponentsDeployed status condition being set.

dirgim: I think it would be good to test both cases - when the `BindingDeploymentStatusConditionType` changes from False to True as well as when the condition doesn't exist and changes to `True`.

dheerajodha: Thanks all! I've updated the code to allow both scenarios when the status changes from "False to True" and from "Nothing to True".

gitops/predicates_test.go
```diff
@@ -87,18 +101,26 @@ var _ = Describe("Predicates", Ordered, func() {
 	})
 
 	AfterAll(func() {
```
hongweiliu17, line 103: `bindingMissingStatus` should be deleted together here. Others look good to me. Thanks!

hongweiliu17 (approved):
LGTM

MartinBasti (approved):
:heart:

dirgim:
@dheerajodha the e2e tests are failing on ([log](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/redhat-appstudio_integration-service/237/pull-ci-redhat-appstudio-integration-service-main-integration-service-e2e/1678812425333248000/artifacts/integration-service-e2e/gather-extra/artifacts/pods/integration-service_integration-service-integration-service-controller-managerg6p78_manager.log)):
```
Failed to create SnapshotEnvironmentBinding for snapshot","snapshot":"integration-e2e-tenant/snapshot-sample-ynun","application":"integration-e2e-tenant/integ-app-hoej","environment":"development","snapshot":"snapshot-sample-ynun","error":"snapshotenvironmentbindings.appstudio.redhat.com \"integ-app-hoej-development-binding-ds4rc\" is forbidden: cannot set blockOwnerDeletion if an ownerReference refers to a resource you can't set finalizers on: , <nil>","
```
Rebasing on the latest main should solve that.

dheerajodha:
> Rebasing on the latest main should solve that.

Thanks @dirgim!

Just for the record, I've tested my changes on a cluster.

[Regarding STONEINTG-498]: Here are the logs from the integration service's pod, which show that the Test PLR was created when IntegrationTestScenario contained an environment within it. You can also notice that Snapshot was tested in the ephemeral namespace successfully.
image

[Regarding STONEINTG-504]: The Test PLR now contains Application and Component name in its labels: 
image

dirgim (approved):
LGTM, nice work!

gitops/predicates_test.go
dirgim: Resolving this discussion to unblock the PR as the points have been addressed. We can follow up on slack if anything needs to be cleared up.

PR #240, opened 2023-07-12
chore(STONEINTG-476): SnapshotEnvironmentBinding Controller diagram

14rcole:
Add a diagram to documentation that maps the flow of the SnapshotEnvironmentBinding controller

### DIFF
--- a/docs/binding-controller.md
+++ b/docs/binding-controller.md
@@ -1,0 +1,34 @@
+```mermaid
+%%{init: {'theme':'forest'}}%%
+flowchart TD
+  %% Defining the styles
+    classDef Red fill:#FF9999;
+    classDef Amber fill:#FFDEAD;
+    classDef Green fill:#BDFFA4;
+
+  predicate((PREDICATE:  <br>SnapshotEnvironmentBinding<br>is updated or successfully<br>deployed))
+
+%%%%%%%%%%%%%%%%%%%%%%% Drawing EnsureIntegrationTestPipelineForScenarioExists() function
+
+%% Node definitions
+ensure1(Proceed further if:<br>Snapshot testing <br> not finished yet)
+isThereAnITS{"Is there an<br>IntegrationTestScenario<br>present in the binding?"}
+continueProcessing1[/Controller continues processing.../]
+getLatestPipelineRun("Get latest pipelineRun for<br>snapshot and scenario")
+isPipelineRunExisting{"Does an integration<br>pipelineRun exist?"}
+createNewPipelineRun("Create a new pipelineRun<br>for the snapshot")
+
+
+%% Node connections
+predicate               ----> |"EnsureIntegrationTestPipelineForScenarioExists()"|ensure1
+ensure1                 ---->       isThereAnITS
+isThereAnITS            --No-->     continueProcessing1
+isThereAnITS            --Yes-->    getLatestPipelineRun
+getLatestPipelineRun    ---->       isPipelineRunExisting
+isPipelineRunExisting   --Yes-->    continueProcessing1
+isPipelineRunExisting   --No-->     createNewPipelineRun
+createNewPipelineRun    ---->       continueProcessing1
+
+%% Assigning styles to nodes
+class predicate Amber;
+```


14rcole:
This is the error occuring in the `Go Test on Pull Requests / Check sources (pull_request)` job:
```
[FAILED] Unexpected error:
      <errors.aggregate | len:1, cap:1>: 
      timeout waiting for process kube-apiserver to stop
      [
          <*errors.errorString | 0xc00080abe0>{
              s: "timeout waiting for process kube-apiserver to stop",
          },
      ]
  occurred
  In [AfterSuite] at: /__w/integration-service/integration-service/controllers/snapshot/snapshot_suite_test.go:110 @ 07/12/23 19:37:56.548
  ```
  
The error is occuring in the teardown.  Now that we can mock all of the loaders, do we still need the kube-apiserver to be running for the unit tests?

File: docs/binding-controller.md [hunk 1]
```diff
@@ -0,0 +1,60 @@
...
+%%%%%%%%%%%%%%%%%%%%%%% Drawing createIntegrationPipelineRunWithEnvironment() function
+
+%% Node definitions
+getTarget(Get deployment target for<br>environment)
+createPipelineRun(Create new pipelineRun for<br>application/snapshot/environment combination)
+continueProcessing2[/Controller continues processing.../]
+
+
+%% Node connections
+predicate         ---->   |"createIntegrationPipelineRunWithEnvironment()"|getTarget
```
dheerajodha, line 41: I find this link between this function and predicate a little confusing. It makes it seem like the controller calls this function directly, which is not the case. I think we should add this branch within the `EnsureIntegrationTestPipelineForScenarioExists()` branch where it gets. Maybe something like this:


```mermaid
graph TD;
isPipelineRunExisting("Does an integration<br>pipeline run exist?")
getDeploymentTargetClaim(Load deployment target claim<br>from cluster using<br>environment namespace)
getDeploymentTarget(Load deployment target<br>from cluster using<br>deployment target claim)
createPipelineRun(Create new pipelineRun for<br>application/snapshot/environment combination)
continueProcessing3[/Controller continues processing.../]

isPipelineRunExisting   --No-->    getDeploymentTargetClaim
getDeploymentTargetClaim -->   getDeploymentTarget
getDeploymentTarget -->  createPipelineRun
createPipelineRun --> continueProcessing3
```

I basically shifted the content of the first 2 branches here ^
wdyt?

docs/binding-controller.md
```diff
@@ -0,0 +1,60 @@
...
+getLatestPipelineRun    ---->       isPipelineRunExisting
+isPipelineRunExisting   --No-->     continueProcessing1
+isPipelineRunExisting   --Yes-->    createNewPipelineRun
+createNewPipelineRun    ---->       continueProcessing1
+
+%%%%%%%%%%%%%%%%%%%%%%% Drawing createIntegrationPipelineRunWithEnvironment() function
+
+%% Node definitions
+getTarget(Get deployment target for<br>environment)
+createPipelineRun(Create new pipelineRun for<br>application/snapshot/environment combination)
```
dheerajodha, line 36: What do you think about explicitly mentioning that we do add a special parameter ("NAMESPACE") and workspace binding ("cluster-credentials") which takes the value from DT? That's what makes [this Test PLR](https://github.com/redhat-appstudio/integration-service/blob/main/controllers/binding/snapshotenvironmentbinding_adapter.go#L115) special than the one [created by the snapshot controller](https://github.com/redhat-appstudio/integration-service/blob/7b264cf65de4cc2b5bce4633a6ef415a12fcde36/controllers/snapshot/snapshot_adapter.go#L469). Wdyt?

14rcole: I think that gets too much into implementation details.  In fact, I think I should remove or modify the `application/snapshot/environment` bit.

docs/binding-controller.md
```diff
@@ -0,0 +1,60 @@
+```mermaid
+%%{init: {'theme':'forest'}}%%
+flowchart TD
+  %% Defining the styles
+    classDef Red fill:#FF9999;
+    classDef Amber fill:#FFDEAD;
+    classDef Green fill:#BDFFA4;
+
+  predicate((PREDICATE: <br>SnapshotEnvironmentBinding<br>is created or updated))
```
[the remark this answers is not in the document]
14rcole: I think following the [scenario controller diagram](https://github.com/redhat-appstudio/integration-service/blob/main/docs/scenario-controller.md?plain=1#L12) logic would be good here.  The predicate in that chart uses a statement instead of a question

docs/binding-controller.md
14rcole: I'm going to circle back to this once we figure out how to phrase the predicate.  That may impact what we do with the items that follow.

docs/binding-controller.md
dheerajodha: makes sense :+1:

docs/binding-controller.md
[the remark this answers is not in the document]
dheerajodha: Ah, sorry for the misunderstanding. The question mark wasn't part of the suggestion, I was just curious about what you'd think about it. Removed it now.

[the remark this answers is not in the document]
14rcole: Got it, I'm fine with that change,  I'll make it now

docs/binding-controller.md
14rcole: I think that makes sense.  We should probably do something similar with `getDeploymentTargetForEnvironment()`.  That being said, since this flowchart details how the controller itself works there's an argument to be made that only the `Ensure...()` functions should be diagrammed.  That's also in line with the way the already-documented controller flowcharts look.

dirgim: That is a good point - I'm not sure if individual functions need to be branched out here because e.g. `getDeploymentTargetForEnvironment()` is a prerequisite fact-gathering function that sets the stage for further `Ensure...` functions.

docs/binding-controller.md
dirgim: It could be good to mention `snapshot/ITS/environment` combination that uniquely identifies the pipelineRun.

docs/binding-controller.md
14rcole: Sounds like we have consensus.  I'll remove those functions

Josh-Everett (approved):
LGTM, will need some updating to reflect changes that result from [228](https://github.com/redhat-appstudio/integration-service/pull/228).

dheerajodha (commented):
The rest of the diagram looks good to me. I'll add my approval :+1: 

I just have one thing in mind. This diagram doesn't mention anywhere that it focuses only on those SEBs that are related to ephemeral environments. Was hoping to help a [general user like this](https://redhat-internal.slack.com/archives/C030SGP2VB9/p1685717398316689) to get "some" idea (not an exact idea, because our goal is not to pollute the diagram with impl details) about this controller. But it's just optional.

docs/binding-controller.md
```diff
@@ -0,0 +1,34 @@
...
+
+%% Node connections
+predicate               ----> |"EnsureIntegrationTestPipelineForScenarioExists()"|ensure1
+ensure1               ---->       isThereAnITS
+isThereAnITS            --No-->     continueProcessing1
+isThereAnITS            --Yes-->    getLatestPipelineRun
+getLatestPipelineRun    ---->       isPipelineRunExisting
+isPipelineRunExisting   --Yes-->     continueProcessing1
+isPipelineRunExisting   --No-->    createNewPipelineRun
+createNewPipelineRun    ---->       continueProcessing1
```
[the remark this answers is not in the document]
14rcole: Yeah, my tab sizes are different in vscode which is why they're showing up wrong here

dheerajodha (approved):
Thanks for the PR!
/lgtm

PR #243, opened 2023-07-17
fix(stoneintg-505): Use snapshot components when creating SEB

14rcole:
The integration service added a list of components to the SnapshotEnvironmentBinding upon creation.  Prior to this change the list was derived from the Application associated with the snapshot. This changes makes the integration service derive the components from the snapshot itself.

### DIFF
--- a/controllers/snapshot/snapshot_adapter.go
+++ b/controllers/snapshot/snapshot_adapter.go
@@ -344,7 +344,7 @@ func (a *Adapter) EnsureSnapshotEnvironmentBindingExist() (reconciler.OperationR
 		return reconciler.RequeueWithError(err)
 	}
 
-	components, err := a.loader.GetAllApplicationComponents(a.client, a.context, a.application)
+	components, err := a.loader.GetAllSnapshotComponents(a.client, a.context, a.snapshot)
 	if err != nil {
 		return reconciler.RequeueWithError(err)
 	}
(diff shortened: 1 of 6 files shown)


loader/loader_test.go
```diff
@@ -410,6 +410,49 @@ var _ = Describe("Loader", Ordered, func() {
...
+							URL: SampleRepoLink,
+						},
+					},
+				},
+			},
+			Status: applicationapiv1alpha1.ComponentStatus{
+				LastBuiltCommit: "",
+			},
+		}
+		Expect(k8sClient.Create(ctx, otherComp)).Should(Succeed())
```
dirgim, line 445: I think it would be good to delete this component at the end of the test to ensure a clean environment just in case.

dirgim (approved):
LGTM

hongweiliu17 (approved):
lgtm

controllers/snapshot/snapshot_adapter.go
```diff
@@ -344,7 +344,7 @@ func (a *Adapter) EnsureSnapshotEnvironmentBindingExist() (reconciler.OperationR
 		return reconciler.RequeueWithError(err)
 	}
 
-	components, err := a.loader.GetAllApplicationComponents(a.client, a.context, a.application)
+	components, err := a.loader.GetAllSnapshotComponents(a.client, a.context, a.snapshot)
```
hongweiliu17, line 347: You need to replace [ApplicationComponentsContextKey](https://github.com/redhat-appstudio/integration-service/blob/main/controllers/snapshot/snapshot_adapter_test.go#L527) with the newly added  `SnapshotComponentsContextKey`.

hongweiliu17 (approved):
lgtm

PR #244, opened 2023-07-18
fix(STONEINTG-433): refactor EnsureEphemeralEnvironmentsCleanedUp

Josh-Everett:
Logic for the EnsureEphemeralEnvironmentsCleanedUp function within the pipeline_adapter has been moved to a helper function to allow for use by the SnapshotEnvrionmentBinding.
See [this PR](https://github.com/redhat-appstudio/integration-service/pull/228) or [STONEINTG-433](https://issues.redhat.com/browse/STONEINTG-433) for motivation.

### DIFF
--- a/controllers/pipeline/pipeline_adapter.go
+++ b/controllers/pipeline/pipeline_adapter.go
@@ -265,24 +265,12 @@ func (a *Adapter) EnsureEphemeralEnvironmentsCleanedUp() (reconciler.OperationRe
 		return reconciler.ContinueProcessing()
 	}
 
-	isEphemeral := false
-	for _, tag := range testEnvironment.Spec.Tags {
-		if tag == "ephemeral" {
-			isEphemeral = true
-			break
-		}
-	}
+	isEphemeral := h.IsEnvironmentEphemeral(testEnvironment)
 
 	if isEphemeral {
 		dtc, err := a.loader.GetDeploymentTargetClaimForEnvironment(a.client, a.context, testEnvironment)
 		if err != nil || dtc == nil {
 			a.logger.Error(err, "Failed to find deploymentTargetClaim defined in environment", "environment.Name", testEnvironment.Name)
-			return reconciler.RequeueWithError(err)
-		}
-
-		dt, err := a.loader.GetDeploymentTargetForDeploymentTargetClaim(a.client, a.context, dtc)
-		if err != nil || dt == nil {
-			a.logger.Error(err, "Failed to find deploymentTarget defined in deploymentTargetClaim", "deploymentTargetClaim.Name", dtc.Name)
 			return reconciler.RequeueWithError(err)
 		}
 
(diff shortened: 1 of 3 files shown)


helpers/integration.go
```diff
@@ -289,3 +291,43 @@ func HasPipelineRunFinished(object client.Object) bool {
...
+		}
+	}
+	return isEphemeral
+}
+
+func CleanUpEphemeralEnvironments(client client.Client, logger *IntegrationLogger, ctx context.Context, env *applicationapiv1alpha1.Environment,
+	dt *applicationapiv1alpha1.DeploymentTarget, dtc *applicationapiv1alpha1.DeploymentTargetClaim) error {
+
+	logger.Info("Deleting deploymentTarget", "deploymentTarget.Name", dt.Name)
+	err := client.Delete(ctx, dt)
```
dirgim, line 310: I just realized that we shouldn't need to delete the `deploymentTarget` based on [this document](https://hackmd.io/UNMAvSdxSeezyliDk3bbzg#Scenario-Deprovisioning-DTDTC), @Josh-Everett are you ok with removing it as part of this refactor?

Josh-Everett: So it sounds like a DTC will have reclaimPolicy field in its spec that will determine if the DT is reclaimed or deleted? When the DTC gets a deleted timestamp the DT reconciler will deal with the reclaiming/deletion? Sounds like the DT delete needs to be removed from this code and the DT controller will deal with the rest.

dirgim: Yes, exactly. It's also possible deleting DT could cause unpredictable race conditions, so best to not delete it explicitly.

14rcole (approved):
LGTM

controllers/pipeline/pipeline_adapter.go
```diff
@@ -280,43 +274,17 @@ func (a *Adapter) EnsureEphemeralEnvironmentsCleanedUp() (reconciler.OperationRe
...
-		err = a.client.Delete(a.context, dt)
-		if err != nil {
-			a.logger.Error(err, "Failed to delete the deploymentTarget")
-			return reconciler.RequeueWithError(err)
-		}
-		a.logger.LogAuditEvent("DeploymentTarget deleted", dt, h.LogActionDelete)
-
-		a.logger.Info("Deleting deploymentTargetClaim", "deploymentTargetClaim.Name", dtc.Name)
-		err = a.client.Delete(a.context, dtc)
+		err = h.CleanUpEphemeralEnvironments(a.client, &a.logger, a.context, testEnvironment, dtc)
```
hongweiliu17, line 283: Mayber [these code calling Get* functions](https://github.com/redhat-appstudio/integration-service/blob/main/controllers/pipeline/pipeline_adapter.go#L277-L293) can be moved to these  `CleanUpEphemeralEnvironments()` together, then every time they can be run with the deleting code together.

dirgim: I think that the issue with moving those functions is that `CleanUpEphemeralEnvironments()` will also be called as part of the snapshotEnvironmentBinding adapter (#228) , and there will be no need to run [findExistingSnapshotEnvironmentBinding()](https://github.com/redhat-appstudio/integration-service/blob/0046c7466e6dbd07b6f653c452a07c68c840a9e0/controllers/pipeline/pipeline_adapter.go#L289) there since it's already known. 

Additionally,  @Josh-Everett fyi - we can probably remove this call to [GetDeploymentTargetForDeploymentTargetClaim()](https://github.com/redhat-appstudio/integration-service/blob/0046c7466e6dbd07b6f653c452a07c68c840a9e0/controllers/pipeline/pipeline_adapter.go#L283) since we don't need to clean that up any more.

Josh-Everett: @dirgim I might be mistaken but I think the call to `GetDeploymentTargetForDeploymentTargetClaim()` you're referencing should be removed in commit [8e00861](https://github.com/redhat-appstudio/integration-service/commit/8e00861747d591cc96561e35a940cff21ff773bc#diff-b836834f01fdc96e8d8fdf14e0266e2652f2bba56785f2f8dfba25b9b995cf6dL283)

hongweiliu17 (approved):
LGTM

controllers/pipeline/pipeline_adapter.go
Josh-Everett: It IS however still called within the[ adapter_test](https://github.com/redhat-appstudio/integration-service/blob/8e00861747d591cc96561e35a940cff21ff773bc/controllers/pipeline/pipeline_adapter_test.go#L1404)

dirgim: You're right, this should be good to go now in that case.

dirgim (approved):
LGTM

jinqi7 (commented):
I didn't see the "marks the Snapshot as failed" in the PR?  I found there is one more PR for STONEINTG-433.

jinqi7 (approved):

dheerajodha (approved):

PR #245, opened 2023-07-18
chore(gitignore): ignore cover.html and vendor folder

hongweiliu17:
git ignore cover.html and vendor folder


14rcole:
Isn't it common practice to commit vendor folders in go?  I see it in [tekton](https://github.com/tektoncd/pipeline) for example as well as [kubernetes](https://github.com/kubernetes/kubernetes).

hongweiliu17:
> Isn't it common practice to commit vendor folders in go? I see it in [tekton](https://github.com/tektoncd/pipeline) for example as well as [kubernetes](https://github.com/kubernetes/kubernetes).

Thanks for the question. I think since our project and some of RHTAP project are not  having vendor folder currently, then we can ignore it when committing the files. If we decide to keep vendor folder in the future, we may remove this line.  @dirgim may know the decision from high level.

jencull (approved):
lgtm

dirgim:
> Thanks for the question. I think since our project and some of RHTAP project are not having vendor folder currently, then we can ignore it when committing the files. If we decide to keep vendor folder in the future, we may remove this line. @dirgim may know the decision from high level.

I think we can keep the behavior of not having vendor folder until something changes in the future on RHTAP level.

dirgim (approved):
+1

PR #246, opened 2023-07-18
feat(STONEINTG-483): add requeueDelay to ensure* functions

hongweiliu17:
Read the code about `RequeueRequest`, `CancelRequest` and `error` in [ReconcileHandler(operations []ReconcileOperation)](https://github.com/redhat-appstudio/operator-goodies/blob/main/reconciler/handler.go#L23-L35), according to the definitions of [RequeueWithError(error)](https://github.com/redhat-appstudio/operator-goodies/blob/main/reconciler/results.go#L45) and [RequeueOnErrorOrStop(error)](https://github.com/redhat-appstudio/operator-goodies/blob/main/reconciler/results.go#L75), these two functions are the same when the `error` is not `nil`, that is to `requeue`.

1. Replace `RequeueOnErrorOrStop(error)` with `RequeueWithError(error)` when error is not `nil` to be consistent in all  adapters
2. Call `RequeueWithError(err)` when failing to update scenario to requeue

### DIFF
--- a/controllers/pipeline/pipeline_adapter.go
+++ b/controllers/pipeline/pipeline_adapter.go
@@ -126,6 +126,7 @@ func (a *Adapter) EnsureSnapshotExists() (reconciler.OperationResult, error) {
 	if err != nil {
 		a.logger.Error(err, "Failed to update the build pipelineRun with new annotations",
 			"pipelineRun.Name", a.pipelineRun.Name)
+		return reconciler.RequeueWithError(err)
 	}
 
 	return reconciler.ContinueProcessing()
--- a/controllers/snapshot/snapshot_adapter.go
+++ b/controllers/snapshot/snapshot_adapter.go
@@ -94,7 +94,7 @@ func (a *Adapter) EnsureAllIntegrationTestPipelinesExist() (reconciler.Operation
 			if err != nil {
 				a.logger.Error(err, "Failed to get pipelineRuns for snapshot and scenario",
 					"integrationTestScenario.Name", integrationTestScenario.Name)
-				return reconciler.RequeueOnErrorOrStop(err)
+				return reconciler.RequeueWithError(err)
 			}
 			if integrationPipelineRuns != nil && len(*integrationPipelineRuns) > 0 {
 				a.logger.Info("Found existing integrationPipelineRuns",
@@ -106,7 +106,7 @@ func (a *Adapter) EnsureAllIntegrationTestPipelinesExist() (reconciler.Operation
 				pipelineRun, err := a.createIntegrationPipelineRun(a.application, &integrationTestScenario, a.snapshot)
 				if err != nil {
 					a.logger.Error(err, "Failed to create pipelineRun for snapshot and scenario")
-					return reconciler.RequeueOnErrorOrStop(err)
+					return reconciler.RequeueWithError(err)
 				}
 				a.logger.LogAuditEvent("IntegrationTestscenario pipeline has been created", pipelineRun, h.LogActionAdd,
 					"integrationTestScenario.Name", integrationTestScenario.Name)
@@ -161,7 +161,7 @@ func (a *Adapter) EnsureCreationOfEnvironment() (reconciler.OperationResult, err
 
 	if err != nil {
 		a.logger.Error(err, "Failed to get Integration test scenarios for the following application")
-		return reconciler.RequeueOnErrorOrStop(err)
+		return reconciler.RequeueWithError(err)
 	}
 	if integrationTestScenarios == nil {
 		a.logger.Info("No integration test scenario found for Application")
--- a/controllers/binding/snapshotenvironmentbinding_adapter.go
+++ b/controllers/binding/snapshotenvironmentbinding_adapter.go
@@ -81,7 +81,7 @@ func (a *Adapter) EnsureIntegrationTestPipelineForScenarioExists() (reconciler.O
 			a.logger.Error(err, "Failed to get latest pipelineRun for snapshot and scenario",
 				"snapshot", a.snapshot,
 				"integrationTestScenario", a.integrationTestScenario)
-			return reconciler.RequeueOnErrorOrStop(err)
+			return reconciler.RequeueWithError(err)
 		}
 		if integrationPipelineRun != nil {
 			a.logger.Info("Found existing integrationPipelineRun",
@@ -95,7 +95,7 @@ func (a *Adapter) EnsureIntegrationTestPipelineForScenarioExists() (reconciler.O
 			pipelineRun, err := a.createIntegrationPipelineRunWithEnvironment(a.application, a.integrationTestScenario, a.snapshot, a.environment)
 			if err != nil {
 				a.logger.Error(err, "Failed to create pipelineRun for snapshot, environment and scenario")
-				return reconciler.RequeueOnErrorOrStop(err)
+				return reconciler.RequeueWithError(err)
 			}
 			a.logger.LogAuditEvent("PipelineRun for snapshot created", pipelineRun, h.LogActionAdd,
 				"snapshot.Name", a.snapshot.Name)
@@ -119,19 +119,19 @@ func (a *Adapter) EnsureEphemeralEnvironmentsCleanedUp() (reconciler.OperationRe
 	_, err := gitops.MarkSnapshotAsFailed(a.client, a.context, a.snapshot, snapshotErrorMessage)
 	if err != nil {
 		a.logger.Error(err, "Failed to Update Snapshot status")
-		return reconciler.RequeueOnErrorOrStop(err)
+		return reconciler.RequeueWithError(err)
 	}
 
 	deploymentTargetClaim, err := a.loader.GetDeploymentTargetClaimForEnvironment(a.client, a.context, a.environment)
 	if err != nil {
 		a.logger.Error(err, "failed to find deploymentTargetClaim defined in environment %s", a.environment.Name)
-		return reconciler.RequeueOnErrorOrStop(err)
+		return reconciler.RequeueWithError(err)
 	}
 
 	err = h.CleanUpEphemeralEnvironments(a.client, &a.logger, a.context, a.environment, deploymentTargetClaim)
 	if err != nil {
 		a.logger.Error(err, "Failed to delete the Ephemeral Environment")
-		return reconciler.RequeueOnErrorOrStop(err)
+		return reconciler.RequeueWithError(err)
 	}
 
 	return reconciler.ContinueProcessing()
--- a/controllers/scenario/scenario_adapter.go
+++ b/controllers/scenario/scenario_adapter.go
@@ -121,7 +121,7 @@ func (a *Adapter) EnsureCreatedScenarioIsValid() (reconciler.OperationResult, er
 		err := a.client.Status().Patch(a.context, a.scenario, patch)
 		if err != nil {
 			a.logger.Error(err, "Failed to update Scenario")
-			return reconciler.ContinueProcessing()
+			return reconciler.RequeueWithError(err)
 		}
 		a.logger.LogAuditEvent("IntegrationTestScenario marked as Valid", a.scenario, h.LogActionUpdate)
 	}


controllers/pipeline/pipeline_adapter.go
```diff
@@ -36,6 +36,10 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
+	requeueDelay = 10 * time.Second
```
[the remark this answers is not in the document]
hongweiliu17: I reference the time in https://github.com/redhat-appstudio/managed-gitops/tree/main/appstudio-controller/controllers/appstudio.redhat.com when they meet some intermittent error, but I'd like to know your ideas. Thanks.

controllers/snapshot/snapshot_adapter.go
```diff
@@ -251,7 +256,7 @@ TestScenarioLoop:
 				"snapshot", a.snapshot.Name,
 				"environment.Name", copyEnv.Name,
 				"snapshot.Spec.Components", a.snapshot.Spec.Components)
-			return reconciler.RequeueWithError(err)
```
MartinBasti, line 254: Why this is not error anymore?

controllers/snapshot/snapshot_adapter.go
```diff
@@ -237,7 +242,7 @@ TestScenarioLoop:
 
 		if err != nil {
 			a.logger.Error(err, "Copying of environment failed")
-			return reconciler.RequeueWithError(err)
+			return reconciler.RequeueAfter(requeueDelay, err)
```
MartinBasti, line 245: Why do we need manual requeueDelay, when we have by default exponential backoff retries when error happen in RequeueWithError.

MartinBasti (changes requested):
I don't understand purpose of these changes, I may need explanation like for 5y old

EDIT: mean point 2. RequeueAfter changes

controllers/snapshot/snapshot_adapter.go
hongweiliu17: Comparing with `RequeueWithError`, `reconciler.RequeueAfter(requeueDelay, err)` is having an additional delay with requeue.

hongweiliu17:
> I don't understand purpose of these changes, I may need explanation like for 5y old
> 
> EDIT: mean point 2. RequeueAfter changes

I'll prepare a short demo for it.

hongweiliu17:
> I don't understand purpose of these changes, I may need explanation like for 5y old
> 
> EDIT: mean point 2. RequeueAfter changes

I make a [demo](https://drive.google.com/file/d/1I4W-0UiJ1361Axaw_iafXv_9NI-leJLY/view) for this PR.

controllers/snapshot/snapshot_controller.go
```diff
@@ -156,5 +159,8 @@ func setupControllerWithManager(manager ctrl.Manager, reconciler *Reconciler) er
 		For(&applicationapiv1alpha1.Snapshot{}).
 		WithEventFilter(predicate.Or(
 			gitops.IntegrationSnapshotChangePredicate())).
+		WithOptions(controller.Options{
+			RateLimiter: workqueue.NewItemExponentialFailureRateLimiter(2*time.Second, 1*time.Hour),
```
MartinBasti, line 163: Why this change? Was default rate limit 2-100sec not enough?

hongweiliu17: I wanted to find the place to set the default rate limit but I didn't find it. Can you help to paste the link. I want to read it. Thanks!

MartinBasti:
According this article https://danielmangum.com/posts/controller-runtime-client-go-rate-limiting/

image

Is the RequeueAfter value even used when `err != nil`?

controllers/snapshot/snapshot_controller.go
MartinBasti: https://github.com/kubernetes/client-go/blob/master/util/workqueue/default_rate_limiters.go#L39 This one

I had a typo in my previous msg, 1000sec (~17min what we see in logs)

hongweiliu17: Thanks very much. Let me remove it.

hongweiliu17:
> 
If the `err == nil` and  `requeue is false`, don't requeue, otherwise, requeue.

hongweiliu17:
> According this article https://danielmangum.com/posts/controller-runtime-client-go-rate-limiting/
> 
> image
> 
> Is the RequeueAfter value even used when `err != nil`?

If the `err == nil` and  `requeue is false`, don't requeue, otherwise, requeue. `RequeueAfter` is defined from `RequeueAfter: result.RequeueDelay`, otherwise, it is set to `0`, I think.

MartinBasti:
> > According this article https://danielmangum.com/posts/controller-runtime-client-go-rate-limiting/
> > image
> > Is the RequeueAfter value even used when `err != nil`?
> 
> If the `err == nil` and `requeue is false`, don't requeue, otherwise, requeue. `RequeueAfter` is defined from `RequeueAfter: result.RequeueDelay`, otherwise, it is set to `0`, I think.

Here, err is not nil, but we have reqeueuDelay. According the table it means, that requeueDelay will be ignored and just standard error rate limiting will be used.
```
		if err != nil {
			a.logger.Error(err, "Copying of environment failed")
			return reconciler.RequeueAfter(requeueDelay, err)
```

hongweiliu17:
> > > According this article https://danielmangum.com/posts/controller-runtime-client-go-rate-limiting/
> > > image
> > > Is the RequeueAfter value even used when `err != nil`?
> > 
> > 
> > If the `err == nil` and `requeue is false`, don't requeue, otherwise, requeue. `RequeueAfter` is defined from `RequeueAfter: result.RequeueDelay`, otherwise, it is set to `0`, I think.
> 
> Here, err is not nil, but we have reqeueuDelay. According the table it means, that requeueDelay will be ignored and just standard error rate limiting will be used.
> 
> ```
> 		if err != nil {
> 			a.logger.Error(err, "Copying of environment failed")
> 			return reconciler.RequeueAfter(requeueDelay, err)
> ```

Yes. According to the logic in [controller-runtime](https://github.com/kubernetes-sigs/controller-runtime/blob/16bf3ad036b908d897543c415fcc0bafc5cec711/pkg/internal/controller/controller.go#L297-L311), https://github.com/redhat-appstudio/operator-goodies/blob/main/reconciler/handler.go#L27 looks incorrect, the `RequeueAfter` in code `return ctrl.Result{RequeueAfter: result.RequeueDelay}, err` can't take effect totally.

MartinBasti:
> > > > According this article https://danielmangum.com/posts/controller-runtime-client-go-rate-limiting/
> > > > image
> > > > Is the RequeueAfter value even used when `err != nil`?
> > > 
> > > 
> > > If the `err == nil` and `requeue is false`, don't requeue, otherwise, requeue. `RequeueAfter` is defined from `RequeueAfter: result.RequeueDelay`, otherwise, it is set to `0`, I think.
> > 
> > 
> > Here, err is not nil, but we have reqeueuDelay. According the table it means, that requeueDelay will be ignored and just standard error rate limiting will be used.
> > ```
> > 		if err != nil {
> > 			a.logger.Error(err, "Copying of environment failed")
> > 			return reconciler.RequeueAfter(requeueDelay, err)
> > ```
> 
> Yes. According to the logic in [controller-runtime](https://github.com/kubernetes-sigs/controller-runtime/blob/16bf3ad036b908d897543c415fcc0bafc5cec711/pkg/internal/controller/controller.go#L297-L311), https://github.com/redhat-appstudio/operator-goodies/blob/main/reconciler/handler.go#L27 looks incorrect, the `RequeueAfter` in code `return ctrl.Result{RequeueAfter: result.RequeueDelay}, err` can't take effect totally.

Ah nice, thanks for confirming. So we can remove `RequeueDelay` change from this PR and open an issue for them (for operator-goodies).

hongweiliu17:
> > > > > According this article https://danielmangum.com/posts/controller-runtime-client-go-rate-limiting/
> > > > > image
> > > > > Is the RequeueAfter value even used when `err != nil`?
> > > > 
> > > > 
> > > > If the `err == nil` and `requeue is false`, don't requeue, otherwise, requeue. `RequeueAfter` is defined from `RequeueAfter: result.RequeueDelay`, otherwise, it is set to `0`, I think.
> > > 
> > > 
> > > Here, err is not nil, but we have reqeueuDelay. According the table it means, that requeueDelay will be ignored and just standard error rate limiting will be used.
> > > ```
> > > 		if err != nil {
> > > 			a.logger.Error(err, "Copying of environment failed")
> > > 			return reconciler.RequeueAfter(requeueDelay, err)
> > > ```
> > 
> > 
> > Yes. According to the logic in [controller-runtime](https://github.com/kubernetes-sigs/controller-runtime/blob/16bf3ad036b908d897543c415fcc0bafc5cec711/pkg/internal/controller/controller.go#L297-L311), https://github.com/redhat-appstudio/operator-goodies/blob/main/reconciler/handler.go#L27 looks incorrect, the `RequeueAfter` in code `return ctrl.Result{RequeueAfter: result.RequeueDelay}, err` can't take effect totally.
> 
> Ah nice, thanks for confirming. So we can remove `RequeueDelay` change from this PR and open an issue for them (for operator-goodies).

Do you know the place we can file issue for it? cc @dirgim 

Maybe we can just set a requeueAfter in our code now, then we don't need to update our code once operator-goodies fix it?

controllers/pipeline/pipeline_adapter.go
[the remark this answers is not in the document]
hongweiliu17: Removed the manual delay since we can't implement it currently due to the issue in https://github.com/redhat-appstudio/operator-goodies/blob/main/reconciler/handler.go#L27.

controllers/snapshot/snapshot_adapter.go
hongweiliu17: Removed the manual delay since we can't implement it currently due to the issue in https://github.com/redhat-appstudio/operator-goodies/blob/main/reconciler/handler.go#L27.

controllers/snapshot/snapshot_adapter.go
hongweiliu17: Removed the manual delay since we can't implement it currently due to the issue in https://github.com/redhat-appstudio/operator-goodies/blob/main/reconciler/handler.go#L27.

controllers/snapshot/snapshot_controller.go
hongweiliu17: Removed.

hongweiliu17:
> > > > > > According this article https://danielmangum.com/posts/controller-runtime-client-go-rate-limiting/
> > > > > > image
> > > > > > Is the RequeueAfter value even used when `err != nil`?
> > > > > 
> > > > > 
> > > > > If the `err == nil` and `requeue is false`, don't requeue, otherwise, requeue. `RequeueAfter` is defined from `RequeueAfter: result.RequeueDelay`, otherwise, it is set to `0`, I think.
> > > > 
> > > > 
> > > > Here, err is not nil, but we have reqeueuDelay. According the table it means, that requeueDelay will be ignored and just standard error rate limiting will be used.
> > > > ```
> > > > 		if err != nil {
> > > > 			a.logger.Error(err, "Copying of environment failed")
> > > > 			return reconciler.RequeueAfter(requeueDelay, err)
> > > > ```
> > > 
> > > 
> > > Yes. According to the logic in [controller-runtime](https://github.com/kubernetes-sigs/controller-runtime/blob/16bf3ad036b908d897543c415fcc0bafc5cec711/pkg/internal/controller/controller.go#L297-L311), https://github.com/redhat-appstudio/operator-goodies/blob/main/reconciler/handler.go#L27 looks incorrect, the `RequeueAfter` in code `return ctrl.Result{RequeueAfter: result.RequeueDelay}, err` can't take effect totally.
> > 
> > 
> > Ah nice, thanks for confirming. So we can remove `RequeueDelay` change from this PR and open an issue for them (for operator-goodies).
> 
> Do you know the place we can file issue for it? cc @dirgim
> 
> Maybe we can just set a requeueAfter in our code now, then we don't need to update our code once operator-goodies fix it?

Removed the manual delay since we can't implement it currently due to the issue in https://github.com/redhat-appstudio/operator-goodies/blob/main/reconciler/handler.go#L27.

Thanks for the review! @MartinBasti

MartinBasti:
> Maybe we can just set a requeueAfter in our code now, then we don't need to update our code once operator-goodies fix it?

I don't understand how goodies can fix it, this is how k8s works, if there is an error `requeueAfter` won't apply at all

MartinBasti (approved):

jinqi7 (approved):
LGTM

PR #247, opened 2023-07-20
fix(STONEINTG-469): add SetSnapshotIntegrationStatusAsInvalidandNoRequ

jinqi7:
1. Added new Snapshot status "ErrorOccured" and replaced "Invalid" status in the snapshot controller logic with it.
2. Moved the RegisterNewSnapshot() function from NewSnapshot() to the place following a.client.Create(a.context, *Snapshot) action

### DIFF
--- a/controllers/pipeline/pipeline_adapter.go
+++ b/controllers/pipeline/pipeline_adapter.go
@@ -28,6 +28,7 @@ import (
 	"github.com/redhat-appstudio/integration-service/gitops"
 	h "github.com/redhat-appstudio/integration-service/helpers"
 	"github.com/redhat-appstudio/integration-service/loader"
+	"github.com/redhat-appstudio/integration-service/metrics"
 	"github.com/redhat-appstudio/integration-service/status"
 	"github.com/redhat-appstudio/integration-service/tekton"
 	"github.com/redhat-appstudio/operator-goodies/reconciler"
@@ -117,6 +118,7 @@ func (a *Adapter) EnsureSnapshotExists() (reconciler.OperationResult, error) {
 		a.logger.Error(err, "Failed to create Snapshot")
 		return reconciler.RequeueWithError(err)
 	}
+	go metrics.RegisterNewSnapshot()
 
 	a.logger.LogAuditEvent("Created new Snapshot", expectedSnapshot, h.LogActionAdd,
 		"snapshot.Name", expectedSnapshot.Name,
@@ -590,6 +592,7 @@ func (a *Adapter) createCompositeSnapshotsIfConflictExists(application *applicat
 			if err != nil {
 				return nil, err
 			}
+			go metrics.RegisterNewSnapshot()
 			a.logger.LogAuditEvent("CompositeSnapshot created", compositeSnapshot, h.LogActionAdd,
 				"snapshot.Spec.Components", compositeSnapshot.Spec.Components)
 			return compositeSnapshot, nil
--- a/gitops/snapshot.go
+++ b/gitops/snapshot.go
@@ -101,6 +101,9 @@ const (
 	// AppStudioIntegrationStatusInvalid is the reason that's set when the AppStudio integration gets into an invalid state.
 	AppStudioIntegrationStatusInvalid = "Invalid"
 
+	// AppStudioIntegrationStatusErrorOccured is the reason that's set when the AppStudio integration gets into an error state.
+	AppStudioIntegrationStatusErrorOccured = "ErrorOccured"
+
 	// AppStudioIntegrationStatusValid is the reason that's set when the AppStudio integration gets into an valid state.
 	AppStudioIntegrationStatusValid = "Valid"
 
@@ -169,7 +172,18 @@ func SetSnapshotIntegrationStatusAsInvalid(snapshot *applicationapiv1alpha1.Snap
 		Message: message,
 	}
 	meta.SetStatusCondition(&snapshot.Status.Conditions, condition)
-	go metrics.RegisterInvalidSnapshot(condition.Type, condition.Reason)
+	go metrics.RegisterInvalidSnapshot(AppStudioIntegrationStatusCondition, AppStudioIntegrationStatusInvalid)
+}
+
+// SetSnapshotIntegrationStatusAsError sets the AppStudio integration status condition for the Snapshot to error.
+func SetSnapshotIntegrationStatusAsError(snapshot *applicationapiv1alpha1.Snapshot, message string) {
+	condition := metav1.Condition{
+		Type:    AppStudioIntegrationStatusCondition,
+		Status:  metav1.ConditionFalse,
+		Reason:  AppStudioIntegrationStatusErrorOccured,
+		Message: message,
+	}
+	meta.SetStatusCondition(&snapshot.Status.Conditions, condition)
 }
 
 // MarkSnapshotIntegrationStatusAsInProgress sets the AppStudio integration status condition for the Snapshot to In Progress.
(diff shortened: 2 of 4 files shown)


controllers/pipeline/pipeline_adapter.go
```diff
@@ -197,7 +199,7 @@ func (a *Adapter) EnsureSnapshotPassedAllTests() (reconciler.OperationResult, er
 		if compositeSnapshot != nil {
 			a.logger.Info("The global component list has changed in the meantime, marking snapshot as Invalid",
 				"snapshot.Name", existingSnapshot.Name)
-			gitops.SetSnapshotIntegrationStatusAsInvalid(existingSnapshot,
+			gitops.SetSnapshotIntegrationStatusAsInvalidandNoRequeue(existingSnapshot,
```
dirgim, line 202: I'm thinking that we might want to introduce a new status condition in order to separate invalid Snapshots from Snapshots that encountered an error that can be recovered from, wdyt @jinqi7 ?

jinqi7: @digim, it sounds good for me. It can seperate the snapshots which is marked as invalid and stopped with that ones  which is marked as invalid and requeued.  Will you suggest a name of  the status?

gitops/snapshot.go
```diff
@@ -169,7 +169,12 @@ func SetSnapshotIntegrationStatusAsInvalid(snapshot *applicationapiv1alpha1.Snap
 		Message: message,
 	}
 	meta.SetStatusCondition(&snapshot.Status.Conditions, condition)
-	go metrics.RegisterInvalidSnapshot(condition.Type, condition.Reason)
+}
+
+// SetSnapshotIntegrationStatusAsInvalidandNoRequeue sets the AppStudio integration status condition for the Snapshot to invalid and it'll not be requeued.
+func SetSnapshotIntegrationStatusAsInvalidandNoRequeue(snapshot *applicationapiv1alpha1.Snapshot, message string) {
+	SetSnapshotIntegrationStatusAsInvalid(snapshot, message)
+	go metrics.RegisterInvalidSnapshot(AppStudioIntegrationStatusCondition, AppStudioIntegrationStatusInvalid)
```
14rcole, line 177: What's the difference between this line here and [this one](https://github.com/redhat-appstudio/integration-service/blob/main/gitops/snapshot.go#L172) in `SetSnapshotIntegrationStatusAsInvalid()`?

gitops/snapshot_test.go
```diff
@@ -127,7 +127,7 @@ var _ = Describe("Gitops functions for managing Snapshots", Ordered, func() {
 	})
 
 	It("ensures the Snapshots status can be marked as invalid", func() {
-		gitops.SetSnapshotIntegrationStatusAsInvalid(hasSnapshot, "Test message")
+		gitops.SetSnapshotIntegrationStatusAsInvalidandNoRequeue(hasSnapshot, "Test message")
```
14rcole, line 130: Nit: capitalize "and" in function calls and function definition

gitops/snapshot.go
jinqi7: No difference.  I removed that line from SetSnapshotIntegrationStatusAsInvalid()  since  the Snapshot will be requeued in the situation. Then Snapshot is not in the final status and will not be recorded.

gitops/snapshot_test.go
jinqi7: Thanks. Will change

controllers/pipeline/pipeline_adapter.go
dirgim: Wdyt about a more generic `ErrorOccured`  to have some commonality with the other resources like e.g. in [gitOps here](https://github.com/redhat-appstudio/managed-gitops/blob/ac1312cc4eea936ea9ebe4845b9a48ba5c9d0474/appstudio-controller/controllers/appstudio.redhat.com/environment_controller.go#L265-L268)?

gitops/snapshot.go
14rcole: Got it, thanks for the clarification!

gitops/snapshot.go
```diff
@@ -169,7 +172,18 @@ func SetSnapshotIntegrationStatusAsInvalid(snapshot *applicationapiv1alpha1.Snap
...
 	meta.SetStatusCondition(&snapshot.Status.Conditions, condition)
-	go metrics.RegisterInvalidSnapshot(condition.Type, condition.Reason)
+	go metrics.RegisterInvalidSnapshot(AppStudioIntegrationStatusCondition, AppStudioIntegrationStatusInvalid)
+}
+
+// SetSnapshotIntegrationStatusAsError sets the AppStudio integration status condition for the Snapshot to error.
+func SetSnapshotIntegrationStatusAsError(snapshot *applicationapiv1alpha1.Snapshot, message string) {
+	condition := metav1.Condition{
+		Type:    AppStudioIntegrationStatusCondition,
+		Status:  metav1.ConditionFalse,
```
dirgim, line 182: I think this should be set to `metav1.ConditionTrue` because this status condition reports if an error occured.

gitops/snapshot_test.go
```diff
@@ -126,8 +126,8 @@ var _ = Describe("Gitops functions for managing Snapshots", Ordered, func() {
 		Expect(meta.IsStatusConditionTrue(updatedSnapshot.Status.Conditions, gitops.AppStudioTestSuceededCondition)).To(BeFalse())
 	})
 
-	It("ensures the Snapshots status can be marked as invalid", func() {
-		gitops.SetSnapshotIntegrationStatusAsInvalid(hasSnapshot, "Test message")
+	It("ensures the Snapshots status can be marked as error", func() {
+		gitops.SetSnapshotIntegrationStatusAsError(hasSnapshot, "Test message")
 		Expect(hasSnapshot).NotTo(BeNil())
 		Expect(hasSnapshot.Status.Conditions).NotTo(BeNil())
 		Expect(meta.IsStatusConditionTrue(hasSnapshot.Status.Conditions, gitops.AppStudioIntegrationStatusCondition)).To(BeFalse())
```
dirgim, line 133: Should we test for the `AppStudioIntegrationStatusErrorOccured` condition here?

jinqi7: Currently we just have AppStudioIntegrationStatusErrorOccured, it's used in the "reason" field.  There is no AppStudioIntegrationStatusErrorOccuredCondition.

dirgim: My bad, I missed that it was a reason and not a new condition. We can maybe check the condition reason to confirm that it's `ErrorOccured`

controllers/pipeline/pipeline_adapter.go
dirgim: Ack, thanks for following up on this

gitops/snapshot_test.go
jinqi7: Added the check for the condition reason. Thanks!

14rcole (approved):
LGTM

gitops/snapshot.go
```diff
@@ -227,6 +241,18 @@ func IsSnapshotNotStarted(snapshot *applicationapiv1alpha1.Snapshot) bool {
 	return false
 }
 
+// IsSnapshotError if the AppStudio Integration Status condition is in ErrorOcurred status.
+func IsSnapshotError(snapshot *applicationapiv1alpha1.Snapshot) bool {
+	condition := meta.FindStatusCondition(snapshot.Status.Conditions, AppStudioIntegrationStatusCondition)
+	if condition == nil {
+		condition = meta.FindStatusCondition(snapshot.Status.Conditions, LegacyIntegrationStatusCondition)
+	}
+	if condition.Reason == AppStudioIntegrationStatusErrorOccured {
```
hongweiliu17, line 250: Should we change it to `if condition == nil || condition.Reason == AppStudioIntegrationStatusErrorOccured` in case `LegacyIntegrationStatusCondition` can't be found either?

jinqi7: If condition == nil, the expected return value is false.

hongweiliu17: Ah, yes. You are right.

hongweiliu17 (approved):
LGTM

gitops/snapshot_test.go
dirgim: Thanks, looks good!

dirgim (approved):
LGTM

PR #263, opened 2023-08-03
fix(STONEINTG-469): fix the negative number in SnapshotConcurrentTotal

jinqi7:
Added checking before the status of snapshot is updated to avoid repetition action
After the snapshot is set to Invalid status, return immediately and avoid to be set to "Passed" or " Failed" again.

### DIFF
--- a/controllers/integrationpipeline/integrationpipeline_adapter.go
+++ b/controllers/integrationpipeline/integrationpipeline_adapter.go
@@ -30,6 +30,7 @@ import (
 	"github.com/redhat-appstudio/integration-service/tekton"
 	"github.com/redhat-appstudio/operator-toolkit/controller"
 	tektonv1beta1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -125,31 +126,46 @@ func (a *Adapter) EnsureSnapshotPassedAllTests() (controller.OperationResult, er
 		if compositeSnapshot != nil {
 			a.logger.Info("The global component list has changed in the meantime, marking snapshot as Invalid",
 				"snapshot.Name", existingSnapshot.Name)
-			gitops.SetSnapshotIntegrationStatusAsInvalid(existingSnapshot,
-				"The global component list has changed in the meantime, superseding with a composite snapshot")
-			a.logger.LogAuditEvent("Snapshot integration status condition marked as invalid, the global component list has changed in the meantime",
-				existingSnapshot, h.LogActionUpdate)
+			if !gitops.IsSnapshotStatusConditionSet(existingSnapshot, gitops.AppStudioIntegrationStatusCondition,
+				metav1.ConditionFalse, gitops.AppStudioIntegrationStatusInvalid) {
+				patch := client.MergeFrom(existingSnapshot.DeepCopy())
+				gitops.SetSnapshotIntegrationStatusAsInvalid(existingSnapshot,
+					"The global component list has changed in the meantime, superseding with a composite snapshot")
+				err := a.client.Status().Patch(a.context, existingSnapshot, patch)
+				if err != nil {
+					a.logger.Error(err, "Failed to update the status to Invalid for the snapshot",
+						"snapshot.Name", existingSnapshot.Name)
+					return controller.RequeueWithError(err)
+				}
+				a.logger.LogAuditEvent("Snapshot integration status condition marked as invalid, the global component list has changed in the meantime",
+					existingSnapshot, h.LogActionUpdate)
+			}
+			return controller.ContinueProcessing()
 		}
 	}
 
 	// If all Integration Pipeline runs passed, mark the snapshot as succeeded, otherwise mark it as failed
 	// This updates the Snapshot resource on the cluster
 	if allIntegrationPipelineRunsPassed {
-		existingSnapshot, err = gitops.MarkSnapshotAsPassed(a.client, a.context, existingSnapshot, "All Integration Pipeline tests passed")
-		if err != nil {
-			a.logger.Error(err, "Failed to Update Snapshot AppStudioTestSucceeded status")
-			return controller.RequeueWithError(err)
+		if !gitops.IsSnapshotStatusConditionSet(existingSnapshot, gitops.AppStudioTestSuceededCondition, metav1.ConditionTrue, "") {
+			existingSnapshot, err = gitops.MarkSnapshotAsPassed(a.client, a.context, existingSnapshot, "All Integration Pipeline tests passed")
+			if err != nil {
+				a.logger.Error(err, "Failed to Update Snapshot AppStudioTestSucceeded status")
+				return controller.RequeueWithError(err)
+			}
+			a.logger.LogAuditEvent("Snapshot integration status condition marked as passed, all Integration PipelineRuns succeeded",
+				existingSnapshot, h.LogActionUpdate)
 		}
-		a.logger.LogAuditEvent("Snapshot integration status condition marked as passed, all Integration PipelineRuns succeeded",
-			existingSnapshot, h.LogActionUpdate)
 	} else {
-		existingSnapshot, err = gitops.MarkSnapshotAsFailed(a.client, a.context, existingSnapshot, "Some Integration pipeline tests failed")
-		if err != nil {
-			a.logger.Error(err, "Failed to Update Snapshot AppStudioTestSucceeded status")
-			return controller.RequeueWithError(err)
+		if !gitops.IsSnapshotStatusConditionSet(existingSnapshot, gitops.AppStudioTestSuceededCondition, metav1.ConditionFalse, "") {
+			existingSnapshot, err = gitops.MarkSnapshotAsFailed(a.client, a.context, existingSnapshot, "Some Integration pipeline tests failed")
+			if err != nil {
+				a.logger.Error(err, "Failed to Update Snapshot AppStudioTestSucceeded status")
+				return controller.RequeueWithError(err)
+			}
+			a.logger.LogAuditEvent("Snapshot integration status condition marked as failed, some tests within Integration PipelineRuns failed",
+				existingSnapshot, h.LogActionUpdate)
 		}
-		a.logger.LogAuditEvent("Snapshot integration status condition marked as failed, some tests within Integration PipelineRuns failed",
-			existingSnapshot, h.LogActionUpdate)
 	}
 
 	return controller.ContinueProcessing()
--- a/gitops/snapshot.go
+++ b/gitops/snapshot.go
@@ -268,6 +268,24 @@ func IsSnapshotValid(snapshot *applicationapiv1alpha1.Snapshot) bool {
 	return false
 }
 
+// IsSnapshotStatusConditionSet checks if the condition with the conditionType in the status of Snapshot has been marked as the conditionStatus and reason.
+func IsSnapshotStatusConditionSet(snapshot *applicationapiv1alpha1.Snapshot, conditionType string, conditionStatus metav1.ConditionStatus, reason string) bool {
+	condition := meta.FindStatusCondition(snapshot.Status.Conditions, conditionType)
+	if condition == nil && conditionType == AppStudioTestSuceededCondition {
+		condition = meta.FindStatusCondition(snapshot.Status.Conditions, LegacyTestSuceededCondition)
+	}
+	if condition == nil && conditionType == AppStudioIntegrationStatusCondition {
+		condition = meta.FindStatusCondition(snapshot.Status.Conditions, LegacyIntegrationStatusCondition)
+	}
+	if condition == nil || condition.Status != conditionStatus {
+		return false
+	}
+	if reason != "" && reason != condition.Reason {
+		return false
+	}
+	return true
+}
+
 // ValidateImageDigest checks if image url contains valid digest, return error if check fails
 func ValidateImageDigest(imageUrl string) error {
 	_, err := name.NewDigest(imageUrl)
--- a/controllers/integrationpipeline/integrationpipeline_adapter_test.go
+++ b/controllers/integrationpipeline/integrationpipeline_adapter_test.go
@@ -177,30 +177,6 @@ var _ = Describe("Pipeline Adapter", Ordered, func() {
 		}
 		Expect(k8sClient.Create(ctx, hasComp2)).Should(Succeed())
 
-		hasSnapshot = &applicationapiv1alpha1.Snapshot{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      "snapshot-sample",
-				Namespace: "default",
-				Labels: map[string]string{
-					gitops.SnapshotTypeLabel:      "component",
-					gitops.SnapshotComponentLabel: hasComp.Name,
-				},
-				Annotations: map[string]string{
-					gitops.PipelineAsCodeInstallationIDAnnotation: "123",
-				},
-			},
-			Spec: applicationapiv1alpha1.SnapshotSpec{
-				Application: hasApp.Name,
-				Components: []applicationapiv1alpha1.SnapshotComponent{
-					{
-						Name:           hasComp.Name,
-						ContainerImage: SampleImage,
-					},
-				},
-			},
-		}
-		Expect(k8sClient.Create(ctx, hasSnapshot)).Should(Succeed())
-
 		integrationTestScenario = &v1beta1.IntegrationTestScenario{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      "example-pass",
@@ -312,6 +288,30 @@ var _ = Describe("Pipeline Adapter", Ordered, func() {
 	})
 
 	BeforeEach(func() {
+		hasSnapshot = &applicationapiv1alpha1.Snapshot{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "snapshot-sample",
+				Namespace: "default",
+				Labels: map[string]string{
+					gitops.SnapshotTypeLabel:      "component",
+					gitops.SnapshotComponentLabel: hasComp.Name,
+				},
+				Annotations: map[string]string{
+					gitops.PipelineAsCodeInstallationIDAnnotation: "123",
+				},
+			},
+			Spec: applicationapiv1alpha1.SnapshotSpec{
+				Application: hasApp.Name,
+				Components: []applicationapiv1alpha1.SnapshotComponent{
+					{
+						Name:           hasComp.Name,
+						ContainerImage: SampleImage,
+					},
+				},
+			},
+		}
+		Expect(k8sClient.Create(ctx, hasSnapshot)).Should(Succeed())
+
 		integrationPipelineRunComponent = &tektonv1beta1.PipelineRun{
 			ObjectMeta: metav1.ObjectMeta{
 				Name:      "pipelinerun-component-sample",
@@ -365,7 +365,9 @@ var _ = Describe("Pipeline Adapter", Ordered, func() {
 	})
 
 	AfterEach(func() {
-		err := k8sClient.Delete(ctx, integrationPipelineRunComponent)
+		err := k8sClient.Delete(ctx, hasSnapshot)
+		Expect(err == nil || k8serrors.IsNotFound(err)).To(BeTrue())
+		err = k8sClient.Delete(ctx, integrationPipelineRunComponent)
 		Expect(err == nil || k8serrors.IsNotFound(err)).To(BeTrue())
 	})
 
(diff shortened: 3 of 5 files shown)


gitops/snapshot.go
```diff
@@ -320,6 +320,7 @@ func NewSnapshot(application *applicationapiv1alpha1.Application, snapshotCompon
 			Components:  *snapshotComponents,
 		},
 	}
+	go metrics.RegisterNewSnapshot()
```
dirgim, line 323: Won't this register a new Snapshot even when we realize that there is a matching snapshot that already exists like e.g. [here](https://github.com/redhat-appstudio/integration-service/blob/8075f54853a135480cf43610e93032eb66d34e62/controllers/pipeline/pipeline_adapter.go#L100)?

controllers/pipeline/pipeline_adapter.go
```diff
@@ -118,7 +117,6 @@ func (a *Adapter) EnsureSnapshotExists() (controller.OperationResult, error) {
 		a.logger.Error(err, "Failed to create Snapshot")
 		return controller.RequeueWithError(err)
 	}
-	go metrics.RegisterNewSnapshot()
```
hongweiliu17, line 121: It doesn't take effect here and in next place?

gitops/snapshot.go
jinqi7: We ever discussed it before. Currently, I don't have a good solution for it.

controllers/pipeline/pipeline_adapter.go
jinqi7: Seems both.

hongweiliu17: Why? I think when we meet error and `return controller.RequeueWithError(err)`, the `go metrics.RegisterNewSnapshot()` will not be called, but this is expected. Otherwise, `go metrics.RegisterNewSnapshot()` will be called.

jinqi7: But the number of "snapshot_attempt_concurrent_requests" keeps "0" after it's called from the testing. It's weird.

hongweiliu17: Is is really called? I feel you may add some log to see if it is called.

dirgim:
I'm thinking that we should change the [MarkSnapshotAsInvalid](https://github.com/redhat-appstudio/integration-service/blob/main/gitops/snapshot.go#L167) function to first check if the Snapshot is already invalid before updating the status and metrics. I think we should do the same for other similar functions too.

controllers/pipeline/pipeline_adapter.go
jinqi7: The number increased for "snapshot_attempt_concurrent_total" after changing it to use "NewGaugeVec".  So, it's called.

jinqi7:
> I'm thinking that we should change the [MarkSnapshotAsInvalid](https://github.com/redhat-appstudio/integration-service/blob/main/gitops/snapshot.go#L167) function to first check if the Snapshot is already invalid before updating the status and metrics. I think we should do the same for other similar functions too.
  currently, the Snapshot is only marked as Invalid when the composite snapshot created after the component snapshot is finished. It will not be requeued. So, I didn't add this check in the PR.

metrics/integration_test.go
```diff
@@ -33,8 +33,6 @@ import (
 var _ = Describe("Metrics Integration", Ordered, func() {
 	BeforeAll(func() {
 		metrics.Registry.Unregister(SnapshotCreatedToPipelineRunStartedSeconds)
-		metrics.Registry.Unregister(SnapshotConcurrentTotal)
-		metrics.Registry.Unregister(SnapshotDurationSeconds)
```
MartinBasti, line 37: Is this removal right? Shouldn't we keep SnapshotDurationSeconds here?

metrics/integration.go
```diff
@@ -69,13 +70,13 @@ func RegisterCompletedSnapshot(conditiontype, reason string, startTime metav1.Ti
 		"reason": reason,
 	}
 
-	SnapshotConcurrentTotal.Sub(1)
+	SnapshotConcurrentTotal.WithLabelValues().Sub(1)
```
MartinBasti, line 73: If we don't specify any labels, why do we use GaugeVec? Ins't it just Gauge if we don't add "vectors" there then?

metrics/integration.go
```diff
@@ -30,11 +30,12 @@ var (
 		},
 	)
 
-	SnapshotConcurrentTotal = prometheus.NewGauge(
+	SnapshotConcurrentTotal = prometheus.NewGaugeVec(
```
MartinBasti, line 33: IF we change scalar to vector, this may nee follow-up story to upgdate grafana dashboard.

jinqi7: Why updated grafana dashboard?  Release team has already began to use it for some time.

jinqi7:
> Can you please describe in commit message why and how this change fixes the issue?

I updated the code to use "NewGaugeVec" since the unit test need to unregister/register the metric again by using "NewGauge". It's better to use "Reset".

metrics/integration_test.go
jinqi7: It's right from the test.

metrics/integration.go
jinqi7: Please see the reason below. Thanks

MartinBasti:
> > Can you please describe in commit message why and how this change fixes the issue?
> 
> I updated the code to use "NewGaugeVec" since the unit test need to unregister/register the metric again by using "NewGauge". It's better to use "Reset".

How change in unitests is going to fix wrong data in grafana?

metrics/integration_test.go
MartinBasti: What does it mean "right from the test"?

I can write passing unittest for any broken code.

How `SnapshotDurationSeconds` is related to Concurent Snapshot change?

metrics/integration.go
jinqi7: I saw the "Upgrade" grafana dashboard, so I thought it's to upgrade granfana. I didn't noticed you meant to update the data represent.  It's misunderstanding.

jinqi7:
> How change in unitests is going to fix wrong data in grafana?



> > > Can you please describe in commit message why and how this change fixes the issue?
> > 
> > 
> > I updated the code to use "NewGaugeVec" since the unit test need to unregister/register the metric again by using "NewGauge". It's better to use "Reset".
> 
> How change in unitests is going to fix wrong data in grafana?

Since we'll going to move to "operator-toolkit", we'll going to move to  use "*Vec" metric type. So, I try it and it works indeed.

metrics/integration_test.go
jinqi7: It can be restored.

MartinBasti:
> > How change in unitests is going to fix wrong data in grafana?
> 
> > > > Can you please describe in commit message why and how this change fixes the issue?
> > > 
> > > 
> > > I updated the code to use "NewGaugeVec" since the unit test need to unregister/register the metric again by using "NewGauge". It's better to use "Reset".
> > 
> > 
> > How change in unitests is going to fix wrong data in grafana?
> 
> Since we'll going to move to "operator-toolkit", we'll going to move to use "*Vec" metric type. So, I try it and it works indeed.

So change from Gauge to singledimensional vector of GaugeVec, which is basically the same thing, fixed our grafana data?

jinqi7:
> So change from Gauge to singledimensional vector of GaugeVec, which is basically the same thing, fixed our grafana data?
It's different when I am doing test. The data of snapshot_attempt_concurrent_requests keeps as "0" when the Gauge type is being used.

MartinBasti:
> > So change from Gauge to singledimensional vector of GaugeVec, which is basically the same thing, fixed our grafana data?
> > It's different when I am doing test. The data of snapshot_attempt_concurrent_requests keeps as "0" when the Gauge type is being used. And I can see data change when the GaugeVec is being used.

I don't understand how, https://github.com/prometheus/client_golang/blob/e6d9092bafa813a04c075166690996032302f4b2/prometheus/gauge.go#L235 in source code, it just returns and call standard Gauge from GaugeVec

jinqi7:
> I don't understand how, https://github.com/prometheus/client_golang/blob/e6d9092bafa813a04c075166690996032302f4b2/prometheus/gauge.go#L235 in source code, it just returns and call standard Gauge from GaugeVec

I didn't dig into it to find the root cause.  I just tried some reasonable way to make it work and avoid the negative number.

gitops/snapshot.go
```diff
@@ -125,6 +125,9 @@ var (
 // MarkSnapshotAsPassed updates the AppStudio Test succeeded condition for the Snapshot to passed.
 // If the patch command fails, an error will be returned.
 func MarkSnapshotAsPassed(adapterClient client.Client, ctx context.Context, snapshot *applicationapiv1alpha1.Snapshot, message string) (*applicationapiv1alpha1.Snapshot, error) {
+	if !IsSnapshotNotStarted(snapshot) && HaveAppStudioTestsSucceeded(snapshot) {
```
MartinBasti, line 128: Having this conditionals in setter is very suspicious, it may mean we have a serious issues with code flow.

If snapshot is not ready to be marked as passed, this function shouldn't be even called at first place

jinqi7: How about now ?  The !IsSnapshotNotStarted(snapshot)? is not needed any more.

MartinBasti: Why is not `HaveAppStudioTestsSucceeded` before we even call MarkSnapshotPassed?

controllers/integrationpipeline/integrationpipeline_adapter.go
```diff
@@ -129,6 +129,7 @@ func (a *Adapter) EnsureSnapshotPassedAllTests() (controller.OperationResult, er
 				"The global component list has changed in the meantime, superseding with a composite snapshot")
 			a.logger.LogAuditEvent("Snapshot integration status condition marked as invalid, the global component list has changed in the meantime",
 				existingSnapshot, h.LogActionUpdate)
+			return controller.ContinueProcessing()
```
dirgim, line 132: The issue with this change is that the `SetSnapshotIntegrationStatusAsInvalid()` function doesn't update the resource on the cluster, so the Snapshot CR will never be updated/patched like it happens when calling `MarkSnapshotAs*` functions.
Probably the simplest solution would be to add an additional patch call here.

gitops/snapshot.go
dirgim: I think that the intent here is to first check if the Snapshot has the status condition already set to what we want to set it before unnecessarily calling the `adapterClient.Status().Patch()` and updating the metrics. If it's already `Succeeded`, we don't want to waste time and API calls to try to set it again to `Succeeded`.

That said, looking at it now I agree it would be more idiomatic to check the state of the Snapshot's status before calling these functions.

jinqi7: Adding these checks  is to ensure no repetition action during the requests are requeued  and the integration service is restarted.  When the snapshot is already being marked to some status,  we don't want to mark and count it to the same status again.

controllers/integrationpipeline/integrationpipeline_adapter.go
jinqi7: Thanks! Added the patching code as you suggested. Can you please help to review it again?

gitops/snapshot.go
```diff
@@ -125,6 +125,9 @@ var (
 // MarkSnapshotAsPassed updates the AppStudio Test succeeded condition for the Snapshot to passed.
 // If the patch command fails, an error will be returned.
 func MarkSnapshotAsPassed(adapterClient client.Client, ctx context.Context, snapshot *applicationapiv1alpha1.Snapshot, message string) (*applicationapiv1alpha1.Snapshot, error) {
+	if HaveAppStudioTestsSucceeded(snapshot) {
```
dirgim, line 128: I still think we could make this check in the adapter code before calling  `MarkSnapshotAsPassed()`.

gitops/snapshot.go
```diff
@@ -147,6 +150,9 @@ func MarkSnapshotAsPassed(adapterClient client.Client, ctx context.Context, snap
 // MarkSnapshotAsFailed updates the AppStudio Test succeeded condition for the Snapshot to failed.
 // If the patch command fails, an error will be returned.
 func MarkSnapshotAsFailed(adapterClient client.Client, ctx context.Context, snapshot *applicationapiv1alpha1.Snapshot, message string) (*applicationapiv1alpha1.Snapshot, error) {
+	if HaveAppStudioTestsFailed(snapshot) {
```
dirgim, line 153: I still think we could make this check in the adapter code before calling  `MarkSnapshotAsFailed()`.

gitops/snapshot.go
```diff
@@ -168,6 +174,9 @@ func MarkSnapshotAsFailed(adapterClient client.Client, ctx context.Context, snap
 
 // SetSnapshotIntegrationStatusAsInvalid sets the AppStudio integration status condition for the Snapshot to invalid.
 func SetSnapshotIntegrationStatusAsInvalid(snapshot *applicationapiv1alpha1.Snapshot, message string) {
+	if !IsSnapshotValid(snapshot) {
```
dirgim, line 177: I still think we could make this check in the adapter code before calling `IsSnapshotValid()`.

gitops/snapshot.go
dirgim: The reason for my suggestion is that the `MarkSnapshotAsPassed()` function implies that it will update the Snapshot with this status, and with this change it might just return the unmodified snapshot.

The adapter then reports the update event to the audit log (e.g. [here](https://github.com/redhat-appstudio/integration-service/blob/main/controllers/integrationpipeline/integrationpipeline_adapter.go#L143-L144)), even though the Snapshot might not have been updated at that time.

controllers/integrationpipeline/integrationpipeline_adapter.go
dirgim: Thanks for the change, can you also move the `a.logger.LogAuditEvent` call to happen after the resource is updated(patched)?

gitops/snapshot.go
jinqi7: If so, I am wondering if I need to add the check for all the places that MarkSnapshotAsPassed() are called. Such as  [here](https://github.com/redhat-appstudio/integration-service/blob/main/controllers/integrationpipeline/integrationpipeline_adapter.go#L138) in controllers/integrationpipeline/integrationpipeline_adapter.go?

controllers/integrationpipeline/integrationpipeline_adapter.go
jinqi7: OK.

gitops/snapshot.go
jinqi7: Should we add the check for all the places that MarkSnapshotAsFailed() are called in the code ( integrationpipeline_adapter.go / snapshotenvironmentbinding_adapter.go) ?

gitops/snapshot.go
dirgim: Yes, a new common function might be the way to go. If it can be made to generically handle different statuses, even better.

jinqi7: Then in the code logic, before each of the MarkSnapshotAsPassed/ MarkSnapshotAsFailed/ SetSnapshotIntegrationStatusAsInvalid, we'll add the check that the snapshot has never been marked as "Succeeded/Failed/Invalid"  before.  Right?

jinqi7: Thanks. @dirgim, shall we keep the checking inside the function ?

jinqi7: Can we change the message to something like "Snapshot integration status condition marked/ kept as failed, some tests within Integration PipelineRuns failed"?

jinqi7: I added the function of IsSnapshotStatusConditionSet() for checking the status of the snapshot and add the checking before the MarkSnapshotAsPassed/ MarkSnapshotAsFailed/ SetSnapshotIntegrationStatusAsInvalid.  Please help to review if it's as expected. Thanks

controllers/integrationpipeline/integrationpipeline_adapter.go
dirgim: Thanks!

metrics/integration_test.go
MartinBasti: I'd like to have it restore then

gitops/snapshot.go
dirgim: This looks like a reasonable solution to me.

gitops/snapshot.go
dirgim: Yup, thanks for making the updates.

MartinBasti:
Another point why GaugeVec doesn't help: https://paste.centos.org/view/dd247bcb

If I replace `GaugeVec` with `Gauge` and `reset` with `set`, it works.

jinqi7:
> Another point why GaugeVec doesn't help: https://paste.centos.org/view/dd247bcb
> 
> If I replace `GaugeVec` with `Gauge` and `reset` with `set`, it works.

`set(0)` is to set it to a specific value, here is "0". Reset for the GaugeVec is to clear the data.  They are quite different.

MartinBasti:
> > Another point why GaugeVec doesn't help: https://paste.centos.org/view/dd247bcb
> > If I replace `GaugeVec` with `Gauge` and `reset` with `set`, it works.
> 
> `set(0)` is to set it to a specific value, here is "0". Reset for the GaugeVec is to clear the data. They are quite d

MartinBasti:
> > Another point why GaugeVec doesn't help: https://paste.centos.org/view/dd247bcb
> > If I replace `GaugeVec` with `Gauge` and `reset` with `set`, it works.
> 
> `set(0)` is to set it to a specific value, here is "0". Reset for the GaugeVec is to clear the data. They are quite different.

Reset just deletes keys from vector
https://github.com/prometheus/client_golang/blob/51d24f86807c24ce866355d0c82258420f571c17/prometheus/vec.go#L95
which are recreated here on `WithLabel` call
https://github.com/prometheus/client_golang/blob/51d24f86807c24ce866355d0c82258420f571c17/prometheus/vec.go#L504
and thus initialized to 0 as standard Gauge metric

it's like set(0) just with extra steps

jinqi7:
> I disagree with change from Gauge to GaugeVec, I still cannot understand value behind it.
> 
> Gauge is correctly initialized to 0 after creation. I noticed negative value in tests, but I think it comes from the code and frankly I quite don't understand all those initialization and new declarations of metrics in tests files, I'm not even sure what is being tested there.
> 
> I tried it also with GaugeVec, and it was not 0 in test without `.Reset()`, you can do the same with `Gauge.Set(0)` and IMO it fixes nothing than else unittests anyway.
> 
> Following test code works, gauge was initialized to 0, and successfully increased
> 
> ```
> +       It("Test gauge", func() {
> +               TestGauge := prometheus.NewGauge(
> +                       prometheus.GaugeOpts{
> +                               Name: "test_gauge",
> +                               Help: "Test",
> +                       },
> +               )
> +               Expect(testutil.ToFloat64(TestGauge)).To(Equal(float64(0)))
> +               TestGauge.Inc()
> +               Expect(testutil.ToFloat64(TestGauge)).To(Equal(float64(1)))
> +       })
>  })
> ```

From my understanding, GaugeVec has more advantage than Gauge. Although we haven't begin to use Label, it gives us the ability to extend.  It has more convenient functions,  such as Reset().  If we want to add more metrics in the future, what shall we select between "Gauge" and "GaugeVec"?  I am tending to use "GaugeVec".  If we change to use GaugeVec in the future, we need to change the data present for it sooner or later.

jinqi7:
How do you think of the other two changes except the GaugeVec ?  @dirgim @MartinBasti

dirgim:
> How do you think of the other two changes except the GaugeVec ? @dirgim @MartinBasti

The other two changes look good to me.

MartinBasti:
> > I disagree with change from Gauge to GaugeVec, I still cannot understand value behind it.
> > Gauge is correctly initialized to 0 after creation. I noticed negative value in tests, but I think it comes from the code and frankly I quite don't understand all those initialization and new declarations of metrics in tests files, I'm not even sure what is being tested there.
> > I tried it also with GaugeVec, and it was not 0 in test without `.Reset()`, you can do the same with `Gauge.Set(0)` and IMO it fixes nothing than else unittests anyway.
> > Following test code works, gauge was initialized to 0, and successfully increased
> > ```
> > +       It("Test gauge", func() {
> > +               TestGauge := prometheus.NewGauge(
> > +                       prometheus.GaugeOpts{
> > +                               Name: "test_gauge",
> > +                               Help: "Test",
> > +                       },
> > +               )
> > +               Expect(testutil.ToFloat64(TestGauge)).To(Equal(float64(0)))
> > +               TestGauge.Inc()
> > +               Expect(testutil.ToFloat64(TestGauge)).To(Equal(float64(1)))
> > +       })
> >  })
> > ```
> 
> From my understanding, GaugeVec has more advantage than Gauge. Although we haven't begin to use Label, it gives us the ability to extend. It has more convenient functions, such as Reset(). If we want to add more metrics in the future, what shall we select between "Gauge" and "GaugeVec"? I am tending to use "GaugeVec". If we change to use GaugeVec in the future, we need to change the data present for it sooner or later.

We are back to the original point, that the gaugeVec change is not a fix for this issue and shouldn't be included.

Other changes looks good to me

jinqi7:
> > From my understanding, GaugeVec has more advantage than Gauge. Although we haven't begin to use Label, it gives us the ability to extend. It has more convenient functions, such as Reset(). If we want to add more metrics in the future, what shall we select between "Gauge" and "GaugeVec"? I am tending to use "GaugeVec". If we change to use GaugeVec in the future, we need to change the data present for it sooner or later.
> 
> We are back to the original point, that the gaugeVec change is not a fix for this issue and shouldn't be included.
> 
> Other changes looks good to me

I seperated it from the current PR and will create a task to follow up it later.

jinqi7:
> > How do you think of the other two changes except the GaugeVec ? @dirgim @MartinBasti
> 
> The other two changes look good to me.

Thanks. I updated the current PR and  will leave the GaugeVec part to other task.

gitops/snapshot.go
```diff
@@ -268,6 +268,24 @@ func IsSnapshotValid(snapshot *applicationapiv1alpha1.Snapshot) bool {
 	return false
 }
 
+// IsSnapshotStatusConditionSet checks if the condition with the conditionType in the status of Snapshot has been marked as the conditionStatus and reason.
+func IsSnapshotStatusConditionSet(snapshot *applicationapiv1alpha1.Snapshot, conditionType string, conditionStatus metav1.ConditionStatus, reason string) bool {
+	condition := meta.FindStatusCondition(snapshot.Status.Conditions, conditionType)
+	if condition == nil && conditionType == AppStudioTestSuceededCondition {
+		condition = meta.FindStatusCondition(snapshot.Status.Conditions, LegacyTestSuceededCondition)
+	}
```
MartinBasti, line 276: Do we miss unittest coverage in lines below?

jinqi7: It's added. Thanks

MartinBasti: But I still see codecov warnings about missing coverage

MartinBasti: https://app.codecov.io/gh/redhat-appstudio/integration-service/pull/263  patch coverage is just 30%

MartinBasti: minimum is 70% desired 80%, there is a huge code block uncovered

jinqi7: The code in controllers/integrationpipeline/integrationpipeline_adapter.go#L129-L141 is in the end of the function  EnsureSnapshotPassedAllTests(), in the middle of the function, some mocked interactions are needed. Then it's hard to touch the end of the function. So, it's not covered.

MartinBasti: if the rest of the team waives coverage level, it's fine by me then

jinqi7: It's 78.26% now

MartinBasti (approved):
Thank you!

gitops/snapshot.go
dirgim: Great work @jinqi7! I think we can conclude this discussion and move forward.

dirgim (approved):
LGTM!

PR #273, opened 2023-08-11
feat(STONEINTG-488): react to application component deletion

Josh-Everett:
Creation of component controller to watch for component deletion events. When a component is deleted, IS will create a new snapshot to reflect those changes as well as update snapshotenvironmentbindings.

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -50,6 +50,12 @@ rules:
 - apiGroups:
   - appstudio.redhat.com
   resources:
+  - components/finalizers
+  verbs:
+  - update
+- apiGroups:
+  - appstudio.redhat.com
+  resources:
   - components/status
   verbs:
   - get
--- a/controllers/component/component_adapter.go
+++ b/controllers/component/component_adapter.go
@@ -1,0 +1,132 @@
+/*
+Copyright 2023.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package component
+
+import (
+	"context"
+
+	applicationapiv1alpha1 "github.com/redhat-appstudio/application-api/api/v1alpha1"
+	"github.com/redhat-appstudio/integration-service/gitops"
+	h "github.com/redhat-appstudio/integration-service/helpers"
+	"github.com/redhat-appstudio/integration-service/loader"
+	"github.com/redhat-appstudio/integration-service/metrics"
+	"github.com/redhat-appstudio/operator-toolkit/controller"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// Adapter holds the objects needed to reconcile a integration PipelineRun.
+type Adapter struct {
+	component   *applicationapiv1alpha1.Component
+	application *applicationapiv1alpha1.Application
+	loader      loader.ObjectLoader
+	logger      h.IntegrationLogger
+	client      client.Client
+	context     context.Context
+}
+
+// NewAdapter creates and returns an Adapter instance.
+func NewAdapter(component *applicationapiv1alpha1.Component, application *applicationapiv1alpha1.Application, logger h.IntegrationLogger, loader loader.ObjectLoader, client client.Client,
+	context context.Context) *Adapter {
+	return &Adapter{
+		component:   component,
+		application: application,
+		logger:      logger,
+		loader:      loader,
+		client:      client,
+		context:     context,
+	}
+}
+
+// EnsureComponentIsCleanedUp is an operation that will ensure components
+// marked for deletion have a snapshot created without said component
+func (a *Adapter) EnsureComponentIsCleanedUp() (controller.OperationResult, error) {
+	if !hasComponentBeenDeleted(a.component) {
+		return controller.ContinueProcessing()
+	}
+
+	applicationComponents, err := a.loader.GetAllApplicationComponents(a.client, a.context, a.application)
+	if err != nil {
+		a.logger.Error(err, "Failed to load application components")
+		return controller.RequeueWithError(err)
+	}
+
+	var snapshotComponents []applicationapiv1alpha1.SnapshotComponent
+
+	for _, individualComponent := range *applicationComponents {
+		component := individualComponent
+		if a.component.Name != component.Name {
+			containerImage := component.Spec.ContainerImage
+			componentSource := gitops.GetComponentSourceFromComponent(&component)
+			snapshotComponents = append(snapshotComponents, applicationapiv1alpha1.SnapshotComponent{
+				Name:           component.Name,
+				ContainerImage: containerImage,
+				Source:         *componentSource,
+			})
+		}
+	}
+
+	if len(snapshotComponents) == 0 {
+		a.logger.Info("Application has no available snapshot components for snapshot creation")
+		return controller.StopProcessing()
+	}
+
+	_, err = a.createUpdatedSnapshot(&snapshotComponents)
+	if err != nil {
+		a.logger.Error(err, "Failed to create new snapshot after component deletion")
+		return controller.RequeueWithError(err)
+	}
+
+	return controller.ContinueProcessing()
+}
+
+// createUpdatedSnapshot prepares a Snapshot for a given application and component(s).
+// In case the Snapshot can't be created, an error will be returned.
+func (a *Adapter) createUpdatedSnapshot(snapshotComponents *[]applicationapiv1alpha1.SnapshotComponent) (*applicationapiv1alpha1.Snapshot, error) {
+	snapshot := gitops.NewSnapshot(a.application, snapshotComponents)
+	if snapshot.Labels == nil {
+		snapshot.Labels = map[string]string{}
+	}
+	snapshotType := gitops.SnapshotCompositeType
+	if len(*snapshotComponents) == 1 {
+		snapshotType = gitops.SnapshotComponentType
+	}
+	snapshot.Labels[gitops.SnapshotTypeLabel] = snapshotType
+
+	err := ctrl.SetControllerReference(a.application, snapshot, a.client.Scheme())
+	if err != nil {
+		a.logger.Error(err, "Failed to set controller refrence")
+		return nil, err
+	}
+
+	err = a.client.Create(a.context, snapshot)
+	if err != nil {
+		a.logger.Error(err, "Failed to create new snapshot on client")
+		return nil, err
+	}
+
+	go metrics.RegisterNewSnapshot()
+	return snapshot, nil
+}
+
+func hasComponentBeenDeleted(object client.Object) bool {
+
+	if comp, ok := object.(*applicationapiv1alpha1.Component); ok {
+		return !comp.ObjectMeta.DeletionTimestamp.IsZero()
+	}
+	return false
+}
--- a/controllers/component/component_adapter_test.go
+++ b/controllers/component/component_adapter_test.go
@@ -1,0 +1,145 @@
+package component
+
+import (
+	"bytes"
+	"reflect"
+	"time"
+
+	"github.com/tonglil/buflogr"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	applicationapiv1alpha1 "github.com/redhat-appstudio/application-api/api/v1alpha1"
+	"github.com/redhat-appstudio/integration-service/loader"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/redhat-appstudio/integration-service/helpers"
+	"k8s.io/apimachinery/pkg/api/errors"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var _ = Describe("Component Adapter", Ordered, func() {
+	var (
+		adapter *Adapter
+		logger  helpers.IntegrationLogger
+
+		hasApp   *applicationapiv1alpha1.Application
+		hasComp  *applicationapiv1alpha1.Component
+		hasComp2 *applicationapiv1alpha1.Component
+	)
+	const (
+		SampleCommit             = "a2ba645d50e471d5f084b"
+		SampleRepoLink           = "https://github.com/devfile-samples/devfile-sample-java-springboot-basic"
+		sample_revision          = "random-value"
+		SampleDigest             = "sha256:841328df1b9f8c4087adbdcfec6cc99ac8308805dea83f6d415d6fb8d40227c1"
+		SampleImageWithoutDigest = "quay.io/redhat-appstudio/sample-image"
+		SampleImage              = SampleImageWithoutDigest + "@" + SampleDigest
+	)
+
+	BeforeAll(func() {
+
+		hasApp = &applicationapiv1alpha1.Application{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "application-sample",
+				Namespace: "default",
+			},
+			Spec: applicationapiv1alpha1.ApplicationSpec{
+				DisplayName: "application-sample",
+				Description: "This is an example application",
+			},
+		}
+		Expect(k8sClient.Create(ctx, hasApp)).Should(Succeed())
+
+		hasComp = &applicationapiv1alpha1.Component{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "component-sample",
+				Namespace: "default",
+			},
+			Spec: applicationapiv1alpha1.ComponentSpec{
+				ComponentName:  "component-sample-2",
+				Application:    hasApp.Name,
+				ContainerImage: SampleImage,
+				Source: applicationapiv1alpha1.ComponentSource{
+					ComponentSourceUnion: applicationapiv1alpha1.ComponentSourceUnion{
+						GitSource: &applicationapiv1alpha1.GitSource{
+							URL:      SampleRepoLink,
+							Revision: SampleCommit,
+						},
+					},
+				},
+			},
+			Status: applicationapiv1alpha1.ComponentStatus{
+				LastBuiltCommit: "",
+			},
+		}
+		Expect(k8sClient.Create(ctx, hasComp)).Should(Succeed())
+
+		hasComp2 = &applicationapiv1alpha1.Component{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "component-second-sample",
+				Namespace: "default",
+			},
+			Spec: applicationapiv1alpha1.ComponentSpec{
+				ComponentName:  "component-second-sample",
+				Application:    "application-sample",
+				ContainerImage: SampleImage,
+				Source: applicationapiv1alpha1.ComponentSource{
+					ComponentSourceUnion: applicationapiv1alpha1.ComponentSourceUnion{
+						GitSource: &applicationapiv1alpha1.GitSource{
+							URL: SampleRepoLink,
+						},
+					},
+				},
+			},
+		}
+		Expect(k8sClient.Create(ctx, hasComp2)).Should(Succeed())
+	})
+
+	AfterAll(func() {
+		err := k8sClient.Delete(ctx, hasApp)
+		Expect(err == nil || errors.IsNotFound(err)).To(BeTrue())
+		err = k8sClient.Delete(ctx, hasComp)
+		Expect(err == nil || errors.IsNotFound(err)).To(BeTrue())
+		err = k8sClient.Delete(ctx, hasComp2)
+		Expect(err == nil || errors.IsNotFound(err)).To(BeTrue())
+	})
+
+	It("can create a new Adapter instance", func() {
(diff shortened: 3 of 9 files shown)


controllers/component/component_adapter.go
```diff
@@ -0,0 +1,138 @@
...
+		application: application,
+		logger:      logger,
+		loader:      loader,
+		client:      client,
+		context:     context,
+		status:      status.NewAdapter(logger.Logger, client),
+	}
+}
+
+// EnsureSnapshotIsFresh is an operation that will ensure that any snapshots associated with components
```
dirgim, line 60: I don't think we want to update all existing Snapshot when a component is deleted, et least not as part of this ticket.
I think it's enough to generate a new Snapshot when a Component is deleted. This Snapshot will then be automatically processed by the rest of the Integration service controllers (which will be testing, deploying, and releasing it), so there should be no need for further actions such as creating SEBs.

Josh-Everett: Ack, makes sense. I am trying to better understand which situations where application components != snapshot components. If it's only a situation where a build has been added/deleted/updated to an application and the snapshots haven't updated to reflect this. Could a user be running multiple snapshots for an application to test differing component configurations?

dirgim: I'm not aware of that being an option in the current workflow. An alternative would be to have multiple applications with different component configurations, but that would also be out of scope for this ticket.

controllers/component/component_adapter.go
```diff
@@ -0,0 +1,145 @@
...
+// marked for deletion are recreated without said component
+func (a *Adapter) EnsureSnapshotIsFresh() (controller.OperationResult, error) {
+	if !HasComponentBeenDeleted(a.component) {
+		return controller.ContinueProcessing()
+	}
+	applicationSnapshots, err := a.loader.GetAllSnapshots(a.client, a.context, a.application)
```
dirgim, line 67: Hmm, I don't think we need to check existing Snapshots, I think it would be enough to get all existing components and go through that list to make sure that the deleted component is not there before creating a new Snapshot.

controllers/component/component_adapter.go
```diff
@@ -0,0 +1,114 @@
...
+	if len(*remainingComponents) == 0 {
+		return nil, fmt.Errorf("failed to prepare snapshot due to missing valid digest in containerImage for all components of application")
+	}
+
+	snapshot, err := gitops.PrepareSnapshot(a.client, a.context, a.application, remainingComponents, nil, "", nil)
+	if err != nil {
+		return nil, err
+	}
+
+	go metrics.RegisterNewSnapshot()
```
dirgim, line 103: Before registering a new Snapshot, we'll also need to do two things:
- Add the type label so the Snapshot is identified as a `composite` one (like. e.g. [here](https://github.com/redhat-appstudio/integration-service/blob/272ccefcfc31504768a5e28780ee2fdb144b51e3/controllers/integrationpipeline/integrationpipeline_adapter.go#L345-L348))
- Create the Snapshot on the cluster (like e.g. [here](https://github.com/redhat-appstudio/integration-service/blob/272ccefcfc31504768a5e28780ee2fdb144b51e3/controllers/integrationpipeline/integrationpipeline_adapter.go#L389-L392))

controllers/component/component_adapter_test.go
```diff
@@ -0,0 +1,147 @@
...
+				ContextKey: loader.ComponentContextKey,
+				Resource:   hasComp2,
+			},
+			{
+				ContextKey: loader.ApplicationComponentsContextKey,
+				Resource:   []applicationapiv1alpha1.Component{*hasComp, *hasComp2},
+			},
+		})
+
+		applicationSnapshots, err := adapter.loader.GetAllSnapshots(k8sClient, adapter.context, hasApp)
```
dirgim, line 131: Since the `GetAllSnapshots()` function requires cache which isn't available, getting a list of Snapshots in the namespace with a `k8sClient.List()` will be needed here and also after the `EnsureComponentIsCleanedUp()` call lower in the code.
In the second case, since the control plane could be slow, it's best to have an `Eventually` block that waits until the list of `Snapshots` is populated, similar to how it's done [here](https://github.com/redhat-appstudio/integration-service/blob/272ccefcfc31504768a5e28780ee2fdb144b51e3/controllers/binding/snapshotenvironmentbinding_adapter_test.go#L372-L375) for `integrationTestScenarios`.

controllers/component/component_adapter_test.go
```diff
@@ -0,0 +1,147 @@
...
+		Expect(*applicationSnapshots).To(BeNil())
+		now := metav1.NewTime(metav1.Now().Add(time.Second * 1))
+		hasComp.SetDeletionTimestamp(&now)
+
+		Eventually(func() bool {
+			result, err := adapter.EnsureComponentIsCleanedUp()
+			return !result.CancelRequest && err == nil
+		}, time.Second*20).Should(BeTrue())
+
+		applicationSnapshots, err = adapter.loader.GetAllSnapshots(k8sClient, adapter.context, hasApp)
```
dirgim, line 142: In this case case, since the control plane could be slow, it's best to have an Eventually block that waits until the list of Snapshots is populated, similar to how it's done [here](https://github.com/redhat-appstudio/integration-service/blob/272ccefcfc31504768a5e28780ee2fdb144b51e3/controllers/binding/snapshotenvironmentbinding_adapter_test.go#L372-L375) for integrationTestScenarios.

controllers/component/component_adapter.go
```diff
@@ -0,0 +1,160 @@
...
+		return controller.RequeueWithError(err)
+	}
+
+	// TODO broken here
+	//var updatedComponents = []applicationapiv1alpha1.Component{}
+	comp := *(a.component)
+
+	/* 	for i, individualComponent := range *applicationComponents {
+	   		if comp.ObjectMeta.UID == individualComponent.ObjectMeta.UID {
+	   			//if reflect.DeepEqual(comp, individualComponent) {
```
dirgim, line 79: I don't think that using `DeepEqual` here is needed - I think that checking the `name` should be enough for this kind of comparison.

dirgim: Disregard, the approach that's below works even better.

Josh-Everett: @dirgim yeah the deep equal here was causing some grief. I had seen it[ used here](https://github.com/redhat-appstudio/integration-service/blob/d2f5e27de0f49866fe8d61036ad27c22a3496a8d/gitops/snapshot.go#L389) to compare components. I used UID for this patch because I was wondering if there might be a situation where a deleted component might share the name of another vaild component within the application. In the case that that a component has been added to an application to replace the component with a different version/image.

controllers/component/component_controller.go
```diff
@@ -0,0 +1,116 @@
...
+		Log:    logger.WithName("integration pipeline"),
+		Scheme: scheme,
+	}
+}
+
+//+kubebuilder:rbac:groups=appstudio.redhat.com,resources=components,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=appstudio.redhat.com,resources=components/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=appstudio.redhat.com,resources=components/finalizers,verbs=update
+//+kubebuilder:rbac:groups=core,resources=configmaps,verbs=get;list;watch
+//+kubebuilder:rbac:groups=core,resources=secrets,verbs=get;list;watch
```
dheerajodha, line 56: Do we need to also add the entry for the apiGroups: `pipelinesascode.tekton.dev`, like [this](https://github.com/redhat-appstudio/integration-service/blob/fd8631a6931762fdd0863302579d0fb5f17c41db/controllers/buildpipeline/buildpipeline_controller.go#L62)? Because the [check fails on this](https://github.com/redhat-appstudio/integration-service/actions/runs/6153305093/job/16696963235?pr=273#step:13:61), so maybe it would fix the error...

controllers/component/component_controller.go
```diff
@@ -0,0 +1,116 @@
...
+	if err != nil {
+		logger.Error(err, "Failed to get component for", "req", req.NamespacedName)
+		if errors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+
+		return ctrl.Result{}, err
+	}
+
+	application := &applicationapiv1alpha1.Application{}
```
MartinBasti, line 75:
```suggestion
	var application *applicationapiv1alpha1.Application
```

controllers/component/component_adapter.go
```diff
@@ -0,0 +1,129 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	go metrics.RegisterNewSnapshot()
+
+	return snapshot, nil
+}
+
+func HasComponentBeenDeleted(object client.Object) bool {
```
MartinBasti, line 123: first letter should be lowercase if it's only internal function

controllers/component/component_adapter.go
```diff
@@ -0,0 +1,129 @@
...
+	if err != nil {
+		return controller.RequeueWithError(err)
+	}
+
+	return controller.ContinueProcessing()
+}
+
+// CreateUpdatedSnapshot prepares a Snapshot for a given application and component(s).
+// In case the Snapshot can't be created, an error will be returned.
+func (a *Adapter) CreateUpdatedSnapshot(adapterClient client.Client, application *applicationapiv1alpha1.Application, snapshotComponents *[]applicationapiv1alpha1.SnapshotComponent) (*applicationapiv1alpha1.Snapshot, error) {
```
MartinBasti, line 102: first letter should be lowercase if it's only internal function

controllers/component/component_adapter.go
```diff
@@ -0,0 +1,129 @@
...
+				ContainerImage: containerImage,
+				Source:         *componentSource,
+			})
+		}
+	}
+
+	snapshot, err := a.CreateUpdatedSnapshot(a.client, a.application, &snapshotComponents)
+	if err != nil {
+		a.logger.Error(err, "Failed to create new snapshot after component deletion")
+		return controller.RequeueWithError(err)
```
MartinBasti, line 89: if user removes last component but keeps application, will we here cycle here in infinite requeueing?

MartinBasti: I think we should just stop reconciliation without error when snapshotComponents are empty

controllers/component/component_adapter.go
```diff
@@ -0,0 +1,129 @@
...
+// EnsureComponentIsCleanedUp is an operation that will ensure components
+// marked for deletion have a snapshot created without said component
+func (a *Adapter) EnsureComponentIsCleanedUp() (controller.OperationResult, error) {
+	if !HasComponentBeenDeleted(a.component) {
+		return controller.ContinueProcessing()
+	}
+
+	applicationComponents, err := a.loader.GetAllApplicationComponents(a.client, a.context, a.application)
+	if err != nil {
+		return controller.RequeueWithError(err)
```
MartinBasti, line 68: we miss error chainings everywhere, there won't be enough of context for debugging where exactly it failed

controllers/component/component_adapter.go
```diff
@@ -0,0 +1,131 @@
...
+
+// Adapter holds the objects needed to reconcile a integration PipelineRun.
+type Adapter struct {
+	component   *applicationapiv1alpha1.Component
+	application *applicationapiv1alpha1.Application
+	loader      loader.ObjectLoader
+	logger      h.IntegrationLogger
+	client      client.Client
+	context     context.Context
+	status      status.Status
```
hongweiliu17, line 41: I don't think we need this `status` in this adapter.

controllers/component/component_controller.go
hongweiliu17: I think run `make manifests` before committing the PR can fix that error

controllers/component/component_controller.go
```diff
@@ -0,0 +1,117 @@
...
+func SetupController(manager ctrl.Manager, log *logr.Logger) error {
+	return setupControllerWithManager(manager, NewComponentReconciler(manager.GetClient(), log, manager.GetScheme()))
+
+}
+
+// setupControllerWithManager sets up the controller with the Manager which monitors new PipelineRuns and filters
+// out status updates.
+func setupControllerWithManager(manager ctrl.Manager, controller *Reconciler) error {
+	return ctrl.NewControllerManagedBy(manager).
+		For(&tektonv1beta1.PipelineRun{}).
```
hongweiliu17, line 113:
```suggestion
		For(&applicationapiv1alpha1.Component{}).
```

controllers/component/component_controller.go
```diff
@@ -0,0 +1,117 @@
...
+	client.Client
+	Log    logr.Logger
+	Scheme *runtime.Scheme
+}
+
+// NewComponentReconciler creates and returns a Reconciler.
+func NewComponentReconciler(client client.Client, logger *logr.Logger, scheme *runtime.Scheme) *Reconciler {
+	return &Reconciler{
+		Client: client,
+		Log:    logger.WithName("integration pipeline"),
```
MartinBasti, line 47:
```suggestion
		Log:    logger.WithName("component controller"),
```

controllers/component/component_controller.go
Josh-Everett: RBAC failures have been resolved in [1e30d88](https://github.com/redhat-appstudio/integration-service/commit/1e30d880842763338ce6686ac43857701adb2f56)
@hongweiliu17 that is good to know, I will take note of that I ended up doing it manually.

controllers/component/component_adapter.go
Josh-Everett: Should be fixed in [ 0967a8c ](https://github.com/redhat-appstudio/integration-service/commit/0967a8cbb898c2e1daa182ca3cf4f2afb721e885) :  [here](https://github.com/redhat-appstudio/integration-service/commit/0967a8cbb898c2e1daa182ca3cf4f2afb721e885#diff-0a268eab99b22ffc9ca9a74ef2752ed7dd05a822b3da8fe160ffee3a2ad1035eR105)

controllers/component/component_adapter.go
Josh-Everett: I think error chaining should be correctly implemented by [1e30d88](https://github.com/redhat-appstudio/integration-service/commit/1e30d880842763338ce6686ac43857701adb2f56) like in this spot  [here](https://github.com/redhat-appstudio/integration-service/commit/1e30d880842763338ce6686ac43857701adb2f56#diff-0a268eab99b22ffc9ca9a74ef2752ed7dd05a822b3da8fe160ffee3a2ad1035eR64)

controllers/component/component_controller.go
Josh-Everett: artifacts of my "borrowing" from a pipeline controller have been removed [here](https://github.com/redhat-appstudio/integration-service/commit/0967a8cbb898c2e1daa182ca3cf4f2afb721e885)

controllers/component/component_controller.go
```diff
@@ -0,0 +1,117 @@
...
+	EnsureSnapshotIsFresh() (controller.OperationResult, error)
+}
+
+// SetupController creates a new Component controller and adds it to the Manager.
+func SetupController(manager ctrl.Manager, log *logr.Logger) error {
+	return setupControllerWithManager(manager, NewComponentReconciler(manager.GetClient(), log, manager.GetScheme()))
+
+}
+
+// setupControllerWithManager sets up the controller with the Manager which monitors new PipelineRuns and filters
```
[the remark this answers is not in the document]
Josh-Everett: fix updated in https://github.com/redhat-appstudio/integration-service/commit/0967a8cbb898c2e1daa182ca3cf4f2afb721e885

dirgim:
@Josh-Everett can you create a simple diagram ([guide](https://github.com/redhat-appstudio/integration-service/blob/main/docs/README.md)) for the new controller?

Josh-Everett:
Diagram added [here](https://github.com/redhat-appstudio/integration-service/blob/1950be13a1d3eb3aa1e198745275f07c31003b24/docs/component_controller.md)

controllers/component/component_adapter.go
```diff
@@ -0,0 +1,128 @@
...
+		return controller.StopProcessing()
+	}
+
+	snapshot, err := a.createUpdatedSnapshot(a.client, a.application, &snapshotComponents)
+	if err != nil {
+		a.logger.Error(err, "Failed to create new snapshot after component deletion")
+		return controller.RequeueWithError(err)
+	}
+
+	err = ctrl.SetControllerReference(a.application, snapshot, a.client.Scheme())
```
dirgim, line 94: The controller reference needs to be applied on the cluster in order to take effect.
I think it would be best to set it before creating the Snapshot ([here](https://github.com/redhat-appstudio/integration-service/pull/273/files#diff-0a268eab99b22ffc9ca9a74ef2752ed7dd05a822b3da8fe160ffee3a2ad1035eR112)).

controllers/component/component_controller.go
```diff
@@ -0,0 +1,116 @@
...
+// NewComponentReconciler creates and returns a Reconciler.
+func NewComponentReconciler(client client.Client, logger *logr.Logger, scheme *runtime.Scheme) *Reconciler {
+	return &Reconciler{
+		Client: client,
+		Log:    logger.WithName("integration pipeline"),
+		Scheme: scheme,
+	}
+}
+
+//+kubebuilder:rbac:groups=appstudio.redhat.com,resources=components,verbs=get;list;watch;create;update;patch;delete
```
dirgim, line 51: I don't think that we require `create` or `delete` permissions for components, so it would be probably best not to have those permissions according to the principle of least privilege.

controllers/component/component_controller.go
dirgim: On a different topic, I think that we don't need access to secrets for this controller, so this line might not be needed here.

config/rbac/role.yaml
```diff
@@ -42,11 +42,19 @@ rules:
...
+  - delete
   - get
   - list
   - patch
   - update
   - watch
+- apiGroups:
+  - appstudio.redhat.com
+  resources:
+  - components/finalizers
```
14rcole, line 55: What does the slash mean here?  Does it means "This applies to components and finalizers for components?"

controllers/component/component_controller.go
hongweiliu17: Do we need this `configmaps` and `pipelinesascode.tekton.dev`? cc @dirgim  I feel not, but I am not sure.

dirgim: Based on my understanding of the code, I think that they are not needed either, but it would probably be best to confirm by testing.

config/rbac/role.yaml
dirgim: It means finalizers for components, as that's a subresource.

controllers/component/component_adapter.go
```diff
@@ -0,0 +1,160 @@
...
+	return controller.ContinueProcessing()
+}
+
+func (a *Adapter) failStaleSnapshots() error {
+	allSnapshots, err := a.loader.GetAllSnapshots(a.client, a.context, a.application)
+	if err != nil {
+		a.logger.Error(err, "Failed to load all snapshots")
+		return err
+	}
+	for _, snapshot := range *allSnapshots {
```
dirgim, line 109: There is a `G601 (CWE-118): Implicit memory aliasing in for loop.` Go Security [CI failure](https://github.com/redhat-appstudio/integration-service/actions/runs/6307716804/job/17124768810?pr=273) for this loop.
It can be resolved pretty simply by adding an assignment for the `snapshot` variable under this line like e.g. [here](https://github.com/redhat-appstudio/integration-service/blob/02c829368d912e684b82dbab204b34f56e576fba/controllers/snapshot/snapshot_adapter.go#L106).

dirgim:
Running `make manifests` should resolve the current Check sources [CI failure](https://github.com/redhat-appstudio/integration-service/actions/runs/6307716797/job/17124769059?pr=273).

14rcole (approved):

controllers/component/component_adapter.go
```diff
@@ -0,0 +1,161 @@
...
+			componentSource := gitops.GetComponentSourceFromComponent(&component)
+			snapshotComponents = append(snapshotComponents, applicationapiv1alpha1.SnapshotComponent{
+				Name:           component.Name,
+				ContainerImage: containerImage,
+				Source:         *componentSource,
+			})
+		}
+	}
+
+	err = a.failStaleSnapshots()
```
dirgim, line 83: I've been giving this some thought and I think that marking all Snapshots that contain the deleted component as failed testing would introduce a number of problems, while not providing a large benefit.

The downsides to doing this are:

- Updating the status of all (or majority of) previous Snapshots would be very resource intensive (there could be potentially up to a thousand of old Snapshots in the user's workspace).
- The update events would/could trigger reconciliations in other services, compounding the resource issues
- Marking stale Snapshots as failed would send a wrong message to the end-users since it's not the tests that failed, but the validity of the Snapshot itself

The benefit of marking them as failed is also lessened in light of newer changes (such as #340) that prevent old Snapshots from being promoted without being tested again - which would immediately reveal that their component composition is out-of-date anyways.

Josh-Everett: ack this makes sense, I hadn't considered some of those potential issues that could arrive from iterating over all that historical data. I removed the code involved in failing snapshots, but @dirgim can you confirm that would extend to snapshots currently in testing as well?

dirgim: That's a good question - when a Snapshot finishes testing, the integrationpipeline adapter first checks if the components in the Snapshot match the current state of the global candidate list (GCL) ([here](https://github.com/redhat-appstudio/integration-service/blob/92c8f944f0104ba4d687b89bd458008cd341bcb9/controllers/integrationpipeline/integrationpipeline_adapter.go#L117-L120)).
Since the Snapshot that just ended testing would be missing a component, it should be marked as invalid and the integrationpipeline adapter should check if there's already a fresher Snapshot that matches the GCL. In our scenario here, the component_adapter would have already created that fresh Snapshot so the logic would stop there.

dirgim (approved):
LGTM

controllers/component/component_adapter.go
dirgim: Resolving this convo so it doesn't block the PR from going forward.

PR #283, opened 2023-08-23
feat(STONEINTG-523): enum for integration test status

MartinBasti:
This commit provides base data structure for the keeping status of tests.

Type supports string and JSON interfaces to convert between numeric and textual representation.


gitops/snapshot.go
```diff
@@ -117,6 +117,25 @@ const (
 	AppStudioIntegrationStatusFinished = "Finished"
 )
 
+// IntegrationTestScenario test runs status
+type TestStatus int
```
MartinBasti, line 121: Feel free to suggest better name

gitops/snapshot.go
```diff
@@ -117,6 +117,25 @@ const (
...
+
+//go:generate enumer -type=TestStatus -linecomment -json
```
hongweiliu17, line 123: The usage looks great. I didn't know it before, let me search a doc to learn. Thanks

gitops/snapshot.go
MartinBasti: updated

MartinBasti:
Coverage is low because of generated code, can I wave it?

dirgim:
> Coverage is low because of generated code, can I wave it?

I think it would be best to add the generated file to the ignore list in [codecov.yml](https://github.com/redhat-appstudio/integration-service/blob/main/codecov.yml).

hongweiliu17 (approved):
LGTM

dirgim (approved):
LGTM

PR #289, opened 2023-08-24
fix(unitest): integrationpipline has wrong type

MartinBasti:
Integration PLR has a wrong type label in the test suite.


MartinBasti:
Not sure what this PLR object is for, test works the same with or without correct label

dirgim:
> Not sure what this PLR object is for, test works the same with or without correct label

I think that since the PLRs are filtered by the predicate which checks this label, by the time it gets handled by the controller, it doesn't get checked or used anymore, hence why the unit tests work. 
We definitely want the correct label to be there, of course.

dirgim (approved):
LGTM

hongweiliu17 (approved):
LGTM

PR #294, opened 2023-08-28
feat(STONEINTG-528): refactor annotation/label to use operator-toolkit

jinqi7:


controllers/buildpipeline/buildpipeline_adapter.go
```diff
@@ -262,7 +263,11 @@ func (a *Adapter) getSucceededBuildPipelineRunsForComponent(component *applicati
 
 func (a *Adapter) annotateBuildPipelineRunWithSnapshot(pipelineRun *tektonv1beta1.PipelineRun, snapshot *applicationapiv1alpha1.Snapshot) (*tektonv1beta1.PipelineRun, error) {
 	patch := client.MergeFrom(pipelineRun.DeepCopy())
-	h.AddAnnotation(&pipelineRun.ObjectMeta, tekton.SnapshotNameLabel, snapshot.Name)
+	temp_annotations := map[string]string{}
```
dirgim, line 266: nit: `temp_annotations` would probably be easier to understand as something like `newAnnotations`.

controllers/snapshot/snapshot_adapter.go
```diff
@@ -523,7 +524,9 @@ func (a *Adapter) createSnapshotEnvironmentBindingForSnapshot(application *appli
 
 	for _, keyAndValue := range optionalLabelKeysAndValues {
 		if v, ok := keyAndValue[gitops.SnapshotTestScenarioLabel]; ok {
-			h.AddLabel(&snapshotEnvironmentBinding.ObjectMeta, gitops.SnapshotTestScenarioLabel, v)
+			temp_labels := map[string]string{}
```
dirgim, line 527: nit: `temp_labels` would probably be easier to understand as something like `newLabels`.

controllers/snapshot/snapshot_adapter_test.go
```diff
@@ -699,9 +738,12 @@ var _ = Describe("Snapshot Adapter", Ordered, func() {
 			components := []applicationapiv1alpha1.Component{
 				*hasComp,
 			}
-			snapshotEnvironmentBinding, err := adapter.createSnapshotEnvironmentBindingForSnapshot(adapter.application, env, hasSnapshot, &components)
+			temp_labels := map[string]string{}
```
dirgim, line 741: nit: `temp_labels` would probably be easier to understand as something like `newLabels`.

controllers/snapshot/snapshot_suite_test.go
```diff
@@ -99,6 +100,13 @@ var _ = BeforeSuite(func() {
 	k8sClient = k8sManager.GetClient()
 	go func() {
 		defer GinkgoRecover()
+		Expect(cache.SetupIntegrationTestScenarioCache(k8sManager)).To(Succeed())
```
dirgim, line 103: I think we want to prevent setting up cache for unit tests because that causes latency issues and works unreliably in gitHub actions (see [STONEINTG-459](https://issues.redhat.com/browse/STONEINTG-459)).

jinqi7: It's added because I always met  below issue for the test case "ensures Release created successfully" although the mocked context is added in the case.
-  [FAILED] Unexpected error:
      <*errors.errorString | 0xc00085c6b0>: 
      Index with name field:spec.snapshot does not exist
      {
          s: "Index with name field:spec.snapshot does not exist",
      }
  occurred

jinqi7: If we prevent setting up cache, I'll not add the test case in the PR.

controllers/snapshot/snapshot_adapter_test.go
jinqi7: Will updated

controllers/snapshot/snapshot_adapter.go
jinqi7: I will update it.

controllers/snapshot/snapshot_suite_test.go
dirgim: Just took a look at what's missing and I think that the `Release` resources would also need to be mocked because of [this call](https://github.com/redhat-appstudio/integration-service/blob/main/controllers/snapshot/snapshot_adapter.go#L401C40-L401C40), so adding something like this to the mocked resources would make it work without cache:
```
       {
		ContextKey: loader.ReleaseContextKey,
		Resource:   &releasev1alpha1.Release{},
	},
```

dirgim: If the "ensures Release created successfully" test case works with the added mocked Release resource I mentioned above, I think it would be good to have it.

controllers/binding/snapshotenvironmentbinding_adapter.go
```diff
@@ -162,9 +163,11 @@ func (a *Adapter) createIntegrationPipelineRunWithEnvironment(application *appli
 		WithIntegrationLabels(integrationTestScenario).
 		WithEnvironmentAndDeploymentTarget(deploymentTarget, environment.Name).
 		AsPipelineRun()
+
 	// copy PipelineRun PAC annotations/labels from snapshot to integration test PipelineRuns
-	h.CopyAnnotationsByPrefix(&snapshot.ObjectMeta, &pipelineRun.ObjectMeta, gitops.PipelinesAsCodePrefix, gitops.PipelinesAsCodePrefix)
-	h.CopyLabelsByPrefix(&snapshot.ObjectMeta, &pipelineRun.ObjectMeta, gitops.PipelinesAsCodePrefix, gitops.PipelinesAsCodePrefix)
+	_ = metadata.CopyAnnotationsByPrefix(&snapshot.ObjectMeta, &pipelineRun.ObjectMeta, gitops.PipelinesAsCodePrefix)
```
jsztuka, line 168: I see that previous function version used 4 inputs and new one uses 3, but I don't se updated function definition, is it planned in idfferent PR?

controllers/buildpipeline/buildpipeline_adapter.go
```diff
@@ -204,12 +205,12 @@ func (a *Adapter) prepareSnapshotForPipelineRun(pipelineRun *tektonv1beta1.Pipel
 
 	// Copy PipelineRun PAC annotations/labels from Build to snapshot.
 	// Modify the prefix so the PaC controller won't react to PipelineRuns generated from the snapshot.
-	h.CopyLabelsByPrefix(&pipelineRun.ObjectMeta, &snapshot.ObjectMeta, "pipelinesascode.tekton.dev", gitops.PipelinesAsCodePrefix)
-	h.CopyAnnotationsByPrefix(&pipelineRun.ObjectMeta, &snapshot.ObjectMeta, "pipelinesascode.tekton.dev", gitops.PipelinesAsCodePrefix)
+	_ = metadata.CopyLabelsWithPrefixReplacement(&pipelineRun.ObjectMeta, &snapshot.ObjectMeta, "pipelinesascode.tekton.dev", gitops.PipelinesAsCodePrefix)
+	_ = metadata.CopyAnnotationsWithPrefixReplacement(&pipelineRun.ObjectMeta, &snapshot.ObjectMeta, "pipelinesascode.tekton.dev", gitops.PipelinesAsCodePrefix)
```
jsztuka, line 209: I don't have such function in main branch and I don't see it defined here. Is it planned in different PR?

controllers/binding/snapshotenvironmentbinding_adapter.go
jinqi7: Currently, we are using functions from [here](https://github.com/redhat-appstudio/operator-toolkit/blob/main/metadata/metadata.go).

controllers/buildpipeline/buildpipeline_adapter.go
dirgim: This function is imported from the [operator-toolkit repo](https://github.com/redhat-appstudio/operator-toolkit/tree/main/metadata), that's why it's not visible in our code.

controllers/snapshot/snapshot_suite_test.go
jinqi7: Yes. It works with the Release resource added in the mocked context as you suggested. Thanks!

dirgim (approved):
LGTM

jsztuka (approved):
lgtm

controllers/buildpipeline/buildpipeline_adapter.go
```diff
@@ -262,7 +263,11 @@ func (a *Adapter) getSucceededBuildPipelineRunsForComponent(component *applicati
 
 func (a *Adapter) annotateBuildPipelineRunWithSnapshot(pipelineRun *tektonv1beta1.PipelineRun, snapshot *applicationapiv1alpha1.Snapshot) (*tektonv1beta1.PipelineRun, error) {
 	patch := client.MergeFrom(pipelineRun.DeepCopy())
-	h.AddAnnotation(&pipelineRun.ObjectMeta, tekton.SnapshotNameLabel, snapshot.Name)
+	newAnnotations := map[string]string{}
+	newAnnotations[tekton.SnapshotNameLabel] = snapshot.Name
+
+	_ = metadata.AddAnnotations(&pipelineRun.ObjectMeta, newAnnotations)
+
```
MartinBasti, line 270: What is the benefit of this change?

Now I have to write this boilerplate code to add single annotation into list, to update it and copy paste it, instead of having a single functions that hides it

MartinBasti: Also could be just
```
newAnnotations := map[string]string{
  tekton.SnapshotNameLabel: snapshot.Name
}
```

PR #297, opened 2023-09-04
fix(STONEINTG_587): check kubernetes manifests is not working

jinqi7:
1. Use ubuntu directly with removing fedora container
2. Update the task to check the whole dir and remove duplicated task

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -20,7 +20,6 @@ jobs:
   go:
     name: Check sources
     runs-on: ubuntu-20.04
-    container: fedora:35
     env:
       OPERATOR_SDK_VERSION: v1.18.0
     steps:
@@ -32,8 +31,6 @@ jobs:
         uses: actions/checkout@v3
         with:
           fetch-depth: 0
-      - name: Download fedora environment dependencies # Would be nice to cache this somehow
-        run: sudo dnf -y install diffutils findutils gcc wget which jq git
       - name: Cache Operator SDK ${{ env.OPERATOR_SDK_VERSION }}
         uses: actions/cache@v3
         id: cache-operator-sdk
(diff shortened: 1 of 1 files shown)


.github/workflows/pr.yaml
```diff
@@ -91,18 +88,9 @@ jobs:
         with:
           version: "2023.1.3"
           install-go: false
-      - name: Check manifests # make generate manifests touches files even for no changes
```
hongweiliu17, line 94: I am wondering why this step is deleted. Can it fix the issue?

jinqi7: This step is not supported when we use "Ubuntu" directly.  When I try to make the "Check kubernetes manifests" task work,  I found it can't work well with using the "fedora:35" image. I am not sure if  the step ever work well before.

jinqi7:
> Can you please express in commit message why and how it fixes the issue? Root cause what was wrong would be also good to have mentioned there.
> 
> I'm not sure if I can consider removal of check as a "fix" without further explanation.

See the comment above.

.github/workflows/pr.yaml
MartinBasti: What does it mean it's not supported? What kind of error did you get?

jinqi7: I didn't save the error.  But from Johnny Berian, that way is not a supposed one.

MartinBasti: Who is Johnny Bieren?

Sorry but I don't understand what's going on here, I filed story to fix `Check kubernetes manifests` but I received:
* removal of fedora container and keeping default ubuntu, without explanation why (probably removal of task to install dependencies is related)
* the other task, which was working on fedora, is suddenly removed, without an explanation why, without showing what is the error
* no explanation in commit message 
* and task is fixed without mentioning what was the root cause

jinqi7: Some idea for the fix is from referring other code with better practice.  I can't explain so much to you.

MartinBasti (approved):

MartinBasti:
approved as git commit was updated with explanation and task has been updated to keep original functionality to validate whole repo

hongweiliu17 (approved):
lgtm

PR #303, opened 2023-09-08
fix(STONEINTG-523): test race condition with timestamps

MartinBasti:
Sometimes test is too quick and the timestamp is the same as before update.
Adding sleep explictily to make sure timestamp differs


dirgim (approved):
LGTM

PR #304, opened 2023-09-11
fix(STONEINTG-524): update NS of pipelines-as-code secret

hongweiliu17:
* update NS of pipelines-as-code secret since the NS has been changed to openshift-pipelines


dheerajodha (approved):
LGTM

PR #305, opened 2023-09-11
feat: additional SLI dashboards in grafana

jencull:

### DIFF
--- a/config/grafana/dashboards/integration-service-dashboard.json
+++ b/config/grafana/dashboards/integration-service-dashboard.json
@@ -1045,7 +1045,232 @@
       "yBucketBound": "auto",
       "yBucketNumber": null,
       "yBucketSize": null
+    },
+
+    {
+      "id": 26,
+      "gridPos": {
+        "h": 8,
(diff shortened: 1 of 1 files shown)


MartinBasti:
Would it make sense to plot all kinds of CR into one snapshot, so for example `Work Queue Duration Seconds` panel would show all snapshot, ITS, ... as separate plots in the same panel

jencull:
> Would it make sense to plot all kinds of CR into one panel per metric, so for example `Work Queue Duration Seconds` panel would show all snapshot, ITS, ... as separate plots in the same panel

That makes sense. Will do it this way for the screenshots and see what you think.

MartinBasti:
> Commit messages are descriptive enough

No they are not, it's not even written what SLI is being added in description and what measures

MartinBasti (approved):

hongweiliu17 (approved):
It is better to add the ticket number to PR title like `feat(STONEINTG-510)`, then we can find it easily in the future. Others look good to me.

PR #310, opened 2023-09-13
feat: new grafana panel for release latency seconds

jencull:
new metric release latency seconds
new grafana panel of same name
measures latency between integration tests
completing and release creation

### DIFF
--- a/config/grafana/dashboards/integration-service-dashboard.json
+++ b/config/grafana/dashboards/integration-service-dashboard.json
@@ -1310,6 +1310,76 @@
       "yBucketBound": "auto",
       "yBucketNumber": null,
       "yBucketSize": null
+    },
+    {
+      "cards": {
+        "cardPadding": null,
+        "cardRound": null
+      },
+      "color": {
+        "cardColor": "#b4ff00",
+        "colorScale": "sqrt",
+        "colorScheme": "interpolateOranges",
+        "exponent": 0.5,
+        "mode": "spectrum"
+      },
+      "dataFormat": "tsbuckets",
+      "description": "Measure release creation latency from end of testing to release created",
+      "fieldConfig": {
+        "defaults": {},
+        "overrides": []
+      },
+      "gridPos": {
+        "h": 8,
+        "w": 12,
+        "x": 0,
+        "y": 27
+      },
+      "heatmap": {},
+      "hideZeroBuckets": true,
+      "highlightCards": true,
+      "id": 28,
+      "legend": {
+        "show": false
+      },
+      "maxDataPoints": 25,
+      "pluginVersion": "7.5.17",
+      "reverseYBuckets": false,
+      "targets": [
+        {
+          "exemplar": true,
+          "expr": "histogram_quantile(0.90, sum(rate(release_latency_seconds_bucket[5m])) by (le)) ",
+          "format": "heatmap",
+          "interval": "",
+          "legendFormat": "{{le}}",
+          "refId": "A"
+        }
+      ],
+      "timeFrom": null,
+      "timeShift": null,
+      "title": "Latency of Release Creation",
+      "tooltip": {
+        "show": true,
+        "showHistogram": false
+      },
+      "type": "heatmap",
+      "xAxis": {
+        "show": true
+      },
+      "xBucketNumber": null,
+      "xBucketSize": null,
+      "yAxis": {
+        "decimals": null,
+        "format": "short",
+        "logBase": 1,
+        "max": null,
+        "min": null,
+        "show": true,
+        "splitFactor": null
+      },
+      "yBucketBound": "auto",
+      "yBucketNumber": null,
+      "yBucketSize": null
     }
 
   ],
--- a/controllers/snapshot/snapshot_adapter.go
+++ b/controllers/snapshot/snapshot_adapter.go
@@ -29,6 +29,7 @@ import (
 	"github.com/redhat-appstudio/integration-service/api/v1beta1"
 	"github.com/redhat-appstudio/integration-service/gitops"
 	h "github.com/redhat-appstudio/integration-service/helpers"
+	"github.com/redhat-appstudio/integration-service/metrics"
 	"github.com/redhat-appstudio/integration-service/release"
 	"github.com/redhat-appstudio/integration-service/tekton"
 
@@ -474,6 +475,8 @@ func (a *Adapter) createMissingReleasesForReleasePlans(application *applicationa
 		return err
 	}
 
+	firstRelease := true
+
 	for _, releasePlan := range *releasePlans {
 		releasePlan := releasePlan // G601
 		existingRelease := release.FindMatchingReleaseWithReleasePlan(releases, releasePlan)
@@ -506,6 +509,14 @@ func (a *Adapter) createMissingReleasesForReleasePlans(application *applicationa
 				return err
 			}
 			a.logger.Info("Marked Release status automated", "release.Name", newRelease.Name)
+		}
+		// Register the first release time for metrics calculation
+		if firstRelease {
+			startTime, ok := gitops.GetAppStudioTestsFinishedTime(a.snapshot)
+			if ok {
+				metrics.RegisterReleaseLatency(startTime)
+				firstRelease = false
+			}
 		}
 	}
 	return nil
--- a/gitops/snapshot.go
+++ b/gitops/snapshot.go
@@ -360,7 +360,29 @@ func HaveAppStudioTestsSucceeded(snapshot *applicationapiv1alpha1.Snapshot) bool
 	if meta.FindStatusCondition(snapshot.Status.Conditions, AppStudioTestSucceededCondition) == nil {
 		return meta.IsStatusConditionTrue(snapshot.Status.Conditions, LegacyTestSucceededCondition)
 	}
+
 	return meta.IsStatusConditionTrue(snapshot.Status.Conditions, AppStudioTestSucceededCondition)
+}
+
+// GetTestSucceededCondition checks status of tests on the snapshot
+func GetTestSucceededCondition(snapshot *applicationapiv1alpha1.Snapshot) (condition *metav1.Condition, ok bool) {
+
+	condition = meta.FindStatusCondition(snapshot.Status.Conditions, AppStudioTestSucceededCondition)
+	if condition == nil {
+		condition = meta.FindStatusCondition(snapshot.Status.Conditions, LegacyTestSucceededCondition)
+	}
+
+	ok = (condition != nil && condition.Status != metav1.ConditionUnknown)
+	return
+}
+
+// GetAppStudioTestsFinishedTime finds the timestamp of tests succeeded condition
+func GetAppStudioTestsFinishedTime(snapshot *applicationapiv1alpha1.Snapshot) (metav1.Time, bool) {
+	condition, ok := GetTestSucceededCondition(snapshot)
+	if ok {
+		return condition.LastTransitionTime, true
+	}
+	return metav1.Time{}, false
 }
 
 // CanSnapshotBePromoted checks if the Snapshot in question can be promoted for deployment and release.
--- a/metrics/integration.go
+++ b/metrics/integration.go
@@ -17,6 +17,8 @@ limitations under the License.
 package metrics
 
 import (
+	"time"
+
 	"github.com/prometheus/client_golang/prometheus"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/metrics"
@@ -87,6 +89,14 @@ var (
 		},
 		[]string{"type", "reason"},
 	)
+
+	ReleaseLatencySeconds = prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Name:    "release_latency_seconds",
+			Help:    "Latency between integration tests completion and release creation",
+			Buckets: []float64{0.05, 0.1, 0.5, 1, 2, 3, 4, 5, 10, 15, 30},
+		},
+	)
 )
 
 func RegisterCompletedSnapshot(conditiontype, reason string, startTime metav1.Time, completionTime *metav1.Time) {
@@ -131,6 +141,11 @@ func RegisterNewIntegrationPipelineRun(snapshotCreatedTime metav1.Time, pipeline
 	RegisterPipelineRunStarted(snapshotCreatedTime, pipelineRunStartTime)
 }
 
+func RegisterReleaseLatency(startTime metav1.Time) {
+	latency := time.Since(startTime.Time).Seconds()
+	ReleaseLatencySeconds.Observe(latency)
+}
+
 func init() {
 	metrics.Registry.MustRegister(
 		SnapshotCreatedToPipelineRunStartedSeconds,
@@ -141,5 +156,6 @@ func init() {
 		SnapshotDurationSeconds,
 		SnapshotInvalidTotal,
 		SnapshotTotal,
+		ReleaseLatencySeconds,
 	)
 }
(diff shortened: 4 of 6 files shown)


metrics/integration.go
```diff
@@ -100,6 +108,10 @@ func RegisterNewIntegrationPipelineRun(snapshotCreatedTime metav1.Time, pipeline
 	RegisterPipelineRunStarted(snapshotCreatedTime, pipelineRunStartTime)
 }
 
+func RegisterReleaseLatency(integrationTestsFinishTime metav1.Time, releaseCreationTime metav1.Time) {
```
MartinBasti, line 111: I may miss something, but where is this function used?

metrics/integration.go
```diff
@@ -61,6 +61,14 @@ var (
 		},
 		[]string{"type", "reason"},
 	)
+
+	ReleaseLatencySeconds = prometheus.NewHistogram(
+		prometheus.HistogramOpts{
+			Name:    "release_latency_seconds",
+			Help:    "Latency between integration tests completion and release creation",
+			Buckets: []float64{7, 15, 30, 60, 150, 300, 450, 600, 750, 900, 1050},
```
MartinBasti, line 69: this should be quick operation, I'd use the same buckets as in SnapshotCreatedToPipelineRunStartedSeconds

config/grafana/dashboards/integration-service-dashboard.json
```diff
@@ -1271,6 +1271,100 @@
...
 
+    {
+      "id": 28,
+      "gridPos": {
+        "h": 8,
+        "w": 12,
+        "x": 12,
+        "y": 42
+      },
+      "type": "graph",
```
jinqi7, line 1282: Why not use heatmap for express the Histogram type metric? It seems more clear.

jinqi7 (commented):
Test is needed for the newly added metric.

MartinBasti:
> [ x] Code coverage from testing does not decrease and new code is covered
Why is this checked? From coverage report new code isn't covered, also as mentioned Jing tests are missing

metrics/integration.go
jencull: thank you

config/grafana/dashboards/integration-service-dashboard.json
jencull: done. thank you.

config/grafana/dashboards/integration-service-dashboard.json
```diff
@@ -1310,6 +1311,77 @@
       "yBucketBound": "auto",
       "yBucketNumber": null,
       "yBucketSize": null
+    },
+
```
14rcole, line 1315: nit: remove newline

gitops/snapshot.go
```diff
@@ -324,6 +324,15 @@ func HaveAppStudioTestsFinished(snapshot *applicationapiv1alpha1.Snapshot) bool
...
 }
 
+func PrepareStartCreateRelease(snapshot *applicationapiv1alpha1.Snapshot) {
+	// Find the desired status condition
+	condition := meta.FindStatusCondition(snapshot.Status.Conditions, AppStudioTestSucceededCondition)
+
+	// Use the condition's LastTransitionTime timestamp as startCreateRelease timestamp
+	startCreateRelease := condition.LastTransitionTime
+	go metrics.RegisterReleaseLatency(startCreateRelease)
+}
```
14rcole, line 334: Two things here:

1. Where is `PrepareStartCreateRelease()` called?
2. I don't know that this needs to be its own function. I think you can take the two lines in here and just put them into `RegisterReleaseLatency()`

config/grafana/dashboards/integration-service-dashboard.json
jencull: If this pops up again its due to rebase and i can fix again.

gitops/snapshot.go
jencull: suggested changes made, thank you.

metrics/integration.go
jencull: error on my part

gitops/snapshot.go
```diff
@@ -344,6 +344,7 @@ func HaveAppStudioTestsSucceeded(snapshot *applicationapiv1alpha1.Snapshot) bool
 	if meta.FindStatusCondition(snapshot.Status.Conditions, AppStudioTestSucceededCondition) == nil {
 		return meta.IsStatusConditionTrue(snapshot.Status.Conditions, LegacyTestSucceededCondition)
 	}
+
```
14rcole, line 347: Why was this newline added?

metrics/integration.go
```diff
@@ -115,6 +127,13 @@ func RegisterNewIntegrationPipelineRun(snapshotCreatedTime metav1.Time, pipeline
 	RegisterPipelineRunStarted(snapshotCreatedTime, pipelineRunStartTime)
 }
 
+func RegisterReleaseLatency(snapshot *applicationapiv1alpha1.Snapshot) {
+	condition := meta.FindStatusCondition(snapshot.Status.Conditions, "AppStudioTestSucceeded")
```
14rcole, line 131: 1. You should probably use `gitops.AppStudioTestSucceededCondition` instead of the string here
2. I just realized that we also check for `gitops.LegacyTestsSucceededCondition` in other parts of the code.  Is that something we still need to check for or is it so out-of-date that its not a concern? @Josh-Everett it looks like you were responsible for this code at one point.

gitops/snapshot.go
jencull: i ran go fmt ./... if it was added automatically it will add again after I remove it. But I will try remove it and format it again and see what happens

metrics/integration.go
jencull: 1. calling it that way created a circular import because gitops imports metrics. Calling the string was a way around that. Am open to suggestions on how to get around that circular import problem.
2. The function in gitops checks if AppStudioTestSucceeded is true. If its not true then it checks for LegacyTests and then returns true for 'status condition' if they are found instead of the first set. So either one of those being true is assigned to statusCondition is my understanding of this function. (does not mean I am correct!!)

MartinBasti: 1. in metrics in `RegisterReleaseLatency` just use `startTime metav1.Time` as parameter (keep endCreateRelase as time.Now()),  and define function that extracts startTime directly in snapshot controller adapter
2. you have to write new function to get data, `confition, ok := getStatusCondition(snapshot)` or smth like that, and also keep LegacyTestsSucceededCondition there

metrics/integration_test.go
```diff
@@ -217,4 +218,41 @@ var _ = Describe("Metrics Integration", Ordered, func() {
...
+			condition := meta.FindStatusCondition(snapshot.Status.Conditions, "AppStudioTestSucceeded")
+			startTime := condition.LastTransitionTime
+			endTime := metav1.NewTime(time.Now())
+
+			// Calculate observed latency
+			observedLatency := endTime.Sub(startTime.Time).Seconds()
+
+			// Ensure the observed latency is within a range ±5ms 60 seconds
+			Expect(observedLatency).To(BeNumerically(">=", 59.995))
+			Expect(observedLatency).To(BeNumerically("<=", 60.005))
```
MartinBasti, line 253: I have hard time to understand, how this part of the test is testing `RegisterReleaseLatency`, it looks like the test is testing itself only

controllers/snapshot/snapshot_adapter.go
```diff
@@ -366,6 +369,29 @@ func (a *Adapter) EnsureGlobalCandidateImageUpdated() (controller.OperationResul
 	return controller.ContinueProcessing()
 }
 
+func GetStatusCondition(snapshot *applicationapiv1alpha1.Snapshot) (condition *metav1.Condition, ok bool) {
```
dirgim, line 372: Thinking if this function would make more sense being placed in [gitops/snapshot.go](https://github.com/redhat-appstudio/integration-service/blob/main/gitops/snapshot.go)?

controllers/snapshot/snapshot_adapter.go
```diff
@@ -366,6 +369,29 @@ func (a *Adapter) EnsureGlobalCandidateImageUpdated() (controller.OperationResul
...
+	const (
```
dirgim, line 373: Could [these constants](https://github.com/redhat-appstudio/integration-service/blob/02c829368d912e684b82dbab204b34f56e576fba/gitops/snapshot.go#L86-L90) be used instead of defining them here?

controllers/snapshot/snapshot_adapter.go
```diff
@@ -366,6 +369,29 @@ func (a *Adapter) EnsureGlobalCandidateImageUpdated() (controller.OperationResul
...
+	condition = meta.FindStatusCondition(snapshot.Status.Conditions, AppStudioTestSucceededCondition)
+	if condition == nil {
+		condition = meta.FindStatusCondition(snapshot.Status.Conditions, LegacyTestSucceededCondition)
+	}
+
+	ok = (condition != nil && condition.Status != metav1.ConditionUnknown)
+	return
+}
+
+func GetAppStudioTestsFinishedTime(snapshot *applicationapiv1alpha1.Snapshot) (metav1.Time, bool) {
```
dirgim, line 387: Same as above, this function might make more sense being placed in [gitops/snapshot.go](https://github.com/redhat-appstudio/integration-service/blob/main/gitops/snapshot.go).

controllers/snapshot/snapshot_adapter.go
jencull: It initially started out life in gitops/snapshot.go. Current change is based on feedback from mbasti on that code, feedback:

1. in metrics in RegisterReleaseLatency just use startTime metav1.Time as parameter (keep endCreateRelase as time.Now()), and define function that extracts startTime directly in snapshot controller adapter
2. you have to write new function to get data, confition, ok := getStatusCondition(snapshot) or smth like that, and also keep LegacyTestsSucceededCondition there

Its possible that number 2 was to remain in gitops/snapshot.go and I misunderstood. i will have a look at that and see how it might work.

MartinBasti: works for me to have it in gitops/snapshot

controllers/snapshot/snapshot_adapter.go
```diff
@@ -463,6 +494,12 @@ func (a *Adapter) EnsureSnapshotEnvironmentBindingExist() (controller.OperationR
 
 		}
 	}
+	// Get the start time when AppStudioTests have finished
```
MartinBasti, line 497: This is not the right place for this metric, it cares about release not SEBs, and also this metric shouldn't be duplicated, I see it on 2 places already

controllers/snapshot/snapshot_adapter.go
```diff
@@ -396,6 +422,11 @@ func (a *Adapter) EnsureAllReleasesExist() (controller.OperationResult, error) {
 			a.snapshot, h.LogActionUpdate)
 		return controller.RequeueOnErrorOrStop(a.client.Status().Patch(a.context, a.snapshot, patch))
 	}
+	startTime, ok := GetAppStudioTestsFinishedTime(a.snapshot)
```
MartinBasti, line 425: This looks like the right place

MartinBasti: actually it's not, when we rerun reconciliation, this may generate wrong statistics, it should be recorded only when release is created, we cannot tell from this place

MartinBasti: but recorded only once, if at least one release has been created

metrics/integration.go
```diff
@@ -115,6 +125,12 @@ func RegisterNewIntegrationPipelineRun(snapshotCreatedTime metav1.Time, pipeline
 	RegisterPipelineRunStarted(snapshotCreatedTime, pipelineRunStartTime)
 }
 
+// RegisterReleaseLatency, assess the time between HaveAppStudioTestsFinished and Release created. .
+func RegisterReleaseLatency(startTime metav1.Time) {
+	endCreateRelease := &metav1.Time{Time: time.Now()}
+	ReleaseLatencySeconds.Observe(endCreateRelease.Sub(startTime.Time).Seconds())
```
14rcole, line 131: You don't even need to do the `time.Now()` thing here.  You can just do 
```
latency := time.Since(startTime).Seconds()
ReleaseLatencySeconds.Observe(latency)
```

controllers/snapshot/snapshot_adapter.go
```diff
@@ -507,6 +508,11 @@ func (a *Adapter) createMissingReleasesForReleasePlans(application *applicationa
 			}
 			a.logger.Info("Marked Release status automated", "release.Name", newRelease.Name)
 		}
+		//get time tests finished and use as start time to calculate release created latency
+		startTime, ok := gitops.GetAppStudioTestsFinishedTime(a.snapshot)
```
14rcole, line 512: Does it make more sense to title this function `GetReleaseStartTime?`

MartinBasti: No, function returns time when test finished. It has nothing with release

gitops/snapshot.go
```diff
@@ -360,9 +360,31 @@ func HaveAppStudioTestsSucceeded(snapshot *applicationapiv1alpha1.Snapshot) bool
...
 	return meta.IsStatusConditionTrue(snapshot.Status.Conditions, AppStudioTestSucceededCondition)
 }
 
+// checks status of tests on the snapshot as this timestamp will be used as startTime fore creating release
+func GetStatusCondition(snapshot *applicationapiv1alpha1.Snapshot) (condition *metav1.Condition, ok bool) {
```
MartinBasti, line 368: Name function should be updated to getting status of which condition,  `GetTestSucceededCondition`

gitops/snapshot.go
```diff
@@ -360,9 +360,31 @@ func HaveAppStudioTestsSucceeded(snapshot *applicationapiv1alpha1.Snapshot) bool
 	if meta.FindStatusCondition(snapshot.Status.Conditions, AppStudioTestSucceededCondition) == nil {
 		return meta.IsStatusConditionTrue(snapshot.Status.Conditions, LegacyTestSucceededCondition)
 	}
+
 	return meta.IsStatusConditionTrue(snapshot.Status.Conditions, AppStudioTestSucceededCondition)
 }
 
+// checks status of tests on the snapshot as this timestamp will be used as startTime fore creating release
```
MartinBasti, line 367:
```suggestion
// <NAME OF FUNCTION> checks status of tests on the snapshot as this timestamp will be used as startTime fore creating release
```

controllers/snapshot/snapshot_adapter.go
```diff
@@ -507,6 +508,11 @@ func (a *Adapter) createMissingReleasesForReleasePlans(application *applicationa
 			}
 			a.logger.Info("Marked Release status automated", "release.Name", newRelease.Name)
 		}
+		//get time HaveAppStudioTestsFinished
```
dirgim, line 511: While this would work fine in case there is only one `releasePlan`, I think it would cause issue when there are multiple of them because the latency would be recorded multiple times, once for every release created.

If I understood the [STONEINTG-532](https://issues.redhat.com/browse/STONEINTG-532) SLI correctly, the intent is to measure the time between finishing testing and creating the first release. If that's true, then we'd need to ensure that this is called only on the first pass. @MartinBasti feel free to correct me on this.

jencull: Something like this might work? @MartinBasti 

firstRelease := true

for _, releasePlan := range *releasePlans {
    // this is the existing code here

    if firstRelease {
        startTime, ok := gitops.GetAppStudioTestsFinishedTime(a.snapshot)
        if ok {
            metrics.RegisterReleaseLatency(startTime)
            firstRelease = false
        }
    }
}

return nil

MartinBasti: yes, that sound's good

jencull: done and this flag is tested in the integration_test, where the metric function runs, to ensure only the first release is measured and the flag is working

dirgim: The solution looks good to me, thanks.

dirgim (approved):
LGTM

Josh-Everett (approved):

MartinBasti (approved):

PR #317, opened 2023-09-14
fix(stoneintg-564): ignore empty values

14rcole:
Previously empty values in the violations graphs were reported as zero.  This is because in order to calculate the percentage of requests meeting our thresholds we divide the "good" requests by the total requests.  If there are no total requests then grafana sets 0/0 to 0 rather than dividing by 0.  This change ignores the empty values and changes the display from a line to bars to improve readability.

### DIFF
--- a/config/grafana/dashboards/integration-service-dashboard.json
+++ b/config/grafana/dashboards/integration-service-dashboard.json
@@ -143,110 +143,95 @@
       }
     },
     {
-      "aliasColors": {},
-      "bars": false,
-      "dashLength": 10,
-      "dashes": false,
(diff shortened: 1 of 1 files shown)


MartinBasti (changes requested):
This PR updates one SLO, but we have at least two SLOs

14rcole:
Five years and I still can't commit things properly.  Fixed now, I double-checked and I only saw those two SLOs having this issue

MartinBasti (approved):

jinqi7 (approved):
LGTM

PR #327, opened 2023-09-21
feat(STONEINTG-538): refactor EnsureAllIntegrationTestPipelinesExist()

hongweiliu17:
* refactor the function to create integrationPipelinerun

### DIFF
--- a/controllers/snapshot/snapshot_adapter.go
+++ b/controllers/snapshot/snapshot_adapter.go
@@ -120,29 +120,14 @@ func (a *Adapter) EnsureAllIntegrationTestPipelinesExist() (controller.Operation
 					"integrationTestScenario.Name", integrationTestScenario.Name,
 					"len(integrationPipelineRuns)", len(*integrationPipelineRuns))
 			} else {
-				a.logger.Info("Creating new pipelinerun for integrationTestscenario",
-					"integrationTestScenario.Name", integrationTestScenario.Name)
 				pipelineRun, err := a.createIntegrationPipelineRun(a.application, &integrationTestScenario, a.snapshot)
 				if err != nil {
 					a.logger.Error(err, "Failed to create pipelineRun for snapshot and scenario")
 					return controller.RequeueWithError(err)
 				}
-				a.logger.LogAuditEvent("IntegrationTestscenario pipeline has been created", pipelineRun, h.LogActionAdd,
-					"integrationTestScenario.Name", integrationTestScenario.Name)
 				testStatuses.UpdateTestStatusIfChanged(
 					integrationTestScenario.Name, gitops.IntegrationTestStatusInProgress,
 					fmt.Sprintf("IntegrationTestScenario pipeline '%s' has been created", pipelineRun.Name))
-				gitops.PrepareToRegisterIntegrationPipelineRun(a.snapshot)
-				if gitops.IsSnapshotNotStarted(a.snapshot) {
-					_, err := gitops.MarkSnapshotIntegrationStatusAsInProgress(a.client, a.context, a.snapshot, "Snapshot starts being tested by the integrationPipelineRun")
-					if err != nil {
-						a.logger.Error(err, "Failed to update integration status condition to in progress for snapshot")
-					} else {
-						a.logger.LogAuditEvent("Snapshot integration status marked as In Progress. Snapshot starts being tested by the integrationPipelineRun",
-							a.snapshot, h.LogActionUpdate)
-					}
-				}
-
 			}
 		}
 
@@ -538,6 +523,9 @@ func (a *Adapter) findAvailableEnvironments() (*[]applicationapiv1alpha1.Environ
 // createIntegrationPipelineRun creates and returns a new integration PipelineRun. The Pipeline information and the parameters to it
 // will be extracted from the given integrationScenario. The integration's Snapshot will also be passed to the integration PipelineRun.
 func (a *Adapter) createIntegrationPipelineRun(application *applicationapiv1alpha1.Application, integrationTestScenario *v1beta1.IntegrationTestScenario, snapshot *applicationapiv1alpha1.Snapshot) (*pipeline.PipelineRun, error) {
+	a.logger.Info("Creating new pipelinerun for integrationTestscenario",
+		"integrationTestScenario.Name", integrationTestScenario.Name)
+
 	pipelineRun := tekton.NewIntegrationPipelineRun(snapshot.Name, application.Namespace, *integrationTestScenario).
 		WithSnapshot(snapshot).
 		WithIntegrationLabels(integrationTestScenario).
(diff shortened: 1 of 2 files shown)


controllers/snapshot/snapshot_adapter.go
```diff
@@ -120,29 +120,14 @@ func (a *Adapter) EnsureAllIntegrationTestPipelinesExist() (controller.Operation
 					"integrationTestScenario.Name", integrationTestScenario.Name,
 					"len(integrationPipelineRuns)", len(*integrationPipelineRuns))
 			} else {
-				a.logger.Info("Creating new pipelinerun for integrationTestscenario",
-					"integrationTestScenario.Name", integrationTestScenario.Name)
 				pipelineRun, err := a.createIntegrationPipelineRun(a.application, &integrationTestScenario, a.snapshot)
-				if err != nil {
+				if pipelineRun == nil && err != nil {
 					a.logger.Error(err, "Failed to create pipelineRun for snapshot and scenario")
```
jsztuka, line 125: We may inform which pipelinerun failed as part as prompting error in the log.
Just out of curiosity, asking about the condition:  **pipelineRun == nil && err != nil**
What is the case when pipelinerun is not nil and error is not nil either? Is there such situation?

controllers/snapshot/snapshot_adapter.go
```diff
@@ -554,13 +542,25 @@ func (a *Adapter) createIntegrationPipelineRun(application *applicationapiv1alph
...
+		return nil, fmt.Errorf("failed to call client.Create to create pipelineRun for snapshot %s: %w", snapshot.Name, err)
 	}
 
+	a.logger.LogAuditEvent("IntegrationTestscenario pipeline has been created", pipelineRun, h.LogActionAdd,
+		"integrationTestScenario.Name", integrationTestScenario.Name)
+	gitops.PrepareToRegisterIntegrationPipelineRun(a.snapshot)
+	if gitops.IsSnapshotNotStarted(a.snapshot) {
+		_, err := gitops.MarkSnapshotIntegrationStatusAsInProgress(a.client, a.context, a.snapshot, "Snapshot starts being tested by the integrationPipelineRun")
+		if err != nil {
+			a.logger.Error(err, "Failed to update integration status condition to in progress for snapshot")
```
jsztuka, line 558: Can we include snapshot iitself here? So its easier to search for it.

dirgim: I think that the Snapshot that's being reconciled will always be reported in the logs on account of the `r.Log.WithValues` call [here](https://github.com/redhat-appstudio/integration-service/blob/d302862916d13621967f34c14f5d91f46dac8d28/controllers/snapshot/snapshot_controller.go#L78).

controllers/snapshot/snapshot_adapter.go
MartinBasti: I agree, if there is err != nil, we shouldn't care about return variable at all

hongweiliu17: 1. When the pipelineRun is not created successfully, I don't think we need to log it out.
2. When the pipelineRun is created successfully but it fails to update snapshot to `InProgress`, we just log some error but won't `return controller.RequeueWithError(err)`, see [here](https://github.com/redhat-appstudio/integration-service/blob/main/controllers/snapshot/snapshot_adapter.go#L138-L139).
    When we failed to set ControllerReference or create pipelineRun, we will return `nil, error` and then `return controller.RequeueWithError(err)`

controllers/snapshot/snapshot_adapter.go
hongweiliu17: Thanks Kruno for reply. Yes.

controllers/snapshot/snapshot_adapter.go
MartinBasti: 2. it still means that we need to check err != nil only in that condition, it cannot happen that we return error and the working pipeline at the same time

hongweiliu17: Good point, there was a check for `err != nil`, but I removed it. Updated. Thanks!

MartinBasti (approved):
Looks good to me, I'm not sure if it covers everything what Ryan would need in his epic, but this particular PR is good start, further improvements can be in new PRs.

dirgim (approved):
LGTM

dirgim:
@14rcole can you also take a look if this aligns with the plans for [STONEINTG-463](https://issues.redhat.com/browse/STONEINTG-463)?

controllers/snapshot/snapshot_adapter.go
```diff
@@ -120,29 +120,14 @@ func (a *Adapter) EnsureAllIntegrationTestPipelinesExist() (controller.Operation
...
 				pipelineRun, err := a.createIntegrationPipelineRun(a.application, &integrationTestScenario, a.snapshot)
 				if err != nil {
 					a.logger.Error(err, "Failed to create pipelineRun for snapshot and scenario")
 					return controller.RequeueWithError(err)
 				}
-				a.logger.LogAuditEvent("IntegrationTestscenario pipeline has been created", pipelineRun, h.LogActionAdd,
-					"integrationTestScenario.Name", integrationTestScenario.Name)
 				testStatuses.UpdateTestStatusIfChanged(
 					integrationTestScenario.Name, gitops.IntegrationTestStatusInProgress,
 					fmt.Sprintf("IntegrationTestScenario pipeline '%s' has been created", pipelineRun.Name))
```
14rcole, line 130: This is the only piece I'm not sure about.  Why isn't this also being moved to `createIntegrationPipelineRun()`? We would need to update the test status for the scenario in the case of a re-run.  Do you think this is just out of scope?

14rcole:
I had a question about another unmodified part of the code
```
if !reflect.ValueOf(integrationTestScenario.Spec.Environment).IsZero() {
				// the test pipeline for scenario needing an ephemeral environment will be handled in STONEINTG-333
				a.logger.Info("IntegrationTestScenario has environment defined, skipping creation of pipelinerun.", "IntegrationTestScenario", integrationTestScenario)
				continue
			}
```

Will `integrationTestScenario.Spec.Environment` always be defined for a scenario that has been run before?  Or do we need to check if the environment is defined in the case of a re-run?

controllers/snapshot/snapshot_adapter.go
hongweiliu17: I had the similar question [here](https://redhat-internal.slack.com/archives/C034UFMUSV6/p1695279001912279), the status will be reported [here](https://github.com/redhat-appstudio/integration-service/blob/main/controllers/integrationpipeline/integrationpipeline_adapter.go#L194) as `Inprogress/TestPass/TestFailed` as well, I think it is reasonable to not to add a lot of code here when creating PLR in snapshot and binding controller.

hongweiliu17:
> I had a question about another unmodified part of the code
> 
> ```
> if !reflect.ValueOf(integrationTestScenario.Spec.Environment).IsZero() {
> 	// the test pipeline for scenario needing an ephemeral environment will be handled in STONEINTG-333
> 	a.logger.Info("IntegrationTestScenario has environment defined, skipping creation of pipelinerun.", "IntegrationTestScenario", integrationTestScenario)
> 	continue
> }
> ```
> 
> Will `integrationTestScenario.Spec.Environment` always be defined for a scenario that has been run before? Or do we need to check if the environment is defined in the case of a re-run?



> I had a question about another unmodified part of the code
> 
> ```
> if !reflect.ValueOf(integrationTestScenario.Spec.Environment).IsZero() {
> 	// the test pipeline for scenario needing an ephemeral environment will be handled in STONEINTG-333
> 	a.logger.Info("IntegrationTestScenario has environment defined, skipping creation of pipelinerun.", "IntegrationTestScenario", integrationTestScenario)
> 	continue
> }
> ```
> 
> Will `integrationTestScenario.Spec.Environment` always be defined for a scenario that has been run before? Or do we need to check if the environment is defined in the case of a re-run?

I guess the rerun logic would be like this: If environment is defined in ITS, call `CreateCopyOfEnvironment()` which is being refactored in [STONEINTG-539](https://issues.redhat.com/browse/STONEINTG-539), otherwise, call `CreateIntegrationPLR()` defined here.

14rcole:
> I guess the rerun logic would be like this: If environment is defined in ITS, call CreateCopyOfEnvironment() which is being refactored in [STONEINTG-539](https://issues.redhat.com/browse/STONEINTG-539), otherwise, call CreateIntegrationPLR() defined here. Please correct me if my understanding is incorrect.

That sounds right

Josh-Everett (approved):

hongweiliu17:
> > I guess the rerun logic would be like this: If environment is defined in ITS, call CreateCopyOfEnvironment() which is being refactored in [STONEINTG-539](https://issues.redhat.com/browse/STONEINTG-539), otherwise, call CreateIntegrationPLR() defined here. Please correct me if my understanding is incorrect.
> 
> That sounds right

@14rcole Thanks for checking. So does this PR look good to you?

dheerajodha (approved):

sonam1412:
> > > I guess the rerun logic would be like this: If environment is defined in ITS, call CreateCopyOfEnvironment() which is being refactored in [STONEINTG-539](https://issues.redhat.com/browse/STONEINTG-539), otherwise, call CreateIntegrationPLR() defined here. Please correct me if my understanding is incorrect.
> > 
> > 
> > That sounds right
> 
> @14rcole Thanks for checking. So does this PR look good to you?

@14rcole Can you please confirm if this looks good to you and can be merged  ?

14rcole (approved):

PR #329, opened 2023-09-21
fix(stoneintg-579): do not requeue snapshot if component cannot be found

14rcole:
If the snapshot component cannot be found, we should retry.  If it still cannot be found the snapshot should not be requeued.  This change adds retries to the controller and does not signal the controller to retry if the component cannot be found.

### DIFF
--- a/controllers/snapshot/snapshot_controller.go
+++ b/controllers/snapshot/snapshot_controller.go
@@ -29,6 +29,7 @@ import (
 	"github.com/redhat-appstudio/operator-toolkit/controller"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/util/retry"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
@@ -96,10 +97,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	}
 	logger = logger.WithApp(*application)
 
-	component, err := loader.GetComponentFromSnapshot(r.Client, ctx, snapshot)
+	var component *applicationapiv1alpha1.Component
+	err = retry.OnError(retry.DefaultRetry, func(_ error) bool { return true }, func() error {
+		component, err = loader.GetComponentFromSnapshot(r.Client, ctx, snapshot)
+		return err
+	})
 	if err != nil {
-		logger.Error(err, "Failed to get Component from the Snapshot")
-		return ctrl.Result{}, err
+		return logger.HandleLoaderError(err, "Component", "Snapshot")
 	}
 
 	adapter := NewAdapter(snapshot, application, component, logger, loader, r.Client, ctx)
--- a/controllers/snapshot/snapshot_controller_test.go
+++ b/controllers/snapshot/snapshot_controller_test.go
@@ -1,8 +1,9 @@
 package snapshot
 
 import (
-	"k8s.io/apimachinery/pkg/api/errors"
 	"reflect"
+
+	"k8s.io/apimachinery/pkg/api/errors"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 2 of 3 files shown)


controllers/snapshot/snapshot_controller.go
```diff
@@ -96,10 +97,18 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+			logger.Error(err, "Failed to get Component from the Snapshot")
+			return err
+		}
+		return nil
+	})
 	if err != nil {
-		logger.Error(err, "Failed to get Component from the Snapshot")
-		return ctrl.Result{}, err
+		logger.Error(err, "Could not get component from Snapshot.  Component may have been removed.  Declining to proceed with reconciliation")
+		return ctrl.Result{}, nil
```
MartinBasti, line 111: Here you should exactly check if error is NotFound error, and only then stop reconciliation.
```
errors.IsNotFound(err)
```

controllers/snapshot/snapshot_controller.go
```diff
@@ -96,10 +97,23 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	}
 	logger = logger.WithApp(*application)
 
-	component, err := loader.GetComponentFromSnapshot(r.Client, ctx, snapshot)
+	var component *applicationapiv1alpha1.Component
+	err = retry.OnError(retry.DefaultRetry, func(_ error) bool { return true }, func() error {
+		component, err = loader.GetComponentFromSnapshot(r.Client, ctx, snapshot)
+		if err != nil {
+			logger.Error(err, "Failed to get Component from the Snapshot")
```
MartinBasti, line 104: This should be probably a debug message or removed altogether

controllers/snapshot/snapshot_controller.go
```diff
@@ -96,10 +97,23 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+			return err
+		}
+		return nil
+	})
 	if err != nil {
-		logger.Error(err, "Failed to get Component from the Snapshot")
-		return ctrl.Result{}, err
+		if errors.IsNotFound(err) {
+			logger.Error(err, "Could not get component from Snapshot.  Component may have been removed.  Declining to proceed with reconciliation")
```
MartinBasti, line 111: We should IMHO just log at info level, it's expected that component can be removed while we keep snapshot CR present

MartinBasti (commented):
I commented about logs, I think we should reduce amount of ERROR messages and keep ERROR log level for real errors, not expected states of system.

SRE must monitor errors and this would generate false positive error rates

controllers/snapshot/snapshot_controller.go
```diff
@@ -96,10 +97,23 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+			return ctrl.Result{}, nil
+		} else {
```
MartinBasti, line 113: This else doesn't have to be here, if statement has return in it

controllers/snapshot/snapshot_controller.go
14rcole: I kept it as an error to match the log levels of [these statements](https://github.com/redhat-appstudio/integration-service/pull/329/files/679cab5fd254920f4e2f0bbad72bbe1b3c6cf474#diff-bae1efbc7e547dded9da5f1294a79a891ee187e482dcb304bfe039b5302c7932R85) above it.  I think it's better for them to be consistent.

MartinBasti: those are errors, that state shouldn't ever happen, but here we are dealing with action that user intentionally removed component.

controllers/snapshot/snapshot_controller.go
```diff
@@ -96,8 +97,16 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
 
-	component, err := loader.GetComponentFromSnapshot(r.Client, ctx, snapshot)
+	var component *applicationapiv1alpha1.Component
+	err = retry.OnError(retry.DefaultRetry, func(_ error) bool { return true }, func() error {
+		component, err = loader.GetComponentFromSnapshot(r.Client, ctx, snapshot)
+		return err
+	})
 	if err != nil {
+		if errors.IsNotFound(err) {
+			logger.Info(err, "Could not get component from Snapshot.  Component may have been removed.  Declining to proceed with reconciliation")
```
MartinBasti, line 107: `err` param cannot be used here

14rcole: Why, does `logger.Info` only allow for basic strings?

14rcole: nvm, I dug through code

Josh-Everett (approved):
lgtm

MartinBasti:
I think that in case of major change in PR, my previous approvals should be discarded and not reused.

I was against to have it as part of `log`, this PR just implemented it as logger method, only in another module. Like what's difference.

PR #331, opened 2023-09-22
fix(STONEINTG-595): resolve race condition about DeploymentTargetClaim

jinqi7:

Changed  'Requeues the environment cleanup if the binding is younger
 than the threshold' test case to totally use mocked context CRs.

### DIFF
--- a/controllers/binding/snapshotenvironmentbinding_adapter_test.go
+++ b/controllers/binding/snapshotenvironmentbinding_adapter_test.go
@@ -236,8 +236,8 @@ var _ = Describe("Binding Adapter", Ordered, func() {
 
 		deploymentTargetClaim = &applicationapiv1alpha1.DeploymentTargetClaim{
 			ObjectMeta: metav1.ObjectMeta{
-				GenerateName: "dtc" + "-",
-				Namespace:    "default",
+				Name:      "dtcname",
+				Namespace: "default",
 			},
 			Spec: applicationapiv1alpha1.DeploymentTargetClaimSpec{
 				DeploymentTargetClassName: applicationapiv1alpha1.DeploymentTargetClassName("dtcls-name"),
@@ -431,7 +431,7 @@ var _ = Describe("Binding Adapter", Ordered, func() {
 		}, time.Second*10).Should(BeTrue())
 	})
 
-	It("ensures ephemeral environment is deleted for the given pipelineRun ", func() {
+	It("ensures ephemeral environment is deleted for the given pipelineRun", func() {
 		var buf bytes.Buffer
 		log := helpers.IntegrationLogger{Logger: buflogr.NewWithBuffer(&buf)}
 		hasEnv.Spec.Tags = append(hasEnv.Spec.Tags, "ephemeral")
@@ -482,6 +482,8 @@ var _ = Describe("Binding Adapter", Ordered, func() {
 
 	It("Requeues the environment cleanup if the binding is younger than the threshold", func() {
 		var buf bytes.Buffer
+		var dtc *applicationapiv1alpha1.DeploymentTargetClaim
+
 		log := helpers.IntegrationLogger{Logger: buflogr.NewWithBuffer(&buf)}
 		hasEnv.Spec.Tags = append(hasEnv.Spec.Tags, "ephemeral")
 		hasBinding.Status = applicationapiv1alpha1.SnapshotEnvironmentBindingStatus{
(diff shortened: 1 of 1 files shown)


MartinBasti:
I don't understand this fix, why we cannot use the same CR?
Shouldn't be something fixed in `BeforeEach` to have correct data in each test case?

jinqi7:
> I don't understand this fix, why we cannot use the same CR? Shouldn't be something fixed in `BeforeEach` to have correct data in each test case?

I tried that way to change something in `BeforeEach` . But after the case succeeded with some change, error from another case came up. Such as  there are two cases that expect `EnsureIntegrationTestPipelineForScenarioExists` succeeded. One succeed, then the other failed occasionally.  Adding special CRs is to resolve the race condition thoroughly.

jinqi7:
> I don't understand this fix, why we cannot use the same CR? Shouldn't be something fixed in `BeforeEach` to have correct data in each test case?

The DeploymentTargetClaim CR in `BeforeEach` is right as currently placed.  The case "ensures ephemeral environment is deleted for the given pipelineRun" which deletes the DTC is the suspicious to cause the race condition here. Switch it down to reduce the possibility.

controllers/binding/snapshotenvironmentbinding_adapter_test.go
```diff
@@ -505,14 +458,25 @@ var _ = Describe("Binding Adapter", Ordered, func() {
...
 			},
+			{
+				ContextKey: loader.DeploymentTargetClaimContextKey,
+				Resource:   deploymentTargetClaim,
+			},
 		})
 
-		dtc, _ := adapter.loader.GetDeploymentTargetClaimForEnvironment(k8sClient, adapter.context, hasEnv)
-		Expect(dtc).NotTo(BeNil())
+		Eventually(func() error {
```
hongweiliu17, line 475: We don't need this `Eventually` as long as we use mocked dtc, the mocked dtc won't be really deleted.

controllers/binding/snapshotenvironmentbinding_adapter_test.go
```diff
@@ -588,6 +552,54 @@ var _ = Describe("Binding Adapter", Ordered, func() {
 			Expect(ok).To(BeTrue())
 			Expect(detail.Status).To(Equal(gitops.IntegrationTestStatusDeploymentError))
 		})
+	})
 
+	It("ensures ephemeral environment is deleted for the given pipelineRun", func() {
```
hongweiliu17, line 557: I think we don't need to move this case here as long as as case `Requeues the environment cleanup if the binding is younger than the threshold` use the mocked dtc.

controllers/binding/snapshotenvironmentbinding_adapter_test.go
jinqi7: Then I don't understand why the error in STONEINTG-595 happened. Can you please explain it ?

jinqi7: > We don't need this `Eventually` as long as we use mocked dtc, the mocked dtc won't be really deleted.

Here is waiting the DTC created in `BeforeEach`.

controllers/binding/snapshotenvironmentbinding_adapter_test.go
```diff
@@ -431,57 +431,10 @@ var _ = Describe("Binding Adapter", Ordered, func() {
 		}, time.Second*10).Should(BeTrue())
 	})
 
-	It("ensures ephemeral environment is deleted for the given pipelineRun ", func() {
```
MartinBasti, line 434: Why was this test moved to the end of file?
It generates hard to read git diff, can we keep the test on the same place? I don't see reason for moving it

controllers/binding/snapshotenvironmentbinding_adapter_test.go
jinqi7: moved back.

controllers/binding/snapshotenvironmentbinding_adapter_test.go
jinqi7: Moved back

controllers/binding/snapshotenvironmentbinding_adapter_test.go
jinqi7: Change to use DTC from mockedData.

jinqi7: @hongweiliu17, please help to review it again.

controllers/binding/snapshotenvironmentbinding_adapter_test.go
```diff
@@ -501,14 +503,30 @@ var _ = Describe("Binding Adapter", Ordered, func() {
...
+				ContextKey: loader.DeploymentTargetClaimContextKey,
+				Resource:   deploymentTargetClaim,
+			},
+			{
+				ContextKey: loader.SnapshotEnvironmentBindingContextKey,
+				Resource:   hasBinding,
+			},
 		})
 
 		dtc, _ := adapter.loader.GetDeploymentTargetClaimForEnvironment(k8sClient, adapter.context, hasEnv)
```
dirgim, line 532: Since the `var dtc` definition was introduced above, the source check is complaining about this assignment:
```
no new variables on left side of :=
```
Changing to `=` should fix it.

dirgim: Ah, already taken care of, thanks @jinqi7 !

MartinBasti (approved):

controllers/binding/snapshotenvironmentbinding_adapter_test.go
```diff
@@ -501,17 +503,33 @@ var _ = Describe("Binding Adapter", Ordered, func() {
...
+				Resource:   deploymentTargetClaim,
+			},
+			{
+				ContextKey: loader.SnapshotEnvironmentBindingContextKey,
+				Resource:   hasBinding,
+			},
 		})
 
-		dtc, _ := adapter.loader.GetDeploymentTargetClaimForEnvironment(k8sClient, adapter.context, hasEnv)
+		dtc, _ = adapter.loader.GetDeploymentTargetClaimForEnvironment(k8sClient, adapter.context, hasEnv)
```
hongweiliu17, line 532: You don't need to declare `dtc` above, can just keep `:=`, but it doesn't matter. LGTM

hongweiliu17 (approved):

dirgim (approved):
+1

dirgim:
@hongweiliu17 can you check the remaining open conversations and resolve them if they've been addressed?

hongweiliu17:
> @hongweiliu17 can you check the remaining open conversations and resolve them if they've been addressed?

Thanks Kruno for reminder. We discussed all my points comments yesterday so that I forgot to mark them in this PR.

PR #335, opened 2023-09-26
fix(stoneintg-582): do not requeue SEB if snapshot cannot be found

14rcole:
Ordinarilly if the snapshot associated with an SEB cannot be found we fail and requeue the reconciliation for that SEB.  This change adds retries to the retrieval of the Snapshot and does away with the requeue if it cannot be found.  This way if the user deletes a snapshot the SEB will not be permanently stuck in a reconcile state.

## Maintainers will complete the following section

- [x] Commit messages are descriptive enough ([hints](https://www.freecodecamp.org/news/how-to-write-better-git-commit-messages/))
- [x] Code coverage from testing does not decrease and new code is covered
- [x] [Controllers diagrams](https://github.com/redhat-appstudio/integration-service/tree/main/docs) are updated when PR changes controllers code  (if applicable)

PR #338, opened 2023-09-28
fix(stoneintg-584): do not requeue SEB if application cannot be found

14rcole:
Ordinarilly if the application associated with an SEB cannot be found we fail and requeue the reconciliation for that SEB.  This change adds retries to the retrieval of the application and does away with the requeue if it cannot be found.  This way if the user deletes an application the SEB will not be permanently stuck in a reconcile state.

### DIFF
--- a/controllers/binding/snapshotenvironmentbinding_controller.go
+++ b/controllers/binding/snapshotenvironmentbinding_controller.go
@@ -29,6 +29,7 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/retry"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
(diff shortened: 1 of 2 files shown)


controllers/binding/snapshotenvironmentbinding_controller_test.go
```diff
@@ -298,11 +298,19 @@ var _ = Describe("BindingController", func() {
...
-			_, err := bindingReconciler.Reconcile(ctx, req)
-			return err
-		}).ShouldNot(BeNil())
+		err := k8sClient.Delete(ctx, hasApp)
+		Eventually(func() bool {
+			err := k8sClient.Get(ctx, types.NamespacedName{
+				Namespace: hasComp.ObjectMeta.Namespace,
+				Name:      hasComp.ObjectMeta.Name,
+			}, hasApp)
+			return err != nil
```
dirgim, line 307: Should we also check for `errors.IsNotFound(err))` here already to make this more robust?

14rcole: Fixed here and in #335

dirgim (approved):
LGTM

controllers/binding/snapshotenvironmentbinding_controller.go
```diff
@@ -82,10 +83,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
-	application, err := r.getApplicationFromSnapshotEnvironmentBinding(ctx, snapshotEnvironmentBinding)
+	var application *applicationapiv1alpha1.Application
+	err = retry.OnError(retry.DefaultRetry, func(_ error) bool { return true }, func() error {
+		application, err = r.getApplicationFromSnapshotEnvironmentBinding(ctx, snapshotEnvironmentBinding)
+		return err
+	})
 	if err != nil {
-		logger.Error(err, "Failed to get Application from the SnapshotEnvironmentBinding")
-		return ctrl.Result{}, err
+		return logger.HandleLoaderError(err, "Application", "SnapshotEnvironmentBinding")
```
MartinBasti, line 92: Logger shouldn't have responsibility to decide code flow (yes there is exception in go, for `panic` and `fatal` but that ends everything)

https://github.com/redhat-appstudio/integration-service/pull/329

MartinBasti (changes requested):

controllers/binding/snapshotenvironmentbinding_controller.go
14rcole: #342

MartinBasti (approved):
Thank you

PR #340, opened 2023-09-29
fix(STONEINTG-631): add release, deployment, GCL status conditions

dirgim:
* Add AutoRelease status condition to Snapshots to mark if Snapshot was previously auto-released
* Add DeployedtoRootEnvironments status condition to Snapshots to mark if Snapshot was deployed to all root environments
* Add AddedToGlobalCandidateList status condition to Snapshot to mark if Snapshot component was added to the Global Candidate List
* Set the above conditions in the snapshot_adapter where appropriate
* Add unit tests for the above changes

### DIFF
--- a/controllers/snapshot/snapshot_adapter.go
+++ b/controllers/snapshot/snapshot_adapter.go
@@ -334,36 +334,58 @@ TestScenarioLoop:
 // EnsureGlobalCandidateImageUpdated is an operation that ensure the ContainerImage in the Global Candidate List
 // being updated when the Snapshot passed all the integration tests
 func (a *Adapter) EnsureGlobalCandidateImageUpdated() (controller.OperationResult, error) {
-	if (a.component != nil) && gitops.HaveAppStudioTestsSucceeded(a.snapshot) && !gitops.IsSnapshotCreatedByPACPullRequestEvent(a.snapshot) {
-		patch := client.MergeFrom(a.component.DeepCopy())
-		for _, component := range a.snapshot.Spec.Components {
-			if component.Name == a.component.Name {
-				a.component.Spec.ContainerImage = component.ContainerImage
-				err := a.client.Patch(a.context, a.component, patch)
+	if a.component == nil || gitops.IsSnapshotCreatedByPACPullRequestEvent(a.snapshot) {
+		a.logger.Info("The Snapshot wasn't created for a single component push event, not updating the global candidate list.")
+		return controller.ContinueProcessing()
+	}
+	if !gitops.HaveAppStudioTestsSucceeded(a.snapshot) {
+		a.logger.Info("The Snapshot hasn't yet passed all required integration tests, not updating the global candidate list.")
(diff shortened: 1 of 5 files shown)


controllers/snapshot/snapshot_adapter.go
```diff
@@ -334,36 +334,57 @@ TestScenarioLoop:
...
+	}
+	if gitops.IsSnapshotMarkedAsAddedToGlobalCandidateList(a.snapshot) {
+		a.logger.Info("The Snapshot's component was previously added to the global candidate list, skipping adding it.")
+		return controller.ContinueProcessing()
+	}
+
+	patch := client.MergeFrom(a.component.DeepCopy())
+	for _, component := range a.snapshot.Spec.Components {
+		if component.Name == a.component.Name {
+			a.component.Spec.ContainerImage = component.ContainerImage
```
MartinBasti, line 353: `patch` variable should be within for and if statements

and probably break statement is missing here

docs/snapshot-controller.md
```diff
@@ -109,28 +113,30 @@ flowchart TD
...
   any_existing_SEB           --Yes--> update_existing_SEB
   any_existing_SEB           --No-->  create_SEB_for_non_eph_env
   update_existing_SEB        -->      encountered_error5
   create_SEB_for_non_eph_env -->      encountered_error5
   encountered_error5         --Yes--> mark_snapshot_Invalid5
-  encountered_error5         --No-->  continue_processing5
+  encountered_error5         --No-->  mark_snapshot_deployed
+  mark_snapshot_deployed     -->      continue_processing5
 
-  %% Assigning styles to nodes
```
MartinBasti, line 133: This line shouldn't be changed

MartinBasti:
Does it make sense to even run reconciliation when snapshot was deplyoed and released? Shouldn't we just stop it at beginning for everything?

controllers/snapshot/snapshot_adapter.go
dirgim: Ack, my initial idea was to only refactor this to only make the different conditions more distinguishable, but I agree that both of those changes make sense. Updated.

docs/snapshot-controller.md
dirgim: Ack, removed

dirgim:
> Does it make sense to even run reconciliation when snapshot was deplyoed and released? Shouldn't we just stop it at beginning for everything?

I considered adding a catch-all [around here](https://github.com/redhat-appstudio/integration-service/blob/aa00ca3d73ff971a0af17a787cb5501aae3e1ec6/controllers/snapshot/snapshot_controller.go#L92) that would stop reconciliation if the Snapshot already has all relevant statuses set, but decided against it as it would add another thing that would need to be maintained and changed if we add new functionalities (e.g. a new Ensure* function) in the future.

MartinBasti (approved):

jsztuka (approved):
lgtm

PR #341, opened 2023-10-02
feat(STONEINTG-599): add finalizers to Integration PLR

dheerajodha:

### DIFF
--- a/controllers/integrationpipeline/integrationpipeline_adapter.go
+++ b/controllers/integrationpipeline/integrationpipeline_adapter.go
@@ -32,7 +32,6 @@ import (
 	tektonv1beta1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
 // Adapter holds the objects needed to reconcile an integration PipelineRun.
@@ -65,12 +64,16 @@ func NewAdapter(pipelineRun *tektonv1beta1.PipelineRun, component *applicationap
 // EnsureStatusReportedInSnapshot will ensure that status of the integration test pipelines is reported to snapshot
 // to be consumed by user
 func (a *Adapter) EnsureStatusReportedInSnapshot() (controller.OperationResult, error) {
+	var pipelinerunStatus intgteststat.IntegrationTestStatus
+	var snapshot *applicationapiv1alpha1.Snapshot
+	var detail string
+	var err error
 
 	// pipelines run in parallel and have great potential to cause conflict on update
-	// thus `RetryOnConflict` is easy solution here, given the snaphost must be loaded specifically here
-	err := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+	// thus `RetryOnConflict` is easy solution here, given the snapshot must be loaded specifically here
+	err = retry.RetryOnConflict(retry.DefaultRetry, func() error {
 
-		snapshot, err := a.loader.GetSnapshotFromPipelineRun(a.client, a.context, a.pipelineRun)
+		snapshot, err = a.loader.GetSnapshotFromPipelineRun(a.client, a.context, a.pipelineRun)
 		if err != nil {
 			return err
 		}
@@ -80,11 +83,11 @@ func (a *Adapter) EnsureStatusReportedInSnapshot() (controller.OperationResult, 
 			return err
 		}
 
-		status, detail, err := GetIntegrationPipelineRunStatus(a.client, a.context, a.pipelineRun)
+		pipelinerunStatus, detail, err = GetIntegrationPipelineRunStatus(a.client, a.context, a.pipelineRun)
 		if err != nil {
 			return err
 		}
-		statuses.UpdateTestStatusIfChanged(a.pipelineRun.Labels[tekton.ScenarioNameLabel], status, detail)
+		statuses.UpdateTestStatusIfChanged(a.pipelineRun.Labels[tekton.ScenarioNameLabel], pipelinerunStatus, detail)
 		if err = statuses.UpdateTestPipelineRunName(a.pipelineRun.Labels[tekton.ScenarioNameLabel], a.pipelineRun.Name); err != nil {
 			return err
 		}
@@ -98,18 +101,13 @@ func (a *Adapter) EnsureStatusReportedInSnapshot() (controller.OperationResult, 
 		return controller.RequeueWithError(fmt.Errorf("failed to update test status in snapshot: %w", err))
 	}
 
-	if h.HasPipelineRunFinished(a.pipelineRun) {
-		// Remove the finalizer from this Integration PipelineRun
-		patch := client.MergeFrom(a.pipelineRun.DeepCopy())
-		controllerutil.RemoveFinalizer(a.pipelineRun, h.IntegrationPipelineRunFinalizer)
-		err = a.client.Patch(a.context, a.pipelineRun, patch)
+	// Remove the finalizer from Integration PLRs only if they aren't related to Snapshots created by Pull-Request event
+	// If they are related, then the statusreport controller removes the finalizers from these PLRs
+	if !gitops.IsSnapshotCreatedByPACPullRequestEvent(snapshot) && (h.HasPipelineRunFinished(a.pipelineRun) || pipelinerunStatus == intgteststat.IntegrationTestStatusDeleted) {
+		err = h.RemoveFinalizer(a.client, a.logger, a.context, a.pipelineRun, h.IntegrationPipelineRunFinalizer)
 		if err != nil {
-			return controller.RequeueWithError(fmt.Errorf("failed to remove finalizer from the Integration PipelineRun: %w", err))
+			return controller.RequeueWithError(fmt.Errorf("failed to remove the finalizer: %w", err))
 		}
-
-		a.logger.LogAuditEvent("Removed Finalizer from the Integration PipelineRun",
-			a.pipelineRun, h.LogActionUpdate,
-			"finalizer", h.IntegrationPipelineRunFinalizer)
 	}
 
 	return controller.ContinueProcessing()
@@ -160,7 +158,12 @@ func (a *Adapter) EnsureEphemeralEnvironmentsCleanedUp() (controller.OperationRe
 func GetIntegrationPipelineRunStatus(adapterClient client.Client, ctx context.Context, pipelineRun *tektonv1beta1.PipelineRun) (intgteststat.IntegrationTestStatus, string, error) {
 	// Check if the pipelineRun finished from the condition of status
 	if !h.HasPipelineRunFinished(pipelineRun) {
-		return intgteststat.IntegrationTestStatusInProgress, fmt.Sprintf("Integration test is running as pipeline run '%s'", pipelineRun.Name), nil
+		// Mark the pipelineRun's status as "Deleted" if its not finished yet and is marked for deletion (with a non-nil deletionTimestamp)
+		if pipelineRun.GetDeletionTimestamp() != nil {
+			return intgteststat.IntegrationTestStatusDeleted, fmt.Sprintf("Integration test which is running as pipeline run '%s', has been deleted", pipelineRun.Name), nil
+		} else {
+			return intgteststat.IntegrationTestStatusInProgress, fmt.Sprintf("Integration test is running as pipeline run '%s'", pipelineRun.Name), nil
+		}
 	}
 
 	outcome, err := h.GetIntegrationPipelineRunOutcome(adapterClient, ctx, pipelineRun)
--- a/controllers/integrationpipeline/integrationpipeline_adapter.go
+++ b/controllers/integrationpipeline/integrationpipeline_adapter.go
@@ -104,7 +104,7 @@ func (a *Adapter) EnsureStatusReportedInSnapshot() (controller.OperationResult, 
 	// Remove the finalizer from Integration PLRs only if they aren't related to Snapshots created by Pull-Request event
 	// If they are related, then the statusreport controller removes the finalizers from these PLRs
 	if !gitops.IsSnapshotCreatedByPACPullRequestEvent(snapshot) && (h.HasPipelineRunFinished(a.pipelineRun) || pipelinerunStatus == intgteststat.IntegrationTestStatusDeleted) {
-		err = h.RemoveFinalizer(a.client, a.logger, a.context, a.pipelineRun, h.IntegrationPipelineRunFinalizer)
+		err = h.RemoveFinalizerFromPipelineRun(a.client, a.logger, a.context, a.pipelineRun, h.IntegrationPipelineRunFinalizer)
 		if err != nil {
 			return controller.RequeueWithError(fmt.Errorf("failed to remove the finalizer: %w", err))
 		}
--- a/controllers/integrationpipeline/integrationpipeline_adapter.go
+++ b/controllers/integrationpipeline/integrationpipeline_adapter.go
@@ -32,6 +32,7 @@ import (
 	tektonv1beta1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
 	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
 // Adapter holds the objects needed to reconcile an integration PipelineRun.
@@ -96,6 +97,21 @@ func (a *Adapter) EnsureStatusReportedInSnapshot() (controller.OperationResult, 
 		a.logger.Error(err, "Failed to update pipeline status in snapshot")
 		return controller.RequeueWithError(fmt.Errorf("failed to update test status in snapshot: %w", err))
 	}
+
+	if h.HasPipelineRunFinished(a.pipelineRun) {
+		// Remove the finalizer from this Integration PipelineRun
+		patch := client.MergeFrom(a.pipelineRun.DeepCopy())
+		controllerutil.RemoveFinalizer(a.pipelineRun, h.IntegrationPipelineRunFinalizer)
+		err = a.client.Patch(a.context, a.pipelineRun, patch)
+		if err != nil {
+			return controller.RequeueWithError(fmt.Errorf("failed to remove finalizer from the Integration PipelineRun: %w", err))
+		}
+
+		a.logger.LogAuditEvent("Removed Finalizer from the Integration PipelineRun",
+			a.pipelineRun, h.LogActionUpdate,
+			"finalizer", h.IntegrationPipelineRunFinalizer)
+	}
+
 	return controller.ContinueProcessing()
 }
 
--- a/controllers/statusreport/statusreport_adapter.go
+++ b/controllers/statusreport/statusreport_adapter.go
@@ -79,6 +79,16 @@ func (a *Adapter) EnsureSnapshotTestStatusReportedToGitHub() (controller.Operati
 		}
 	}
 
+	if gitops.IsSnapshotMarkedAsPassed(a.snapshot) || gitops.IsSnapshotMarkedAsFailed(a.snapshot) || gitops.IsSnapshotMarkedAsInvalid(a.snapshot) {
+		a.logger.Info("Status has been reported successfully, now in the process of removing the finalizer")
+		// Remove the finalizer from all the Integration PLRs related to the Snapshot
+		err = loader.RemoveFinalizerFromAllIntegrationPipelineRunsOfSnapshot(a.client, a.logger, a.context, *a.snapshot, helpers.IntegrationPipelineRunFinalizer)
+		if err != nil {
+			return controller.RequeueWithError(err)
+		}
+		a.logger.Info("Successfully removed the finalizer from all the Integration PLRs related to current snapshot", "snapshot.Name", a.snapshot.Name)
+	}
+
 	return controller.ContinueProcessing()
 }
 
--- a/controllers/statusreport/statusreport_adapter.go
+++ b/controllers/statusreport/statusreport_adapter.go
@@ -82,7 +82,7 @@ func (a *Adapter) EnsureSnapshotTestStatusReportedToGitHub() (controller.Operati
 	if gitops.IsSnapshotMarkedAsPassed(a.snapshot) || gitops.IsSnapshotMarkedAsFailed(a.snapshot) || gitops.IsSnapshotMarkedAsInvalid(a.snapshot) {
 		a.logger.Info("Status has been reported successfully, now in the process of removing the finalizer")
 		// Remove the finalizer from all the Integration PLRs related to the Snapshot
-		err = loader.RemoveFinalizerFromAllIntegrationPipelineRunsOfSnapshot(a.client, a.logger, a.context, *a.snapshot, helpers.IntegrationPipelineRunFinalizer)
+		err = helpers.RemoveFinalizerFromAllIntegrationPipelineRunsOfSnapshot(a.client, a.logger, a.context, *a.snapshot, helpers.IntegrationPipelineRunFinalizer)
 		if err != nil {
 			return controller.RequeueWithError(err)
 		}
--- a/controllers/statusreport/statusreport_adapter.go
+++ b/controllers/statusreport/statusreport_adapter.go
@@ -187,7 +187,7 @@ func (a *Adapter) determineIfAllIntegrationTestsFinishedAndPassed(integrationTes
 	for _, integrationTestScenario := range *integrationTestScenarios {
 		integrationTestScenario := integrationTestScenario // G601
 		testDetails, ok := testStatuses.GetScenarioStatus(integrationTestScenario.Name)
-		if !ok || (testDetails.Status != intgteststat.IntegrationTestStatusTestPassed && testDetails.Status != intgteststat.IntegrationTestStatusTestFail) {
+		if !ok || (testDetails.Status != intgteststat.IntegrationTestStatusTestPassed && testDetails.Status != intgteststat.IntegrationTestStatusTestFail && testDetails.Status != intgteststat.IntegrationTestStatusDeleted) {
 			allIntegrationTestsFinished = false
 		}
 		if ok && testDetails.Status != intgteststat.IntegrationTestStatusTestPassed {
--- a/controllers/statusreport/statusreport_controller.go
+++ b/controllers/statusreport/statusreport_controller.go
@@ -81,8 +81,8 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 	adapter := NewAdapter(snapshot, application, logger, loader, r.Client, ctx)
 	return controller.ReconcileHandler([]controller.Operation{
-		adapter.EnsureSnapshotTestStatusReportedToGitHub,
 		adapter.EnsureSnapshotFinishedAllTests,
+		adapter.EnsureSnapshotTestStatusReportedToGitHub,
 	})
 }
 
--- a/docs/integration_pipeline_controller.md
+++ b/docs/integration_pipeline_controller.md
@@ -32,4 +32,4 @@ flowchart TD
   class predicate Amber;
   class error,requeue Red;
 
-  ```
+ ```
--- a/docs/integration_pipeline_controller.md
+++ b/docs/integration_pipeline_controller.md
@@ -10,10 +10,13 @@ flowchart TD
   classDef Green fill:#BDFFA4;
 
   %% Node definitions
-  predicate((PREDICATE: <br>Integration Pipeline <br> reconciliation))
+  predicate((PREDICATE: <br>Integration Pipeline just got<br> Started OR Finished<br> OR marked for Deletion))
   get_resources{Get pipeline, <br> component, <br> & application}
   report_status_snapshot(Report status of the test <br> into snapshot annotation <br> `test.appstudio.openshift.io/status`)
-  report_status(Report status if Snapshot was created <br> for Pull requests)
+  report_status(Report status <br> if Snapshot was created <br> for Pull requests)
+  is_snapshot_of_pr_event{Is <br> Snapshot created<br> for Pull requests?}
+  is_plr_finished_or_getting_deleted{Is <br> Integration PLR <br> finished or marked for<br> deletion?}
+  remove_finalizer(Remove <br> `test.appstudio.openshift.io/pipelinerun`<br> finalizer)
   clean_environment(Clean up ephemeral environment <br> if testing finished)
   error(Return error)
   requeue(Requeue)
@@ -24,7 +27,12 @@ flowchart TD
   predicate                                   --> clean_environment
   get_resources     --No                      --> error
   get_resources     --Yes                     --> report_status_snapshot
-  report_status_snapshot                    ----> report_status
+  report_status_snapshot                      --> is_snapshot_of_pr_event
+  is_snapshot_of_pr_event            --Yes    --> report_status
+  is_snapshot_of_pr_event            --No     --> is_plr_finished_or_getting_deleted
+  is_plr_finished_or_getting_deleted --Yes    --> remove_finalizer
+  is_plr_finished_or_getting_deleted --No     --> report_status
+  remove_finalizer                            --> report_status
   clean_environment --No                      --> requeue
   clean_environment --yes                     ---> continue
 
--- a/helpers/integration.go
+++ b/helpers/integration.go
@@ -379,16 +379,18 @@ func CleanUpEphemeralEnvironments(client client.Client, logger *IntegrationLogge
 	return nil
 }
 
-func RemoveFinalizer(adapterClient client.Client, logger IntegrationLogger, ctx context.Context, pipelineRun *tektonv1beta1.PipelineRun, finalizer string) error {
-	// Remove the finalizer from this Integration PipelineRun
+// RemoveFinalizerFromPipelineRun removes the finalizer from the PipelineRun.
+// If finalizer was not removed successfully, a non-nil error is returned.
+func RemoveFinalizerFromPipelineRun(adapterClient client.Client, logger IntegrationLogger, ctx context.Context, pipelineRun *tektonv1beta1.PipelineRun, finalizer string) error {
 	patch := client.MergeFrom(pipelineRun.DeepCopy())
-	controllerutil.RemoveFinalizer(pipelineRun, finalizer)
-	err := adapterClient.Patch(ctx, pipelineRun, patch)
-	if err != nil {
-		return fmt.Errorf("error occurred while removing finalizer from the Integration PipelineRun: %w", err)
-	}
+	if ok := controllerutil.RemoveFinalizer(pipelineRun, finalizer); ok {
+		err := adapterClient.Patch(ctx, pipelineRun, patch)
+		if err != nil {
+			return fmt.Errorf("error occurred while patching the updated PipelineRun after finalizer removal: %w", err)
+		}
 
-	logger.LogAuditEvent("Removed Finalizer from the Integration PipelineRun", pipelineRun, LogActionUpdate, "finalizer", IntegrationPipelineRunFinalizer)
+		logger.LogAuditEvent("Removed Finalizer from the Integration PipelineRun", pipelineRun, LogActionUpdate, "finalizer", finalizer)
+	}
 
 	return nil
 }
--- a/helpers/integration.go
+++ b/helpers/integration.go
@@ -379,6 +379,35 @@ func CleanUpEphemeralEnvironments(client client.Client, logger *IntegrationLogge
 	return nil
 }
 
+// RemoveFinalizerFromAllIntegrationPipelineRunsOfSnapshot fetches all the Integration
+// PipelineRuns associated with the given Snapshot. After fetching them, it removes the
+// finalizer from the PipelineRun, and returns error if any.
+func RemoveFinalizerFromAllIntegrationPipelineRunsOfSnapshot(adapterClient client.Client, logger IntegrationLogger, ctx context.Context, snapshot applicationapiv1alpha1.Snapshot, finalizer string) error {
+	integrationPipelineRuns := &tektonv1beta1.PipelineRunList{}
+	opts := []client.ListOption{
+		client.InNamespace(snapshot.Namespace),
+		client.MatchingLabels{
+			"appstudio.openshift.io/snapshot": snapshot.Name,
+		},
+	}
+
+	err := adapterClient.List(ctx, integrationPipelineRuns, opts...)
+	if err != nil {
+		return err
+	}
+
+	// Remove finalizer from each of the PipelineRuns
+	for _, pipelineRun := range integrationPipelineRuns.Items {
+		pipelineRun := pipelineRun
+		err = RemoveFinalizerFromPipelineRun(adapterClient, logger, ctx, &pipelineRun, finalizer)
+		if err != nil {
+			return err
+		}
+	}
+
+	return nil
+}
+
 // RemoveFinalizerFromPipelineRun removes the finalizer from the PipelineRun.
 // If finalizer was not removed successfully, a non-nil error is returned.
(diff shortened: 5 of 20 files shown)


controllers/integrationpipeline/integrationpipeline_adapter.go
```diff
@@ -227,6 +228,42 @@ func (a *Adapter) EnsureStatusReportedInSnapshot() (controller.OperationResult,
...
+		}
+
+		patch := client.MergeFrom(a.pipelineRun.DeepCopy())
+		controllerutil.RemoveFinalizer(a.pipelineRun, h.IntegrationPipelineRunFinalizer)
+		err = a.client.Patch(a.context, a.pipelineRun, patch)
+		if err != nil {
+			return controller.RequeueWithError(err)
+		}
+
+		a.logger.Info("Removed Finalizer from the Integration PipelineRun",
```
dirgim, line 258: I think this should be logged like an audit event - like e.g. [here](https://github.com/redhat-appstudio/integration-service/blob/8f740239643c4a1f5f2f298118a915eae2cabbfe/controllers/integrationpipeline/integrationpipeline_adapter.go#L105-L106).

dheerajodha: Thanks! Updated it with `LogActionUpdate`, I hope it's the correct one.
For future reference, when should I log an audit event?

dirgim: We should do it whenever we change resources on the cluster - so when creating, updating, or deleting them.

controllers/integrationpipeline/integrationpipeline_adapter.go
```diff
@@ -333,6 +370,18 @@ func (a *Adapter) determineIfAllIntegrationPipelinesPassed(integrationPipelineRu
 			// prevent mess in logs
 			pipelineRunOutcome.LogResults(a.logger.Logger)
 		}
+
+		// Remove the finalizer from this Integration PipelineRun
```
dirgim, line 374: I'm thinking it would make more sense to have the finalizer removed at the end of [EnsureStatusReportedInSnapshot](https://github.com/redhat-appstudio/integration-service/blob/2e6db82e809d22589bf20b69c1382580d7e66f4a/controllers/integrationpipeline/integrationpipeline_adapter.go#L194) since that is the crucial point where we record the pipeline testing outcome.

dirgim:
nit just to not forget: the commit message will need a bit more description in it.

dheerajodha:
> nit just to not forget: the commit message will need a bit more description in it.

@dirgim Thanks for the reminder! Due to some issues while building the image locally, I'll be pushing my changes here and using the image built by RHTAP to test my changes. I'll quickly update the commit afterward.

controllers/integrationpipeline/integrationpipeline_adapter.go
```diff
@@ -114,6 +117,23 @@ func (a *Adapter) EnsureStatusReportedInSnapshot() (controller.OperationResult,
 		a.logger.Error(err, "Failed to update pipeline status in snapshot")
 		return controller.RequeueWithError(fmt.Errorf("failed to update test status in snapshot: %w", err))
 	}
+
+	a.logger.Info(".........status of plr", "PLR.name", a.pipelineRun.Name, "PLR.status", pipelinerunStatus)
```
MartinBasti, line 121: If this is debugging log entry it should be removed, if not, it should be reworded, but keep on mind it may spam logs.

controllers/statusreport/statusreport_adapter.go
```diff
@@ -121,7 +121,9 @@ func (a *Adapter) EnsureSnapshotFinishedAllTests() (controller.OperationResult,
 	if !allIntegrationTestsFinished {
 		a.logger.Info("Not all required Integration PipelineRuns finished",
 			"snapshot.Name", a.snapshot.Name)
-		return controller.ContinueProcessing()
+		return controller.ContinueProcessing() // ---------> would we get stuck here indefintely because allIntegrationTestsFinished will
```
MartinBasti, line 124: Comments should be above the affected line, not bellow.

Should be this comment removed?

pkg/integrationteststatus/integration_test_status.go
```diff
@@ -33,6 +33,8 @@ const (
 	IntegrationTestStatusPending IntegrationTestStatus = iota + 1 // Pending
 	// Starting to handle an integration test scenario for a snapshot
 	IntegrationTestStatusInProgress // InProgress
+	// Integration PLR cancelled for this ITS and snapshot
+	IntegrationTestStatusCancelled
```
MartinBasti, line 37:
```suggestion
	IntegrationTestStatusCancelled  // Cancelled
```

pkg/integrationteststatus/integration_test_status.go
```diff
@@ -168,10 +172,10 @@ func (sits *SnapshotIntegrationTestStatuses) UpdateTestStatusIfChanged(scenarioN
 func (sits *SnapshotIntegrationTestStatuses) UpdateTestPipelineRunName(scenarioName string, pipelineRunName string) error {
 	detail, ok := sits.GetScenarioStatus(scenarioName)
 	if !ok {
-		return fmt.Errorf("scenario name %s not found and cannot be updated", scenarioName)
+		return fmt.Errorf("scenario name %s not found within the SnapshotIntegrationTestStatus, and cannot be updated", scenarioName)
 	}
 
-	if detail.TestPipelineRunName != pipelineRunName {
+	if detail.TestPipelineRunName != pipelineRunName && detail.Status != IntegrationTestStatusCancelled {
```
MartinBasti, line 178: Why is this needed? Even cancelled pipeline has still name

tekton/utils.go
```diff
@@ -94,6 +94,19 @@ func hasPipelineRunStateChangedToStarted(objectOld, objectNew client.Object) boo
 	return false
 }
 
+// hasPipelineRunStateChangedToDeleting returns a boolean indicating whether the PipelineRun just got marked for deletion or not.
+// If the objects passed to this function are not PipelineRuns, the function will return false.
+func hasPipelineRunStateChangedToDeleting(objectOld, objectNew client.Object) bool {
```
MartinBasti, line 99: Where is this function used?

controllers/integrationpipeline/integrationpipeline_adapter.go
```diff
@@ -161,10 +181,14 @@ func (a *Adapter) EnsureEphemeralEnvironmentsCleanedUp() (controller.OperationRe
 // GetIntegrationPipelineRunStatus checks the Tekton results for a given PipelineRun and returns status of test.
 func GetIntegrationPipelineRunStatus(adapterClient client.Client, ctx context.Context, pipelineRun *tektonv1beta1.PipelineRun) (intgteststat.IntegrationTestStatus, string, error) {
 	// Check if the pipelineRun finished from the condition of status
-	if !h.HasPipelineRunFinished(pipelineRun) {
+	if !h.HasPipelineRunFinished(pipelineRun) && pipelineRun.GetDeletionTimestamp() == nil {
 		return intgteststat.IntegrationTestStatusInProgress, fmt.Sprintf("Integration test is running as pipeline run '%s'", pipelineRun.Name), nil
 	}
 
+	if !h.HasPipelineRunFinished(pipelineRun) && pipelineRun.GetDeletionTimestamp() != nil {
```
MartinBasti, line 188: it may deserve comment what non-nil DeletionTimestamp means here

controllers/integrationpipeline/integrationpipeline_adapter.go
```diff
@@ -114,6 +117,23 @@ func (a *Adapter) EnsureStatusReportedInSnapshot() (controller.OperationResult,
 		a.logger.Error(err, "Failed to update pipeline status in snapshot")
 		return controller.RequeueWithError(fmt.Errorf("failed to update test status in snapshot: %w", err))
 	}
+
+	a.logger.Info(".........status of plr", "PLR.name", a.pipelineRun.Name, "PLR.status", pipelinerunStatus)
+
+	if h.HasPipelineRunFinished(a.pipelineRun) || (pipelinerunStatus.IsAIntegrationTestStatus() && pipelinerunStatus == intgteststat.IntegrationTestStatusCancelled) {
```
MartinBasti, line 123: `IsAIntegrationTestStatus()` why do we need this check here (or respectivelly only here)?

MartinBasti (changes requested):
go fmt failed in CI

controllers/integrationpipeline/integrationpipeline_adapter.go
dheerajodha: sorry, this is a result of me not pushing my latest changes to the PR.

controllers/statusreport/statusreport_adapter.go
dheerajodha: sorry, this is a result of me not pushing my latest changes to the PR.

pkg/integrationteststatus/integration_test_status.go
dheerajodha: this is a result of me not pushing my latest changes to the PR.

pkg/integrationteststatus/integration_test_status.go
```diff
@@ -158,6 +161,7 @@ func (sits *SnapshotIntegrationTestStatuses) UpdateTestStatusIfChanged(scenarioN
 	if detail.Details != details {
 		detail.Details = details
 		detail.LastUpdateTime = timestamp
+		sits.statuses[scenarioName] = detail
```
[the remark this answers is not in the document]
dheerajodha: I thought that we needed to update the sits object with the latest data, but I can see now that the `detail` variable is a pointer so the data was already updated. Removing this line.

tekton/utils.go
dheerajodha: within the `tekton/predicates.go` file.

pkg/integrationteststatus/integration_test_status.go
dheerajodha: but if we add the name of a canceled PLR, it won't retrigger the creation of a new PLR [here](https://github.com/redhat-appstudio/integration-service/blob/2880904d7988527ce384ad14a5f5c0ece9c06e97/controllers/snapshot/snapshot_adapter.go#L127)

controllers/integrationpipeline/integrationpipeline_adapter.go
dheerajodha: I was thinking about nil-pointer exceptions, but you're right, this is not needed here because `RequeueWithError` will ensure that we don't reach at this line with a `nil` pipelinerunStatus. Removing this.

controllers/integrationpipeline/integrationpipeline_adapter.go
dheerajodha: Added a comment

controllers/integrationpipeline/integrationpipeline_adapter.go
MartinBasti: how exactly will solve it null pointer error? it's built on type int, it can only validate if number is from defined range in enum.

`nil.IsAIntegrationTestStatus()` wouldn't even compile

pkg/integrationteststatus/integration_test_status.go
dheerajodha: alternatively, I can update the if statement within the snapshot_adapter:
https://github.com/redhat-appstudio/integration-service/blob/2880904d7988527ce384ad14a5f5c0ece9c06e97/controllers/snapshot/snapshot_adapter.go#L127

to this:

```
if ok && integrationTestScenarioStatus.TestPipelineRunName != "" && integrationTestScenarioStatus.Status != intgteststat.IntegrationTestStatusCancelled
```

wdyt?

controllers/integrationpipeline/integrationpipeline_adapter.go
dheerajodha: you're right, thanks!

pkg/integrationteststatus/integration_test_status.go
MartinBasti: How exactly it would help? Status would have already recorded pipeline name from InProgress status.

Do we even want to restart cancelled pipeline?

MartinBasti: If you want to restart pipeline, it must be handled [here](https://github.com/redhat-appstudio/integration-service/blob/2880904d7988527ce384ad14a5f5c0ece9c06e97/controllers/snapshot/snapshot_adapter.go#L127)
```
if pipeline is cancelled:
    recreate pipeline
```

pkg/integrationteststatus/integration_test_status.go
```diff
@@ -163,18 +166,24 @@ func (sits *SnapshotIntegrationTestStatuses) UpdateTestStatusIfChanged(scenarioN
...
 	}
 
 	if detail.TestPipelineRunName != pipelineRunName {
 		detail.TestPipelineRunName = pipelineRunName
 		sits.dirty = true
 	}
+
+	if detail.Status == IntegrationTestStatusCancelled {
+		// Empty TestPipelineRunName field so that it triggers creation of a new PipelineRun for this Scenario
+		detail.TestPipelineRunName = ""
```
MartinBasti, line 184: I don't think we should erase pipeline here, it reports current status, logic to decide if cancelled pipeline should be replaced must be somewhere else.

pkg/integrationteststatus/integration_test_status.go
```diff
@@ -33,6 +33,8 @@ const (
 	IntegrationTestStatusPending IntegrationTestStatus = iota + 1 // Pending
 	// Starting to handle an integration test scenario for a snapshot
 	IntegrationTestStatusInProgress // InProgress
+	// Integration PLR cancelled for this ITS and snapshot
+	IntegrationTestStatusCancelled // Cancelled
```
MartinBasti, line 37: And because you added new state, we need followup story to update status reporters in pull requests

pkg/integrationteststatus/integration_test_status.go
MartinBasti: I was wrong, it makes sense to remove it, but we shouldn't call status cancelled, because cancelled is something else. Cancelled pipeline is still valid pipeline that's kept in k8s.

Deleted would be more specific

https://tekton.dev/docs/pipelines/pipelineruns/#pipelinerun-status

dirgim: Tbh I'm not sure if we do want to restart cancelled pipeline runs - there are definitely scenarios where we wouldn't want to do that automatically, e.g. when the test uses an ephemeral environment that costs money/resources and the user cancelled it because of an issue in testing where it got stuck.

In any case I don't think that restarting cancelled integration pipelines should be handled as part of this ticket.

dheerajodha: not sure if I understand how it is different from what I proposed above.

> Status would have already recorded pipeline name from InProgress status.

Yes, and that's why with the addition of an extra condition to ensure status is not "Cancelled" (described above), the controller would create a new PLR, am I missing something?

MartinBasti: I had still old PR before force push in memory and it was not there, so my comments are off now, I had to start all over again.

MartinBasti: > I was wrong, it makes sense to remove it

Taking back what I said, it should keep pipelinerun name, it express the status that pipeline with this particular name has been deleted.

dheerajodha: Thanks guys for the explanation.

Do we then eventually, expect to see this when a user deletes a PLR:
* Mark the snapshot as "Failed", and also update its annotation ofcourse
* remove the finalizer if PLR was not created by a PR event -> "Deleted" PLR gets deleted.

dirgim: Yes, that sounds correct to me.

dheerajodha: Perfect

dheerajodha: Updated the logic to reflect the conclusion of this discussion.

Now when an Intg PLR is deleted while it is running, its status within the Snapshot's status annotation gets updated to "Deleted", and Snapshot gets marked as Failed and Finished, and the Finalizer is removed from the PLR.

controllers/integrationpipeline/integrationpipeline_adapter.go
```diff
@@ -114,6 +118,14 @@ func (a *Adapter) EnsureStatusReportedInSnapshot() (controller.OperationResult,
 		a.logger.Error(err, "Failed to update pipeline status in snapshot")
 		return controller.RequeueWithError(fmt.Errorf("failed to update test status in snapshot: %w", err))
 	}
+
+	if !gitops.IsSnapshotCreatedByPACPullRequestEvent(snapshot) && (h.HasPipelineRunFinished(a.pipelineRun) || pipelinerunStatus == intgteststat.IntegrationTestStatusDeleted) {
```
MartinBasti, line 122:
```suggestion
	if !gitops.IsSnapshotCreatedByPACPullRequestEvent(snapshot) && ((h.HasPipelineRunFinished(a.pipelineRun) || pipelinerunStatus == intgteststat.IntegrationTestStatusDeleted)) {
```

It should be like this right? Only non pull request snapshots.

Also it may deserve a comment why we care only about PAC snapshots here and where are the other cases handled

controllers/integrationpipeline/integrationpipeline_adapter.go
MartinBasti: looking at the code it could be
```
if PLR hasnt fnished:
   if deletion_timestamp:
       return StatusDeleted
   else:
       return  StatusRunning
```

instead of separate conditions with `&&`

helpers/integration.go
```diff
@@ -325,6 +326,19 @@ func HasPipelineRunSucceeded(object client.Object) bool {
 	return false
 }
 
+func GetPipelineWithName(pipelineRunName string, namespace string, k8sClient client.Client, ctx context.Context) (*tektonv1beta1.PipelineRun, error) {
```
MartinBasti, line 329: missing docstring

status/reporters.go
```diff
@@ -679,13 +690,24 @@ func (r *GitHubReporter) ReportStatusForSnapshot(k8sClient client.Client, ctx co
...
+			}
+
 			commitStatus, err := r.createCommitStatusAdapterForSnapshot(snapshot, integrationTestStatusDetail, owner, repo, sha)
 			if err != nil {
 				logger.Error(err, "failed to create CommitStatusAdapter for snapshot",
 					"snapshot.NameSpace", snapshot.Namespace, "snapshot.Name", snapshot.Name)
 				return err
 			}
 
+			err = helpers.RemoveFinalizer(k8sClient, *logger, ctx, pipelineRun)
```
MartinBasti, line 706: I don't think that finalizer should be removed here
This function may fail later and we would still need the pipeline details on requeue.

also from SRP, this function shouldn't care about finalizers, it shold only care about updating status in GH. I think removal of finalizers should be done as separate function, once final GH report has been done.

status/reporters.go
```diff
@@ -625,13 +625,24 @@ func (r *GitHubReporter) ReportStatusForSnapshot(k8sClient client.Client, ctx co
...
+			}
+
 			checkRun, err := r.createCheckRunAdapterForSnapshot(snapshot, integrationTestStatusDetail, owner, repo, sha)
 			if err != nil {
 				logger.Error(err, "failed to create checkRunAdapter for snapshot",
 					"snapshot.NameSpace", snapshot.Namespace, "snapshot.Name", snapshot.Name)
 				return err
 			}
 
+			err = helpers.RemoveFinalizer(k8sClient, *logger, ctx, pipelineRun)
```
MartinBasti, line 641: How do we know that we can already remove finalizers here, if I'm not mistaken, this function also reports other statuses as well, so you mey remove finalizer when PLR is still in progress.

tekton/predicates.go
```diff
@@ -38,7 +38,7 @@ func IntegrationPipelineRunPredicate() predicate.Predicate {
 		},
 		UpdateFunc: func(e event.UpdateEvent) bool {
 			return IsIntegrationPipelineRun(e.ObjectNew) &&
-				(hasPipelineRunStateChangedToStarted(e.ObjectOld, e.ObjectNew) || hasPipelineRunStateChangedToFinished(e.ObjectOld, e.ObjectNew))
+				(hasPipelineRunStateChangedToStarted(e.ObjectOld, e.ObjectNew) || hasPipelineRunStateChangedToFinished(e.ObjectOld, e.ObjectNew) || hasPipelineRunStateChangedToDeleting(e.ObjectOld, e.ObjectNew))
```
MartinBasti, line 41:
```suggestion
				(hasPipelineRunStateChangedToStarted(e.ObjectOld, e.ObjectNew) || 
                                hasPipelineRunStateChangedToFinished(e.ObjectOld, e.ObjectNew) ||
				hasPipelineRunStateChangedToDeleting(e.ObjectOld, e.ObjectNew))
```

controllers/integrationpipeline/integrationpipeline_adapter.go
MartinBasti: or is it done in such way to handle tests deletion also for PR integration tests?

tekton/predicates.go
dheerajodha: Moved them into new lines and then `go fmt ./...` added its own magic.

status/reporters.go
dheerajodha: Thanks for the info, this logic now gets called after all the status reporting is completed.

controllers/integrationpipeline/integrationpipeline_adapter.go
dheerajodha: I don't think adding those round brackets would make any difference since it's already inside a round bracket, right?
This part of the code would only handle non-PR Intg PLRs. I've added a comment to address your concern.

pkg/integrationteststatus/integration_test_status.go
dheerajodha: We agreed to not erase the pipelineRun's names in case they get "Deleted".
Instead, we would simply mark the Snapshot as failed and not re-trigger a new Intg PLR on its behalf.

pkg/integrationteststatus/integration_test_status.go
dheerajodha: Created: [STONEINTG-668](https://issues.redhat.com/browse/STONEINTG-668)

loader/integration.go
```diff
@@ -55,3 +56,32 @@ func GetLatestPipelineRunForSnapshotAndScenario(adapterClient client.Client, ctx
...
+		}
+
+		// Remove finalizer from each of the PipelineRuns
+		for _, pipelineRun := range *integrationPipelineRuns {
+			err = helpers.RemoveFinalizer(adapterClient, logger, ctx, &pipelineRun, finalizer)
+			if err != nil {
+				return err
+			}
+		}
+	}
```
dheerajodha, line 84: The cyclomatic complexity of this function is high, any suggestions to improve it?

helpers/integration.go
```diff
@@ -377,3 +393,17 @@ func CleanUpEphemeralEnvironments(client client.Client, logger *IntegrationLogge
 	logger.LogAuditEvent("Ephemeral environment is deleted and its owning SnapshotEnvironmentBinding is in the process of being deleted", env, LogActionDelete)
 	return nil
 }
+
+func RemoveFinalizer(adapterClient client.Client, logger IntegrationLogger, ctx context.Context, pipelineRun *tektonv1beta1.PipelineRun, finalizer string) error {
+	// Remove the finalizer from this Integration PipelineRun
+	patch := client.MergeFrom(pipelineRun.DeepCopy())
+	controllerutil.RemoveFinalizer(pipelineRun, finalizer)
```
dirgim, line 400: The `controllerutil.RemoveFinalizer()` returns a boolean denoting if the finalizer was found and removed, do we want to use that to decide if we need to attempt to patch the resource?

loader/integration.go
dirgim: The integration pipelineRuns should also have the [Snapshot label](https://github.com/redhat-appstudio/integration-service/blob/b3f521c408d147ecc4f8811c9112172b31c0c405/tekton/integration_pipeline.go#L57-L58), we could fetch all of them using that label instead of looping throught individual integrationTestScenarios.

helpers/integration.go
dheerajodha: Thanks! updated to use the returned boolean from this method before calling Patch.

loader/integration.go
dheerajodha: Thanks, updated the logic to use the Snapshot label to fetch the related Intg PLRs instead.

controllers/integrationpipeline/integrationpipeline_adapter.go
MartinBasti: if we care only only about PAC runs comment won't save it, it is time to learn about precendce of operators in go.

https://www.tutorialspoint.com/go/go_operators_precedence.htm

```
func main() {
    //                #isPR #finished # deleted                 
	fmt.Println("%v", true && true || true)
	fmt.Println("%v", false && true || true)
	fmt.Println("%v", true && (true || true))
	fmt.Println("%v", false && (true || true))
}
```
https://goplay.tools/snippet/IUawbPTWH4a

MartinBasti: ah nevermind, I'm totally blind and stupid :D

tekton/predicates.go
MartinBasti: okay :(

controllers/statusreport/statusreport_adapter.go
```diff
@@ -79,6 +79,16 @@ func (a *Adapter) EnsureSnapshotTestStatusReportedToGitHub() (controller.Operati
 		}
 	}
 
+	if gitops.IsSnapshotMarkedAsPassed(a.snapshot) || gitops.IsSnapshotMarkedAsFailed(a.snapshot) {
+		a.logger.Info("Status has been reported successfully, now in the process of removing the finalizer")
```
MartinBasti, line 83: we are inconsistent with logging here, for non-PAC we are not logging anything, for PAC PR, we are logging start and end of process.

MartinBasti: or maybe it's fine as it is

loader/integration.go
```diff
@@ -55,3 +56,31 @@ func GetLatestPipelineRunForSnapshotAndScenario(adapterClient client.Client, ctx
...
+
+	err := adapterClient.List(ctx, integrationPipelineRuns, opts...)
+	if err != nil {
+		return err
+	}
+
+	// Remove finalizer from each of the PipelineRuns
+	for _, pipelineRun := range integrationPipelineRuns.Items {
+		err = helpers.RemoveFinalizerFromPipelineRun(adapterClient, logger, ctx, &pipelineRun, finalizer)
```
MartinBasti, line 79:
```suggestion
	for _, pipelineRun := range integrationPipelineRuns.Items {
	        pipelineRun := pipelineRun
		err = helpers.RemoveFinalizerFromPipelineRun(adapterClient, logger, ctx, &pipelineRun, finalizer)
```

```
[/github/workspace/loader/integration.go:79] - G601 (CWE-118): Implicit memory aliasing in for loop. (Confidence: MEDIUM, Severity: MEDIUM)
    78: 	for _, pipelineRun := range integrationPipelineRuns.Items {
  > 79: 		err = helpers.RemoveFinalizerFromPipelineRun(adapterClient, logger, ctx, &pipelineRun, finalizer)
    80: 		if err != nil 
```

loader/integration.go
```diff
@@ -55,3 +56,31 @@ func GetLatestPipelineRunForSnapshotAndScenario(adapterClient client.Client, ctx
 
 	return nil, err
 }
+
+// RemoveFinalizerFromAllIntegrationPipelineRunsOfSnapshot fetches all the Integration
+// PipelineRuns associated with the given Snapshot and each of the IntegrationTestScenarios.
+// After fetching them, it removes the finalizer from the PipelineRun, and returns error if any.
+func RemoveFinalizerFromAllIntegrationPipelineRunsOfSnapshot(adapterClient client.Client, logger helpers.IntegrationLogger, ctx context.Context, snapshot applicationapiv1alpha1.Snapshot, finalizer string) error {
```
MartinBasti, line 63: This func is in loaders module, but it updates things not loading them. IMO it shouldn't be here

MartinBasti: Should stay in helpers IMO, it will also remove new dependency on helpers module here

loader/integration.go
dheerajodha: Thank you, updated. I'll keep this in mind.

loader/integration.go
dheerajodha: You're right, I've now moved it to helpers/integration.go file.

dheerajodha:
I have added an [additional commit](https://github.com/redhat-appstudio/integration-service/pull/341/commits/8aa33dc0391a012b5552d505298c52cefc11f30d) that deals with the problem I noticed while testing my changes locally. I've tried to describe them within the commit description.

controllers/statusreport/statusreport_adapter.go
```diff
@@ -79,6 +79,16 @@ func (a *Adapter) EnsureSnapshotTestStatusReportedToGitHub() (controller.Operati
 		}
 	}
 
+	if gitops.IsSnapshotMarkedAsPassed(a.snapshot) || gitops.IsSnapshotMarkedAsFailed(a.snapshot) {
```
MartinBasti, line 82: I realized that snapshot can be also marked as invalid

controllers/statusreport/statusreport_controller.go
```diff
@@ -81,8 +81,8 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 
 	adapter := NewAdapter(snapshot, application, logger, loader, r.Client, ctx)
 	return controller.ReconcileHandler([]controller.Operation{
-		adapter.EnsureSnapshotTestStatusReportedToGitHub,
 		adapter.EnsureSnapshotFinishedAllTests,
+		adapter.EnsureSnapshotTestStatusReportedToGitHub,
```
MartinBasti, line 85: Should be also diagram updated?

controllers/statusreport/statusreport_adapter.go
dheerajodha: Thanks for catching that. Now that I think about it, along with Invalid, the Snapshot's Integration status is also marked as Error but we won't consider it here because the Error state is not a final state and the Snapshot _could_ move out of it. Is that correct?

controllers/integrationpipeline/integrationpipeline_adapter.go
dheerajodha: I really appreciate the time you take to explain things to me

helpers/integration.go
```diff
@@ -325,6 +326,21 @@ func HasPipelineRunSucceeded(object client.Object) bool {
 	return false
 }
 
+// GetPipelineWithName returns a PipelineRun based on the name and namespace
+// If PipelineRun is not found, a non-nil error is returned
+func GetPipelineWithName(pipelineRunName string, namespace string, k8sClient client.Client, ctx context.Context) (*tektonv1beta1.PipelineRun, error) {
```
dirgim, line 331: It looks like this function is not used, can we remove it?

docs/integration_pipeline_controller.md
```diff
@@ -24,12 +27,17 @@ flowchart TD
   predicate                                   --> clean_environment
   get_resources     --No                      --> error
   get_resources     --Yes                     --> report_status_snapshot
-  report_status_snapshot                    ----> report_status
+  report_status_snapshot                      --> is_snapshot_of_pr_event
+  is_snapshot_of_pr_event            --Yes    --> report_status
+  is_snapshot_of_pr_event            --No     --> is_plr_finished_or_getting_deleted
+  is_plr_finished_or_getting_deleted --Yes    --> remove_finalizer
+  is_plr_finished_or_getting_deleted --No     --> report_status
```
hongweiliu17, line 34: I don't understand why there are 3 paths for `Report status if Snapshot was created for Pull requests` in [dragram](https://github.com/dheerajodha/integration-service/blob/STONEINTG-599/docs/integration_pipeline_controller.md), since the right two paths are from `No` of `Is Snapshot created for Pull requests?`, could explain it? Thanks.

dheerajodha: Those 3 paths explain [this if condition](https://github.com/redhat-appstudio/integration-service/pull/341/files#diff-d2d01b4dc526c804c076232db16ae083ed3381a14d78fc2bbbab8fd4bad59077R106).
* Branch 1: when `gitops.IsSnapshotCreatedByPACPullRequestEvent(snapshot)` returns true.
* Branch 2: when it returns False AND both of the `"h.HasPipelineRunFinished(a.pipelineRun)", "pipelinerunStatus == intgteststat.IntegrationTestStatusDeleted"` function returns False.
* Branch 3: when it returns False AND either of the `"h.HasPipelineRunFinished(a.pipelineRun)", "pipelinerunStatus == intgteststat.IntegrationTestStatusDeleted"` function returns True.

Maybe we can make this part of the diagram shorter in some way...

controllers/integrationpipeline/integrationpipeline_adapter.go
```diff
@@ -144,7 +158,12 @@ func (a *Adapter) EnsureEphemeralEnvironmentsCleanedUp() (controller.OperationRe
 func GetIntegrationPipelineRunStatus(adapterClient client.Client, ctx context.Context, pipelineRun *tektonv1beta1.PipelineRun) (intgteststat.IntegrationTestStatus, string, error) {
 	// Check if the pipelineRun finished from the condition of status
 	if !h.HasPipelineRunFinished(pipelineRun) {
-		return intgteststat.IntegrationTestStatusInProgress, fmt.Sprintf("Integration test is running as pipeline run '%s'", pipelineRun.Name), nil
+		// Mark the pipelineRun's status as "Deleted" if its not finished yet and is marked for deletion (with a non-nil deletionTimestamp)
+		if pipelineRun.GetDeletionTimestamp() != nil {
+			return intgteststat.IntegrationTestStatusDeleted, fmt.Sprintf("Integration test which is running as pipeline run '%s', has been deleted", pipelineRun.Name), nil
```
hongweiliu17, line 163: If user deletes one old PLR, I think this will cause the `Deleted` status written into snapshot annotation, this will overwrites the correct status.

hongweiliu17: Or once a PLR is marked as deleted by integration-service, it won't appear in PLR list?

hongweiliu17: Oh, my bad, old PLR is invisible for users, they operate on RHTAP UI.

dirgim: I think this logic is there to handle the weird behavior that happens when the `pipelineRun` is still running (hasn't finished) and something attempts to delete it. It shouldn't take into account old PLRs that had finished.

dheerajodha: `Deleted` status is only added to the PLRs that are still running and have been marked for deletion with a non-nil timestamp (this can happen when a user deletes the PLR manually).

In the case of an old PLR, if a user manually deletes it, there can be 2 scenarios based on the current state of this old PLR:
* If this old PLR is in a "Running" state:
    * Now when a user deletes it, based on the current logic, the controller would change its state within the status annotation from "InProgress" to "Deleted".
    * I think that's what we would ideally want to happen.

* If this old PLR is Finished:
    * Now when a user deletes it, based on the current logic, the controller would not do anything.
    * Why? Because at this point the status annotation would contain the correct status for this PLR like "TestPassed" (or failed, etc).
    * This action would just simply delete the PLR from the cluster since it would not contain any finalizer as it was removed right when the PLR was finished in the first place.

PLease correct me if I'm wrong, thanks!

helpers/integration.go
```diff
@@ -377,3 +393,50 @@ func CleanUpEphemeralEnvironments(client client.Client, logger *IntegrationLogge
...
+
+	return nil
+}
+
+// RemoveFinalizerFromPipelineRun removes the finalizer from the PipelineRun.
+// If finalizer was not removed successfully, a non-nil error is returned.
+func RemoveFinalizerFromPipelineRun(adapterClient client.Client, logger IntegrationLogger, ctx context.Context, pipelineRun *tektonv1beta1.PipelineRun, finalizer string) error {
+	patch := client.MergeFrom(pipelineRun.DeepCopy())
+	if ok := controllerutil.RemoveFinalizer(pipelineRun, finalizer); !ok {
+		return fmt.Errorf("error occurred while removing the finalizer: %s, during updating the PipelineRun's list of finalizers: %q", finalizer, pipelineRun.GetFinalizers())
```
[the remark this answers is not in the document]
dheerajodha: Updated, also added tests for it [here](https://github.com/redhat-appstudio/integration-service/pull/341/files#diff-08825d1599dc68963e3bf82ff1008bbf5016cfd2aa08c29d43a2c14704ed4399R357)

helpers/integration.go
dheerajodha: Thanks! removed it

controllers/statusreport/statusreport_controller.go
hongweiliu17: What is the reason exchange the order of them? Run `EnsureSnapshotFinishedAllTests` firstly and then `EnsureSnapshotTestStatusReportedToGitHub`? But `EnsureSnapshotFinishedAllTests` need the test status written by `EnsureSnapshotTestStatusReportedToGitHub`, is my understanding wrong?

controllers/integrationpipeline/integrationpipeline_adapter.go
hongweiliu17: I see now. Thank you all for the explanations!

dirgim (approved):
LGTM, nice work

docs/integration_pipeline_controller.md
hongweiliu17: I see your meaning now. Here `Report status if Snapshot was created for Pull requests` means reporting status to GitHub right? But it is not done by integration PLR controller, so I think it should not be displayed here.

controllers/statusreport/statusreport_controller.go
dheerajodha: > What is the reason exchange the order of them?

Before the order was reversed:
* the finalizer from Integration PLRs
  (related to the pull_request event) is not removed,
  even after the Snapshot gets marked as Passed/Failed.
* As we know, the EnsureSnapshotFinishedAllTests function
  marks the Snapshot as Passed/Failed.
* And since the EnsureSnapshotFinishedAllTests function
  is called AFTER EnsureSnapshotTestStatusReportedToGitHub
  function, the finalizer is never removed.

After the order was reversed:
* The finalizer gets removed successfully.

> But EnsureSnapshotFinishedAllTests need the test status written by EnsureSnapshotTestStatusReportedToGitHub

As per my understanding, the `EnsureSnapshotFinishedAllTests` function needs the latest test status data from the Snapshot's annotation. And `EnsureSnapshotTestStatusReportedToGitHub` function doesn't make any changes to that annotation, so there's no dependency between them in that way.

docs/integration_pipeline_controller.md
dheerajodha: I agree with you!
Since this change is outside the scope of this PR (as it was not added here), I think it should be tackled within a new PR (if you don't mind).

controllers/statusreport/statusreport_adapter.go
MartinBasti: Technically yes, but it may be also forever

MartinBasti (approved):

controllers/statusreport/statusreport_controller.go
hongweiliu17: Yes, you are right. I am wrong. Thanks!

hongweiliu17 (approved):
LGTM

dheerajodha:
Just for the record: Tested this PR locally on the following 4 cases

**Case 1:** Integration PLR finishes and the finalizer gets removed.
image

**Case 2:** The finalizer gets removed from Integration PLR when gets deleted while it is in the "Running" state.
image

**Case 3:** Test status gets reported to CheckRun, and the finalizers get removed.
image

**Case 4:** Test status gets reported to the Pull request's comment, and the finalizers get removed.
image

and it worked as expected. Thanks everyone for the reviews!

docs/integration_pipeline_controller.md
dheerajodha: Opened a PR to fix this: https://github.com/redhat-appstudio/integration-service/pull/395

PR #346, opened 2023-10-05
refactor(STONEINTG-539): EnsureCreationOfEnvironments

MartinBasti:
Making `EnsureCreationOfEnvironments` reusable for single test creation.

`ensureEphemeralEnvironmentForScenarioExists` contains refactored part that allows to ensure existence of ephemeral env for single scenario

### DIFF
--- a/controllers/snapshot/snapshot_adapter.go
+++ b/controllers/snapshot/snapshot_adapter.go
@@ -161,7 +161,7 @@ func (a *Adapter) EnsureAllIntegrationTestPipelinesExist() (controller.Operation
 	return controller.ContinueProcessing()
 }
 
-// EnsureCreationOfEnvironment makes sure that all envrionments that were requested via
+// EnsureCreationOfEnvironment makes sure that all ephemeral envrionments that were requested via
 // IntegrationTestScenarios get created, in case that environment is already created, provides
 // a message about this fact
 func (a *Adapter) EnsureCreationOfEnvironment() (controller.OperationResult, error) {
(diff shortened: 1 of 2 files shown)


MartinBasti:
This is the most important part, I have few more improvements to do, but I'll create a new PRs for it, to not block any follow up work

controllers/snapshot/snapshot_adapter.go
```diff
@@ -499,6 +407,115 @@ func (a *Adapter) EnsureSnapshotEnvironmentBindingExist() (controller.OperationR
...
+	}
+
+	// envirnoment binding for snapshot
+	binding, err := a.loader.FindExistingSnapshotEnvironmentBinding(a.client, a.context, a.application, environment)
+	if err != nil {
+		a.logger.Error(err, "Failed to find snapshotEnvironmentBinding associated with environment", "environment.Name", environment.Name)
+		return nil, fmt.Errorf("failed to find SnapshotEnvironmentBinding associated with environment: %w", err)
+	}
+	if binding == nil {
+		// create bindging and add scenario name to label of binding
```
dirgim, line 493: tiny typo: `bindging` -> `binding`

MartinBasti: yeah, I just copy pasted it, will change

controllers/snapshot/snapshot_suite_test.go
```diff
@@ -99,6 +100,8 @@ var _ = BeforeSuite(func() {
 	k8sClient = k8sManager.GetClient()
 	go func() {
 		defer GinkgoRecover()
+		Expect(cache.SetupBindingApplicationCache(k8sManager)).To(Succeed())
+		Expect(cache.SetupBindingEnvironmentCache(k8sManager)).To(Succeed())
```
dirgim, line 104: It should be possible to prevent using cache (which causes issues in low-resource environments) in tests by using mockLoaders for these resources in the adapter test itself, unless the issue is somewhere else?

MartinBasti: yeah, I found out how to mock it, but I have to rewrite mocker implementation, I need to return nil and it's not compatible

```
	value := ctx.Value(contextKey)
	if value == nil {
		panic("Mocked data not found in the context")
	}
```

MartinBasti: nevermind, the issue was somewhere else

dirgim: Cool, I can see that it works now, thanks

dirgim (approved):
LGTM

Josh-Everett (approved):
lgtm nice refactor

PR #347, opened 2023-10-06
feat(statusreporting): add feature flag

MartinBasti:
Adding feature flag FEATURE_STATUS_REPORTING_ENABLED to enable feature for reporting tests statuses into pull requestes. Featue is not ready yet, thus is disabled by default.

### DIFF
--- a/controllers/statusreport/statusreport_adapter.go
+++ b/controllers/statusreport/statusreport_adapter.go
@@ -18,6 +18,7 @@ package statusreport
 
 import (
 	"context"
+	"os"
 
 	applicationapiv1alpha1 "github.com/redhat-appstudio/application-api/api/v1alpha1"
 	"github.com/redhat-appstudio/integration-service/helpers"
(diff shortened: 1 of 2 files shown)


dirgim (approved):
LGTM

MartinBasti:
Interestingly, I disabled behavior without updating unittests and not a single test failed.

Basically we don't have tests for this feature. I'll update tests to use feature flag.

jsztuka (approved):
lgtm

dirgim (approved):
+1 after update

hongweiliu17:
> Interestingly, I disabled behavior without updating unittests and not a single test failed.
> 
> Basically we don't have tests for this feature. I'll update tests to use feature flag.

We use mocked `ReportStatusForSnapshot` in test [here](https://github.com/redhat-appstudio/integration-service/blob/statusreport/controllers/statusreport/statusreport_adapter_test.go#L56) because it can't really call github api.

PR #357, opened 2023-10-10
feat(stoneintg-600): Add finalizers to build pipeline runs

14rcole:
Because of the introduction of more aggressive resource pruning in RHTAP, the integration service must add finalizers to build pipeline runs when they are created.  This will ensure that the integration service can create snapshots of for successful builds.  Once the snapshot is created the finalizer will be removed so that the resource can be cleaned up.

### DIFF
--- a/controllers/buildpipeline/buildpipeline_adapter.go
+++ b/controllers/buildpipeline/buildpipeline_adapter.go
@@ -33,6 +33,7 @@ import (
 	"github.com/redhat-appstudio/operator-toolkit/metadata"
 	tektonv1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
 // Adapter holds the objects needed to reconcile a build PipelineRun.
@@ -64,12 +65,16 @@ func NewAdapter(pipelineRun *tektonv1.PipelineRun, component *applicationapiv1al
 // to the build PipelineRun being processed exists. Otherwise, it will create a new pipeline Snapshot.
 func (a *Adapter) EnsureSnapshotExists() (controller.OperationResult, error) {
 	if !h.HasPipelineRunSucceeded(a.pipelineRun) {
+		if h.HasPipelineRunFinished(a.pipelineRun) {
+			// The pipeline run  has failed
+			return a.removeFinalizerAndContinueProcessing()
+		}
+		// The build pipeline run has not finished yet
 		return controller.ContinueProcessing()
 	}
 
 	if a.component == nil {
-		a.logger.Info("The build pipelineRun does not have any component associated with it, will not create a new Snapshot.")
-		return controller.ContinueProcessing()
+		return a.removeFinalizerAndContinueProcessing()
 	}
 
 	isLatest, err := a.isLatestSucceededBuildPipelineRun()
@@ -81,7 +86,7 @@ func (a *Adapter) EnsureSnapshotExists() (controller.OperationResult, error) {
 		// this prevents deploying older pipeline run over new deployment
 		a.logger.Info("The pipelineRun is not the latest successful build pipelineRun for the component, skipping creation of a new Snapshot ",
 			"component.Name", a.component.Name)
-		return controller.ContinueProcessing()
+		return a.removeFinalizerAndContinueProcessing()
 	}
 
 	expectedSnapshot, err := a.prepareSnapshotForPipelineRun(a.pipelineRun, a.component, a.application)
@@ -108,7 +113,7 @@ func (a *Adapter) EnsureSnapshotExists() (controller.OperationResult, error) {
 				return controller.RequeueWithError(err)
 			}
 		}
-		return controller.ContinueProcessing()
+		return a.removeFinalizerAndContinueProcessing()
 	}
 
 	err = a.client.Create(a.context, expectedSnapshot)
@@ -126,6 +131,20 @@ func (a *Adapter) EnsureSnapshotExists() (controller.OperationResult, error) {
 	if err != nil {
 		a.logger.Error(err, "Failed to update the build pipelineRun with new annotations",
 			"pipelineRun.Name", a.pipelineRun.Name)
+		return controller.RequeueWithError(err)
+	}
+
+	return a.removeFinalizerAndContinueProcessing()
+}
+
+func (a *Adapter) EnsurePipelineIsFinalized() (controller.OperationResult, error) {
+	if h.HasPipelineRunFinished(a.pipelineRun) || controllerutil.ContainsFinalizer(a.pipelineRun, h.IntegrationPipelineRunFinalizer) {
+		return controller.ContinueProcessing()
+	}
+
+	err := h.AddFinalizerToPipelineRun(a.client, a.logger, a.context, a.pipelineRun, h.IntegrationPipelineRunFinalizer)
+	if err != nil {
+		a.logger.Error(err, fmt.Sprintf("Could not add finalizer %s to build pipeline %s", h.IntegrationPipelineRunFinalizer, a.pipelineRun.Name))
 		return controller.RequeueWithError(err)
 	}
 
@@ -273,4 +292,13 @@ func (a *Adapter) annotateBuildPipelineRunWithSnapshot(pipelineRun *tektonv1.Pip
 	a.logger.LogAuditEvent("Updated build pipelineRun", pipelineRun, h.LogActionUpdate,
 		"snapshot.Name", snapshot.Name)
 	return pipelineRun, nil
+}
+
+func (a *Adapter) removeFinalizerAndContinueProcessing() (controller.OperationResult, error) {
+	if !controllerutil.ContainsFinalizer(a.pipelineRun, h.IntegrationPipelineRunFinalizer) {
+		return controller.ContinueProcessing()
+	}
+
+	err := h.RemoveFinalizerFromPipelineRun(a.client, a.logger, a.context, a.pipelineRun, h.IntegrationPipelineRunFinalizer)
+	return controller.RequeueOnErrorOrContinue(err)
 }
--- a/controllers/buildpipeline/buildpipeline_adapter_test.go
+++ b/controllers/buildpipeline/buildpipeline_adapter_test.go
@@ -29,6 +29,8 @@ import (
 	v1 "knative.dev/pkg/apis/duck/v1"
 
 	k8serrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/strings/slices"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
@@ -758,6 +760,44 @@ var _ = Describe("Pipeline Adapter", Ordered, func() {
 			Expect(allSnapshots).NotTo(BeNil())
 			existingSnapshot := gitops.FindMatchingSnapshot(hasApp, allSnapshots, hasSnapshot)
 			Expect(existingSnapshot.Name).To(Equal(hasSnapshot.Name))
+		})
+	})
+
+	When("A new Build pipelineRun is created", func() {
+		BeforeAll(func() {
+			adapter = NewAdapter(buildPipelineRun, hasComp, hasApp, logger, loader.NewMockLoader(), k8sClient, ctx)
+		})
+		It("can add and remove finalizers from the pipelineRun", func() {
+			// Mark build PLR as incomplete
+			buildPipelineRun.Status.Conditions = nil
+			Expect(k8sClient.Status().Update(ctx, buildPipelineRun)).Should(Succeed())
+
+			// Ensure PLR does not have finalizer
+			existingBuildPLR := new(tektonv1.PipelineRun)
+			err := k8sClient.Get(ctx, types.NamespacedName{
+				Namespace: buildPipelineRun.ObjectMeta.Namespace,
+				Name:      buildPipelineRun.ObjectMeta.Name,
+			}, existingBuildPLR)
+			Expect(err).To(BeNil())
+			Expect(existingBuildPLR.ObjectMeta.Finalizers).To(BeNil())
+
+			// Add Finalizer to PLR
+			result, err := adapter.EnsurePipelineIsFinalized()
+			Expect(result.CancelRequest).To(BeFalse())
+			Expect(err).To(BeNil())
+
+			// Ensure PLR has finalizer
+			Eventually(func() bool {
+				return slices.Contains(buildPipelineRun.ObjectMeta.Finalizers, helpers.IntegrationPipelineRunFinalizer)
+			}, time.Second*10).Should(BeTrue())
+
+			// Remove finalizer from PLR
+			result, err = adapter.removeFinalizerAndContinueProcessing()
+			Expect(result.CancelRequest).To(BeFalse())
+			Expect(err).To(BeNil())
+
+			// Ensure PLR does not have finalizer
+			Expect(buildPipelineRun.ObjectMeta.Finalizers).NotTo(ContainElement("build.appstudio.openshift.io/pipelinerun"))
 		})
 	})
 
--- a/controllers/buildpipeline/buildpipeline_controller.go
+++ b/controllers/buildpipeline/buildpipeline_controller.go
@@ -19,6 +19,7 @@ package buildpipeline
 import (
 	"context"
 	"fmt"
+
 	"github.com/redhat-appstudio/integration-service/cache"
 
 	"github.com/go-logr/logr"
@@ -106,12 +107,14 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	adapter := NewAdapter(pipelineRun, component, application, logger, loader, r.Client, ctx)
 
 	return controller.ReconcileHandler([]controller.Operation{
+		adapter.EnsurePipelineIsFinalized,
 		adapter.EnsureSnapshotExists,
 	})
 }
 
 // AdapterInterface is an interface defining all the operations that should be defined in an Integration adapter.
 type AdapterInterface interface {
+	EnsurePipelineIsFinalized() (controller.OperationResult, error)
 	EnsureSnapshotExists() (controller.OperationResult, error)
 }
 
@@ -137,6 +140,8 @@ func setupControllerWithManager(manager ctrl.Manager, controller *Reconciler) er
 	return ctrl.NewControllerManagedBy(manager).
 		For(&tektonv1.PipelineRun{}).
 		WithEventFilter(predicate.Or(
-			tekton.BuildPipelineRunSignedAndSucceededPredicate())).
+			tekton.BuildPipelineRunSignedAndSucceededPredicate(),
+			tekton.BuildPipelineRunFailedPredicate(),
+			tekton.BuildPipelineRunCreatedPredicate())).
 		Complete(controller)
 }
--- a/docs/build_pipeline_controller.md
+++ b/docs/build_pipeline_controller.md
@@ -11,16 +11,27 @@ flowchart TD
 
   %% Node definitions
 predicate((PREDICATE: <br> Filter events related to <br> PipelineRuns <br> proccessed by Chains <br> that have <br> succeeded))
-get_pipeline_run{Pipeline found?}
+new_pipeline_run{Pipeline created?}
+get_pipeline_run{Pipeline updated?}
+failed_pipeline_run{Pipeline failed?}
+finalizer_exists{Does the finalizer already exist?}
 retrieve_associated_entity(Retrieve the entity <br> component/application)
 determine_snapshot{Does a snapshot exist?}
 create_snapshot(Gather Application components<br> Add new component  <br> Create snapshot)
 annotate_pipelineRun(Annotate pipeline with <br> name of Snapshot)
+add_finalizer(Add finalizer to build PLR)
+remove_finalizer(Remove finalizer from build PLR)
 error[Return error]
 continue[Continue processing]
 
 %% Node connections
 predicate                        --> get_pipeline_run
+predicate                       -->  new_pipeline_run
+predicate                       -->  failed_pipeline_run
+new_pipeline_run           --Yes-->  finalizer_exists
+finalizer_exists           --No-->   add_finalizer
+add_finalizer                    --> continue
+failed_pipeline_run           --Yes --> remove_finalizer
 get_pipeline_run           --Yes --> retrieve_associated_entity
 get_pipeline_run           --No  --> error
 retrieve_associated_entity --No  --> error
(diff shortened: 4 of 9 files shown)


controllers/buildpipeline/buildpipeline_adapter.go
```diff
@@ -129,6 +133,30 @@ func (a *Adapter) EnsureSnapshotExists() (controller.OperationResult, error) {
 		return controller.RequeueWithError(err)
 	}
 
+	return a.removeFinalizerAndContinueProcessing()
+}
+
+func (a *Adapter) EnsurePipelineIsFinalized() (controller.OperationResult, error) {
+	// Is this necessary in a function that runs only when the pipeline is created?
```
dirgim, line 140: This function also runs when the pipelineRun succeeds on account of the [existing predicate](https://github.com/redhat-appstudio/integration-service/blob/d2b50365a889d09028b6324ee8c958aa818889b3/controllers/buildpipeline/buildpipeline_controller.go#L128).
Additionally, all Ensure* functions need to be idempotent because all CRs are automatically reconciled on service start up without regard for predicate filters.

controllers/buildpipeline/buildpipeline_adapter_test.go
```diff
@@ -290,7 +293,21 @@ var _ = Describe("Pipeline Adapter", Ordered, func() {
 	})
 
 	AfterEach(func() {
-		err := k8sClient.Delete(ctx, buildPipelineRun)
+		patch := client.MergeFrom(buildPipelineRun.DeepCopy())
+		_ = controllerutil.RemoveFinalizer(buildPipelineRun, "build.appstudio.openshift.io/pipelinerun")
+		err := k8sClient.Patch(ctx, buildPipelineRun, patch)
```
14rcole, line 298: I'm running into an issue here.  It looks like the finalizer isn't being removed, which is preventing this deletion from being completed.  As far as I can tell, lines 296-298 are the same (minus error handling) as the ones I've added in `removeFinalizerAndContinueProcessing()`.

controllers/buildpipeline/buildpipeline_adapter.go
```diff
@@ -129,6 +133,29 @@ func (a *Adapter) EnsureSnapshotExists() (controller.OperationResult, error) {
...
+}
+
+func (a *Adapter) EnsurePipelineIsFinalized() (controller.OperationResult, error) {
+	if !h.HasPipelineRunFinished(a.pipelineRun) || controllerutil.ContainsFinalizer(a.pipelineRun, finalizer) {
+		return controller.ContinueProcessing()
+	}
+
+	err := retry.OnError(retry.DefaultRetry, func(_ error) bool { return true }, func() error {
+		patch := client.MergeFrom(a.pipelineRun.DeepCopy())
+		if !controllerutil.AddFinalizer(a.pipelineRun, finalizer) {
```
dirgim, line 146: The `controllerutil.AddFinalizer` function returns a boolean indicating if the finalizer was added or if it was already there ([docs](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#AddFinalizer)), so we shouldn't treat  `false` as an error here, but as an indication if we want to patch the resource on the cluster or not.

14rcole: From the docs:
>  It returns an indication of whether it updated the object's list of finalizers. 
Since this function should only run right after the resource is created, I think it's safe to assume `false` means "we couldn't add the finalizer" rather than "we didn't add the finalizer because it was already there."  Do you think the latter could occur?

dirgim: It could occur, because we cannot expect the predicates to prevent from reconciling the resource more than one time after creation. 
E.g. if the service restarts, it will try to reconcile all existing CRs (disregarding the predicates), so this could end up trying to add the finalizer after it was already added.

dirgim: Since it's not possible to differentiate between the `controllerutil.AddFinalizer` call failing to add the finalizer and it not being added because it's already present on the resource, I propose adopting the approach similar to [this one](https://github.com/redhat-appstudio/release-service/blob/9a0f08573e4ca3f0d5deda2cbda2575b7e7093bd/controllers/release/adapter.go#L131-L150) where we would check if the `a.pipelineRun` doesn't already have the finalizer set on it before doing anything further.

controllers/buildpipeline/buildpipeline_adapter.go
```diff
@@ -129,6 +142,37 @@ func (a *Adapter) EnsureSnapshotExists() (controller.OperationResult, error) {
...
 
+	// Remove before merge
+	a.logger.Info("Reached end of EnsureSnapshotExists()")
+	return a.removeFinalizerAndContinueProcessing()
+}
+
+func (a *Adapter) EnsurePipelineIsFinalized() (controller.OperationResult, error) {
+	// Remove before merge
+	a.logger.Info(fmt.Sprintf("Adding finalizer '%s' to build pipeline run '%s'", finalizer, a.pipelineRun.Name))
+	if !h.HasPipelineRunFinished(a.pipelineRun) || controllerutil.ContainsFinalizer(a.pipelineRun, finalizer) {
```
dheerajodha, line 153:
```suggestion
	if h.HasPipelineRunFinished(a.pipelineRun) || controllerutil.ContainsFinalizer(a.pipelineRun, finalizer) {
```
To allow the finalizer to be added on the build PLR right after it gets created

controllers/buildpipeline/buildpipeline_adapter.go
```diff
@@ -274,3 +318,39 @@ func (a *Adapter) annotateBuildPipelineRunWithSnapshot(pipelineRun *tektonv1beta
...
+	}
+	if !controllerutil.ContainsFinalizer(a.pipelineRun, finalizer) {
+		// Remove before merge
+		a.logger.Info("build PLR does not contain finalizer")
+		return controller.ContinueProcessing()
+	}
+
+	err := retry.OnError(retry.DefaultRetry, func(_ error) bool { return true }, func() error {
+		patch := client.MergeFrom(a.pipelineRun.DeepCopy())
+		if !controllerutil.RemoveFinalizer(a.pipelineRun, finalizer) {
```
dirgim, line 339: Now that STONEINTG-599 was merged, we can probably use [RemoveFinalizerFromPipelineRun](https://github.com/redhat-appstudio/integration-service/blob/main/helpers/integration.go#L411-L425)() here for commonality.

controllers/buildpipeline/buildpipeline_adapter.go
```diff
@@ -32,9 +32,13 @@ import (
 	"github.com/redhat-appstudio/operator-toolkit/controller"
 	"github.com/redhat-appstudio/operator-toolkit/metadata"
 	tektonv1beta1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
+const finalizer string = "build.appstudio.openshift.io/pipelinerun"
```
dirgim, line 40: Can we define this in the same way that [IntegrationPipelineRunFinalizer](https://github.com/redhat-appstudio/integration-service/blob/96dc306bf01e6596e2518c2eec29dcccc0201b47/helpers/finalizers.go#L20) is defined?

controllers/buildpipeline/buildpipeline_adapter.go
```diff
@@ -129,6 +132,30 @@ func (a *Adapter) EnsureSnapshotExists() (controller.OperationResult, error) {
...
+}
+
+func (a *Adapter) EnsurePipelineIsFinalized() (controller.OperationResult, error) {
+	if h.HasPipelineRunFinished(a.pipelineRun) || controllerutil.ContainsFinalizer(a.pipelineRun, h.BuildPipelineRunFinalizer) {
+		return controller.ContinueProcessing()
+	}
+
+	err := retry.OnError(retry.DefaultRetry, func(_ error) bool { return true }, func() error {
+		patch := client.MergeFrom(a.pipelineRun.DeepCopy())
+		if !controllerutil.AddFinalizer(a.pipelineRun, h.BuildPipelineRunFinalizer) {
```
dirgim, line 145: I still think we should use [RemoveFinalizerFromPipelineRun](https://github.com/redhat-appstudio/integration-service/blob/main/helpers/integration.go#L411-L425) here instead, especially since the current code has potential for infinite reconciliations if the finalizer is already set.

14rcole: Why would we use `RemoveFinalizerFromPipelineRun()` to add a finalizer?

14rcole: What I can do is verify that the finalizer doesn't already exist.  I can also create a function in helpers called `AddFinalizerToPipelineRun()` that handles all of this.

dirgim: Sorry, my bad, I thought we had an AddFinalizer function already and didn't check what I was c/p-ing. You're right, creating an `AddFinalizerToPipelineRun()` is the way to go.

controllers/buildpipeline/buildpipeline_adapter.go
```diff
@@ -64,12 +65,12 @@ func NewAdapter(pipelineRun *tektonv1beta1.PipelineRun, component *applicationap
 // to the build PipelineRun being processed exists. Otherwise, it will create a new pipeline Snapshot.
 func (a *Adapter) EnsureSnapshotExists() (controller.OperationResult, error) {
 	if !h.HasPipelineRunSucceeded(a.pipelineRun) {
+		a.logger.Info("Build pipeline has not succeeded")
```
MartinBasti, line 68: Do we need this log here? It may be repeated multiple times, not sure if it gives any value.

helpers/finalizers.go
```diff
@@ -18,3 +18,6 @@ package helpers
 
 // IntegrationPipelineRunFinalizer is the finalizer name to be added to the Integration PipelineRuns
 const IntegrationPipelineRunFinalizer string = "test.appstudio.openshift.io/pipelinerun"
+
+// BuildPipeineRunFinalizer is the finalizer name to be added to Build PipeineRuns
+const BuildPipelineRunFinalizer string = "build.appstudio.openshift.io/pipelinerun"
```
MartinBasti, line 23: Shouldn't this represent that integration service added finalizer? Something like `test.appstudio.openshift.io/SMTH`

MartinBasti (commented):
Minor issues, looks good

helpers/finalizers.go
14rcole: I'm not sure, I added the label that was suggested in the ticket.  I'm not familiar with labeling conventions in RHTAP.  @dirgim since you created the ticket do you have any input?

controllers/buildpipeline/buildpipeline_adapter.go
14rcole: No, that was something I added for debugging that I forgot to remove.

controllers/buildpipeline/buildpipeline_controller.go
```diff
@@ -125,6 +127,6 @@ func setupControllerWithManager(manager ctrl.Manager, controller *Reconciler) er
 	return ctrl.NewControllerManagedBy(manager).
 		For(&tektonv1beta1.PipelineRun{}).
 		WithEventFilter(predicate.Or(
-			tekton.BuildPipelineRunSignedAndSucceededPredicate())).
+			tekton.BuildPipelineRunSignedAndSucceededPredicate(), tekton.BuildPipelineRunCreatedPredicate())).
```
dheerajodha, line 130:
```suggestion
			tekton.BuildPipelineRunSignedAndSucceededPredicate(), 
			tekton.BuildPipelineRunCreatedPredicate())).
```
optional

helpers/integration.go
```diff
@@ -423,3 +423,19 @@ func RemoveFinalizerFromPipelineRun(adapterClient client.Client, logger Integrat
...
+// If finalizer was not added successfully, a non-nil error is returned.
+func AddFinalizerToPipelineRun(adapterClient client.Client, logger IntegrationLogger, ctx context.Context, pipelineRun *tektonv1beta1.PipelineRun, finalizer string) error {
+	patch := client.MergeFrom(pipelineRun.DeepCopy())
+	if ok := controllerutil.AddFinalizer(pipelineRun, finalizer); ok {
+		err := adapterClient.Patch(ctx, pipelineRun, patch)
+		if err != nil {
+			return fmt.Errorf("error occurred while patching the updated PipelineRun after finalizer removal: %w", err)
+		}
+
+		logger.LogAuditEvent("Removed Finalizer from the Integration PipelineRun", pipelineRun, LogActionUpdate, "finalizer", finalizer)
```
dheerajodha, line 437:
```suggestion
		logger.LogAuditEvent("Added Finalizer to the Integration PipelineRun", pipelineRun, LogActionUpdate, "finalizer", finalizer)
```

controllers/buildpipeline/buildpipeline_adapter.go
```diff
@@ -274,3 +289,12 @@ func (a *Adapter) annotateBuildPipelineRunWithSnapshot(pipelineRun *tektonv1beta
 		"snapshot.Name", snapshot.Name)
 	return pipelineRun, nil
 }
+
+func (a *Adapter) removeFinalizerAndContinueProcessing() (controller.OperationResult, error) {
+	if !h.HasPipelineRunFinished(a.pipelineRun) || !controllerutil.ContainsFinalizer(a.pipelineRun, h.BuildPipelineRunFinalizer) {
```
dheerajodha, line 294: do we need this check (`!h.HasPipelineRunFinished(a.pipelineRun)`) within this `if` condition?

Since `removeFinalizerAndContinueProcessing()` is only called within `EnsureSnapshotExists` function that only processes [those PLRs that succeeded](https://github.com/redhat-appstudio/integration-service/pull/357/files#diff-887649ea3498038f46836e7fb616f809a6e09232f7fdec4b833f8d7aa8044db3R67), so this function would always be processing finished (and succeeded) PLRs.

dheerajodha:
nit just to not forget: the diagram for buildpipeline controller will need an update as well.

jsztuka:
Looks good overall, once the required changes are applied, will add +1

helpers/integration.go
hongweiliu17: `Added Finalizer to the Integration PipelineRun`->`Added Finalizer to the build pipelineRun`

controllers/buildpipeline/buildpipeline_adapter.go
hongweiliu17, line 68: Should we also add some code to remove finalizer from the failed build plr? Otherwise I guess the failed plr with a finalizer won't be deleted.

controllers/buildpipeline/buildpipeline_adapter.go
14rcole: We do right now because `removeFinalizerAndContinueProcessing()` is also called if the pipeline has not succeeded yet (I did a find/replace on controllerutils.ContinueProcessing()).  The better solution would be to revert that particular function call and then not rely on this if statement.  It also makes the unit tests simpler.  I'll go change that.

controllers/buildpipeline/buildpipeline_adapter.go
14rcole: You're right, I think we should call `removeFinalizerAndContinueProcessing()` here.  To prevent a case where we prematurely remove the finalizer, we'll need to call `HasPipelineRunFinished` either below line 66 or at the beginning of `removeFinalizerAndContinueProcessing()`.  Do you have an opinion as to which we should do?

@dheerajodha we had discussed something similar in another comment so feel free to weigh in here too.

hongweiliu17: Calling `HasPipelineRunFinished` below line 66 looks good to me.

dheerajodha: maybe something like this?
```
if !h.HasPipelineRunSucceeded(a.pipelineRun) {
          if HasPipelineRunFinished(a.pipelineRun) {
                 // This means that a.pipelineRun has "Failed"
                return removeFinalizerAndContinueProcessing()
          }
          a.logger.Info("Build pipeline has not succeeded YET")
}
```

controllers/buildpipeline/buildpipeline_controller.go
```diff
@@ -137,6 +139,7 @@ func setupControllerWithManager(manager ctrl.Manager, controller *Reconciler) er
 	return ctrl.NewControllerManagedBy(manager).
 		For(&tektonv1beta1.PipelineRun{}).
 		WithEventFilter(predicate.Or(
-			tekton.BuildPipelineRunSignedAndSucceededPredicate())).
+			tekton.BuildPipelineRunSignedAndSucceededPredicate(),
+			tekton.BuildPipelineRunCreatedPredicate())).
```
dirgim, line 143: I think that we'll also need to catch failed build pipelineRuns in order to remove the finalizer, and the current predicates do not catch those events.

14rcole: You're right, my push failed and I wasn't paying attention.  I've just pushed a change with a new predicate called `BuildPipelineRunFailedPredicate()`.

controllers/buildpipeline/buildpipeline_adapter.go
```diff
@@ -64,12 +65,16 @@ func NewAdapter(pipelineRun *tektonv1beta1.PipelineRun, component *applicationap
 // to the build PipelineRun being processed exists. Otherwise, it will create a new pipeline Snapshot.
 func (a *Adapter) EnsureSnapshotExists() (controller.OperationResult, error) {
 	if !h.HasPipelineRunSucceeded(a.pipelineRun) {
+		if h.HasPipelineRunFinished(a.pipelineRun) {
+			// The pipeline run  has failed
+			return a.removeFinalizerAndContinueProcessing()
+		}
+		a.logger.Info("Build pipeline has not succeeded yet")
```
dirgim, line 72: nit: I think this can be dropped, might have been mentioned before.

dirgim (approved):
LGTM aside the nit

helpers/integration.go
```diff
@@ -423,3 +423,19 @@ func RemoveFinalizerFromPipelineRun(adapterClient client.Client, logger Integrat
 
 	return nil
 }
+
+// AddFinalizerToPipelineRun adds the finalizer to the PipelineRun.
+// If finalizer was not added successfully, a non-nil error is returned.
+func AddFinalizerToPipelineRun(adapterClient client.Client, logger IntegrationLogger, ctx context.Context, pipelineRun *tektonv1beta1.PipelineRun, finalizer string) error {
```
dheerajodha, line 429: The PR checks are failing on this line with the error: `undefined: tektonv1beta1 (compile)`.
I think we should be using the new import `tektonv1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"` instead.

docs/build_pipeline_controller.md
```diff
@@ -11,16 +11,23 @@ flowchart TD
 
   %% Node definitions
 predicate((PREDICATE: <br> Filter events related to <br> PipelineRuns <br> proccessed by Chains <br> that have <br> succeeded))
```
dheerajodha, line 13: We would now need to update the predicted to also mention that the controller also picks up failed build PLRs.

docs/build_pipeline_controller.md
```diff
@@ -11,16 +11,23 @@ flowchart TD
...
-get_pipeline_run{Pipeline found?}
+new_pipeline_run{Pipeline created?}
+get_pipeline_run{Pipeline updated?}
```
dheerajodha, line 15: 1. I think the controller doesn't respond to PLR updates, instead, it responds to build PLRs that succeeded.

2. Maybe we need one more branch that describes the "Pipeline failed?" scenario.

docs/build_pipeline_controller.md
```diff
@@ -11,16 +11,23 @@ flowchart TD
...
+finalizer_exists{Does the finalizer already exist?}
 retrieve_associated_entity(Retrieve the entity <br> component/application)
 determine_snapshot{Does a snapshot exist?}
 create_snapshot(Gather Application components<br> Add new component  <br> Create snapshot)
```
dheerajodha, line 19: we might also need a new node that describes the controller removing the finalizer as well.
Ideally after the nodes:
* retrieve_associated_entity --No  -->
* determine_snapshot         --Yes --> 
* (somewhere after) create_snapshot
* retrieve_associated_entity  --No -->

tekton/predicates.go
```diff
@@ -65,3 +65,43 @@ func BuildPipelineRunSignedAndSucceededPredicate() predicate.Predicate {
 		},
 	}
 }
+
+// BuildPipelineRunFailedPredicate returns a predicate which filters out all objects except Build
+// PipelineRuns which have finished and have failed.
+func BuildPipelineRunFailedPredicate() predicate.Predicate {
```
dheerajodha, line 71: can we please have a unit test for this predicate as well?

helpers/integration.go
14rcole: Fixed.  Also fixed a couple of other outdated references to `tektonv1beta1` that I had also missed earlier.

docs/build_pipeline_controller.md
14rcole: I had also forgotten to add a `remove finalizer` step to the existing diagram.  So I added that step and linked it to both the build PLR succeeded control flow and the build PLR failed control flow.

docs/build_pipeline_controller.md
14rcole: Ack, mentioned above

docs/build_pipeline_controller.md
dheerajodha: Thanks!
Just curious, is the text `Pipeline updated?` correct, or it should be `Pipeline succeeded?` instead?

docs/build_pipeline_controller.md
dheerajodha: I think you maybe missed updating this

dheerajodha (approved):
LGTM aside from some small doc-related changes

controllers/buildpipeline/buildpipeline_adapter.go
```diff
@@ -64,12 +65,16 @@ func NewAdapter(pipelineRun *tektonv1.PipelineRun, component *applicationapiv1al
 // to the build PipelineRun being processed exists. Otherwise, it will create a new pipeline Snapshot.
 func (a *Adapter) EnsureSnapshotExists() (controller.OperationResult, error) {
 	if !h.HasPipelineRunSucceeded(a.pipelineRun) {
+		if h.HasPipelineRunFinished(a.pipelineRun) {
+			// The pipeline run  has failed
+			return a.removeFinalizerAndContinueProcessing()
+		}
+		// The build pipeline run has not succeeded YET
```
hongweiliu17, line 72:
```suggestion
		// The build pipeline run has not finished YET
```

hongweiliu17 (approved):
LGTM

jsztuka (approved):
lgtm

PR #362, opened 2023-10-11
refactor(STONEINTG-539): abstract detection of ephemeral env

MartinBasti:
Instead of copy pasting checking for non-empty environemnt definition in scenario spec, abstract it into function with clear semantincs, that make code much easier to read.

Also easy unittest for free.


dirgim (approved):
LGTM

PR #371, opened 2023-10-18
fix(STONEINTG-583): add status.conditions to existing ITS

dirgim:
* If the ITS was already migrated from v1alpha1, it might not have its status.conditions set at all. This change adds it as part of the scenario controller reconciliation

### DIFF
--- a/controllers/scenario/scenario_adapter.go
+++ b/controllers/scenario/scenario_adapter.go
@@ -58,7 +58,19 @@ func NewAdapter(application *applicationapiv1alpha1.Application, scenario *v1bet
 // in case it is, set its owner reference
 func (a *Adapter) EnsureCreatedScenarioIsValid() (controller.OperationResult, error) {
 
-	// First check if application exists or not
+	// First check if status conditions are set to prevent a known issue with incorrectly migrated v1alpha1 scenarios
+	if a.scenario.Status.Conditions == nil {
+		a.logger.Info("The scenario doesn't have status.conditions set correctly, adding them.")
+		patch := client.MergeFrom(a.scenario.DeepCopy())
+		a.scenario.Status.Conditions = []metav1.Condition{}
+		err := a.client.Status().Patch(a.context, a.scenario, patch)
+		if err != nil {
+			a.logger.Error(err, "Failed to add Scenario status condition")
+			return controller.RequeueWithError(err)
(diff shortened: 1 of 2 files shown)


controllers/scenario/scenario_adapter.go
```diff
@@ -58,7 +58,19 @@ func NewAdapter(application *applicationapiv1alpha1.Application, scenario *v1bet
...
 
-	// First check if application exists or not
+	// First check if status conditions are set to prevent a known issue with incorrectly migrated v1alpha1 scenarios
+	if a.scenario.Status.Conditions == nil {
+		a.logger.Info("The scenario doesn't have status.conditions set correctly, adding them.")
+		patch := client.MergeFrom(a.scenario.DeepCopy())
+		a.scenario.Status.Conditions = []metav1.Condition{}
+		err := a.client.Status().Patch(a.context, a.scenario, patch)
+		if err != nil {
+			a.logger.Error(err, "Failed to update Scenario")
```
MartinBasti, line 68:
```suggestion
			a.logger.Error(err, "Failed to add Scenario status condition")
```

dirgim: Ack, updated.

controllers/scenario/scenario_adapter.go
```diff
@@ -58,7 +58,19 @@ func NewAdapter(application *applicationapiv1alpha1.Application, scenario *v1bet
 // in case it is, set its owner reference
 func (a *Adapter) EnsureCreatedScenarioIsValid() (controller.OperationResult, error) {
 
-	// First check if application exists or not
+	// First check if status conditions are set to prevent a known issue with incorrectly migrated v1alpha1 scenarios
+	if a.scenario.Status.Conditions == nil {
```
dheerajodha, line 62: Are we sure that currently-present-scenarios with this issue, at least have the `.status` field defined?

dirgim: Yes, status is a subresource so it always gets created, but it's empty for migrated resources.
Here's an example from the staging cluster:
image

dheerajodha: Thanks for the explanation and screenshot!

dheerajodha (approved):

PR #381, opened 2023-10-25
fix(STONEINTG-642) stop reconciliation if env is not found for ITS

hongweiliu17:
* stop reconciliation if the env defined in ITS doesn't exist

### DIFF
--- a/controllers/snapshot/snapshot_adapter.go
+++ b/controllers/snapshot/snapshot_adapter.go
@@ -22,7 +22,6 @@ import (
 	"reflect"
 	"strings"
 
-	"k8s.io/apimachinery/pkg/types"
 	ctrl "sigs.k8s.io/controller-runtime"
 
 	applicationapiv1alpha1 "github.com/redhat-appstudio/application-api/api/v1alpha1"
@@ -220,6 +219,11 @@ func (a *Adapter) EnsureCreationOfEphemeralEnvironments() (controller.OperationR
 
 		_, err = a.ensureEphemeralEnvironmentForScenarioExists(&integrationTestScenario, testStatuses, components, allEnvironments)
 		if err != nil {
+			if h.IsEnvironmentNotInNamespaceError(err) {
+				a.logger.Error(err, "environment doesn't exist in the same namespace as integrationScenario at all, try again after creating environment",
+					"integrationTestScenario.Name", integrationTestScenario.Name)
+				return controller.StopProcessing()
+			}
 			return controller.RequeueWithError(fmt.Errorf("failed to ensure existence of the environment for scenario %s: %w", integrationTestScenario.Name, err))
 		}
 	}
@@ -426,28 +430,6 @@ func (a *Adapter) getExistingEnvironmentForScenario(integrationTestScenario *v1b
 	return nil, false
 }
 
-// createEnvironmentForScenario creates new epehemeral environment for scenario by copying existing environment
-func (a *Adapter) createEnvironmentForScenario(integrationTestScenario *v1beta1.IntegrationTestScenario) (*applicationapiv1alpha1.Environment, error) {
-	//get the existing environment according to environment name from integrationTestScenario
-	existingEnv, err := a.getEnvironmentFromIntegrationTestScenario(integrationTestScenario)
-	if err != nil {
-		a.logger.Error(err, "Failed to find the env defined in integrationTestScenario",
-			"integrationTestScenario.Namespace", integrationTestScenario.Namespace,
-			"integrationTestScenario.Name", integrationTestScenario.Name)
-		return nil, err
-	}
-
-	//create an ephemeral copy env of existing environment
-	copyEnv, err := a.createCopyOfExistingEnvironment(existingEnv, a.snapshot.Namespace, integrationTestScenario, a.snapshot, a.application)
-
-	if err != nil {
-		a.logger.Error(err, "Copying of environment failed")
-		return nil, fmt.Errorf("copying of environment failed: %w", err)
-	}
-
-	return copyEnv, nil
-}
-
 // createEnvironmentBindingForScenario creates SnapshotEnvironmentBinding for the given test scenario and snapshot
 func (a *Adapter) createEnvironmentBindingForScenario(integrationTestScenario *v1beta1.IntegrationTestScenario, environment *applicationapiv1alpha1.Environment, components *[]applicationapiv1alpha1.Component) (*applicationapiv1alpha1.SnapshotEnvironmentBinding, error) {
 	scenarioLabelAndKey := map[string]string{gitops.SnapshotTestScenarioLabel: integrationTestScenario.Name}
@@ -477,8 +459,20 @@ func (a *Adapter) ensureEphemeralEnvironmentForScenarioExists(integrationTestSce
 	environment, ok := a.getExistingEnvironmentForScenario(integrationTestScenario, allEnvironments)
 	if !ok {
 		// environment doesn't exist, we have to create a new one
-		environment, err = a.createEnvironmentForScenario(integrationTestScenario)
+		existingEnv := findEnvironmentFromAllEnvironmentsByName(allEnvironments, integrationTestScenario.Spec.Environment.Name)
+		if existingEnv == nil {
+			err := h.NewEnvironmentNotInNamespaceError(integrationTestScenario.Spec.Environment.Name, integrationTestScenario.Namespace)
+			testStatuses.UpdateTestStatusIfChanged(
+				integrationTestScenario.Name, intgteststat.IntegrationTestStatusEnvironmentProvisionError,
+				fmt.Sprintf("Creation of copied ephemeral environment failed: %s. Try again after creating environment.", err))
+			a.writeIntegrationTestStatusAtError(testStatuses)
+			return nil, err
+		}
+		//create an ephemeral copy env of existing environment
+		environment, err = a.createCopyOfExistingEnvironment(existingEnv, a.snapshot.Namespace, integrationTestScenario, a.snapshot, a.application)
+
 		if err != nil {
+			a.logger.Error(err, "Copying of environment failed")
 			testStatuses.UpdateTestStatusIfChanged(
 				integrationTestScenario.Name, intgteststat.IntegrationTestStatusEnvironmentProvisionError,
 				fmt.Sprintf("Creation of ephemeral environment failed: %s", err))
@@ -778,23 +772,15 @@ func (a *Adapter) CreateDeploymentTargetClaimForEnvironment(namespace string, de
 	return deploymentTargetClaim, nil
 }
 
-// getEnvironmentFromIntegrationTestScenario looks for already existing environment, if it exists it is returned, if not, nil is returned then together with
-// information about what went wrong
-func (a *Adapter) getEnvironmentFromIntegrationTestScenario(integrationTestScenario *v1beta1.IntegrationTestScenario) (*applicationapiv1alpha1.Environment, error) {
-	existingEnv := &applicationapiv1alpha1.Environment{}
-
-	err := a.client.Get(a.context, types.NamespacedName{
-		Namespace: a.application.Namespace,
-		Name:      integrationTestScenario.Spec.Environment.Name,
-	}, existingEnv)
-
-	if err != nil {
-		a.logger.Info("Environment doesn't exist in same namespace as IntegrationTestScenario at all.",
-			"integrationTestScenario:", integrationTestScenario.Name,
-			"environment:", integrationTestScenario.Spec.Environment.Name)
-		return nil, fmt.Errorf("environment %s doesn't exist in same namespace as IntegrationTestScenario at all: %w", integrationTestScenario.Spec.Environment.Name, err)
+// findEnvironmentFromAllEnvironmentsByName try to find environment from all environments by name, return the environment if it exists, otherwise return nil
+func findEnvironmentFromAllEnvironmentsByName(allEnvironments *[]applicationapiv1alpha1.Environment, envname string) *applicationapiv1alpha1.Environment {
+	for _, e := range *allEnvironments {
+		e := e //G601
+		if e.Name == envname {
+			return &e
+		}
 	}
-	return existingEnv, nil
+	return nil
 }
 
 // writeIntegrationTestStatusAtError writes updates of integration test statuses into snapshot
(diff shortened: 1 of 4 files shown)


controllers/snapshot/snapshot_adapter.go
```diff
@@ -220,6 +221,9 @@ func (a *Adapter) EnsureCreationOfEphemeralEnvironments() (controller.OperationR
 
 		_, err = a.ensureEphemeralEnvironmentForScenarioExists(&integrationTestScenario, testStatuses, components, allEnvironments)
 		if err != nil {
+			if strings.Contains(err.Error(), strconv.Quote(integrationTestScenario.Spec.Environment.Name)+" not found") {
```
dirgim, line 224: Can `errors.IsNotFound(err)` be used here instead?

hongweiliu17: No, because we also add some other info in the returned error [here](https://github.com/redhat-appstudio/integration-service/blob/main/controllers/snapshot/snapshot_adapter.go#L784)

hongweiliu17: Or we update [here](https://github.com/redhat-appstudio/integration-service/blob/main/controllers/snapshot/snapshot_adapter.go#L784) to return `err` without additional info directly?

hongweiliu17: But I hope to also add failure reason [here](https://github.com/redhat-appstudio/integration-service/pull/381/files#diff-8e718c6bf06f754db7e9e94c6f74b5c6704941dff95d637289e3c4cc194c95a0R441) then report to GH ([here](https://github.com/redhat-appstudio/integration-service/blob/main/controllers/snapshot/snapshot_adapter.go#L472-L474)) then user can know it.

dirgim: This got me thinking, we already get the list of existing environments with `a.loader.GetAllEnvironments()` [here](https://github.com/redhat-appstudio/integration-service/pull/381/files#diff-8e718c6bf06f754db7e9e94c6f74b5c6704941dff95d637289e3c4cc194c95a0R205), we could use that to immediately determine if the environment exists for the scenario and skip calling ` a.ensureEphemeralEnvironmentForScenarioExists()` entirely.

hongweiliu17: Yes, that is better than the current logic. Let me update it. Thanks!

controllers/snapshot/snapshot_adapter.go
```diff
@@ -220,6 +226,9 @@ func (a *Adapter) EnsureCreationOfEphemeralEnvironments() (controller.OperationR
 
 		_, err = a.ensureEphemeralEnvironmentForScenarioExists(&integrationTestScenario, testStatuses, components, allEnvironments)
 		if err != nil {
+			if errors.IsNotFound(err) {
+				return controller.RequeueOnErrorOrStop(fmt.Errorf("stop reconciliation, environment %s doesn't exist in the same namespace as integrationScenario %s, try again after creating environment: %w", integrationTestScenario.Spec.Environment.Name, integrationTestScenario.Name, err))
```
MartinBasti, line 230: How `RequeueOnErrorOrStop` works? I assumed it will retry when error is defined, and we have error always defined here.

controllers/snapshot/snapshot_adapter.go
```diff
@@ -477,7 +477,19 @@ func (a *Adapter) ensureEphemeralEnvironmentForScenarioExists(integrationTestSce
 	environment, ok := a.getExistingEnvironmentForScenario(integrationTestScenario, allEnvironments)
 	if !ok {
 		// environment doesn't exist, we have to create a new one
-		environment, err = a.createEnvironmentForScenario(integrationTestScenario)
+		existingEnv := findEnvironmentFromAllEnvironmentsByName(allEnvironments, integrationTestScenario.Spec.Environment.Name)
+		if existingEnv == nil {
+			a.logger.Error(fmt.Errorf("environment doesn't exist in the same namespace as integrationScenario at all, try again after creating environment"),
```
MartinBasti, line 482: https://pkg.go.dev/github.com/go-logr/logr@v1.2.4#Logger.Error

it doesn't match signature of function, `msg` is missing. I think we can pass `nil` as `err` here

controllers/snapshot/snapshot_adapter.go
```diff
@@ -42,6 +43,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
+	appStudioGroupName  = "appstudio.redhat.com"
+	environmentKindName = "environment"
+)
+
```
MartinBasti, line 50: Wouldn't be better to create our custom error? We don't need this data for, and I doesn't like hardocoding those values.

Also `NotFound` is related to go client and we are not even using it, would be better to have error with better semantics `EnvironmentNotInNamespaceError`

controllers/snapshot/snapshot_adapter.go
hongweiliu17: Yes, you are right, I should use `StopProcessing()` here.

controllers/snapshot/snapshot_adapter.go
hongweiliu17: Yes, need a msg. Thanks!

hongweiliu17: Fixed.

controllers/snapshot/snapshot_adapter.go
hongweiliu17: Fixed. Thanks.

controllers/snapshot/snapshot_adapter.go

controllers/snapshot/snapshot_adapter.go
```diff
@@ -427,16 +433,7 @@ func (a *Adapter) getExistingEnvironmentForScenario(integrationTestScenario *v1b
...
-	//get the existing environment according to environment name from integrationTestScenario
-	existingEnv, err := a.getEnvironmentFromIntegrationTestScenario(integrationTestScenario)
-	if err != nil {
-		a.logger.Error(err, "Failed to find the env defined in integrationTestScenario",
-			"integrationTestScenario.Namespace", integrationTestScenario.Namespace,
-			"integrationTestScenario.Name", integrationTestScenario.Name)
-		return nil, err
-	}
-
+func (a *Adapter) createEnvironmentForScenario(existingEnv *applicationapiv1alpha1.Environment, integrationTestScenario *v1beta1.IntegrationTestScenario) (*applicationapiv1alpha1.Environment, error) {
```
dirgim, line 436: Given the change in complexity, I think that we can skip having this function altogether and directly use `a.createCopyOfExistingEnvironment()` [here](https://github.com/redhat-appstudio/integration-service/pull/381/files#diff-8e718c6bf06f754db7e9e94c6f74b5c6704941dff95d637289e3c4cc194c95a0R486) instead of it.

dirgim (approved):
LGTM!

controllers/snapshot/snapshot_adapter.go
```diff
@@ -220,6 +220,12 @@ func (a *Adapter) EnsureCreationOfEphemeralEnvironments() (controller.OperationR
 
 		_, err = a.ensureEphemeralEnvironmentForScenarioExists(&integrationTestScenario, testStatuses, components, allEnvironments)
 		if err != nil {
+			if err.Error() == gitops.EnvironmentNotInNamespaceError {
```
MartinBasti, line 223: I'd propose to create a new module `errors`, with custom IntegrationService errors, instead of hardcoding string. The same how apimachineri api works with the errors https://github.com/kubernetes/apimachinery/blob/ab8b4c89006575f912931de69fbd924e385f2923/pkg/api/errors/errors.go#L35

So we can nicely extend it as needed, without suspicious hardcoding of error messages.

My POC: https://goplay.tools/snippet/TQ8sCZCCEw-

hongweiliu17: Thanks for your POC. It is a good proposal to have our own errors pkg, we may do it when we need it.
But for this `err`, I don't think it is hardcoding the error message, it is just to differentiate this error from others then we can use it in `if` condition to decide the next step.
Or we may just define `EnvironmentNotInNamespaceError = "EnvironmentNotInNamespaceError"`/`ResourceNotInNamespaceError = "ResourceNotInNamespaceError"`?

MartinBasti: Which design pattern or go best practice doc states that comparing textual output of error to decide code flow is good idea?
We cannot update error dynamically to provide name of expected namespace and env, we can only use constant string, so yes, I'd still call it hard-coding.
We cannot use error chaining with this solution neither.

>Or we may just define EnvironmentNotInNamespaceError = "EnvironmentNotInNamespaceError"/ResourceNotInNamespaceError = "ResourceNotInNamespaceError"?

I don't know what do you mean by this.


If you concern is that it will need additional work in this PR, feel free to merge and I will update error hadnling in a separate.

hongweiliu17: I am not concerning if it needs addition work in one PR, just thinking if it is really needed here. Even, this error is not `Environment not found`, it is an error about env creation failure due to nonexisting env, but I can't think out a better expression to return to the caller's `if`. From my point of view, creating a new module `errors` to dynamically provide name of expected namespace and env can't resolve the issue.
I would keep it in mind and always appreciate good ideas. Thanks.

MartinBasti: Why this is bad?
```
if IsEnvironmentNotInNamespaceError(err) {
```
and this is good?
```
if err.Error() == gitops.EnvironmentNotInNamespaceError {
```

MartinBasti: I your original revision you used k8s error, I suggested to use custom error.
Then instead of using the same pattern as for k8s, which was fine for you in first revisions of this PR, you created ad-hoc method for handling custom errors by comparing error messages, instead of using custom error. This custom error doesn't even need to be in separate module, it can be defined only in this controller and simplified as it is used only here.

So I'm trying to understand, why this solution is not good anymore.

dirgim: I have to say that while the current approach in this PR is a step in the right direction, I agree that the approach of having a custom error would be more maintainable in the future.
That way we would not have to depend on the contents of the error message - the message could potentially change in the future and would be hard to keep track of this way.

hongweiliu17: Yes, 'more maintainable' is a good reason. I would make change.

But, I am just curious the code below is not using 'hard-coding' to list the possible error?
```
const (
	ReasonEnvironmentNotInNamespace = "EnvironmentNotInNamespace"
	ReasonUnknownError              = "UnknownError"
)
```

hongweiliu17: > Why this is bad?
> 
> ```
> if IsEnvironmentNotInNamespaceError(err) {
> ```
> 
> and this is good?
> 
> ```
> if err.Error() == gitops.EnvironmentNotInNamespaceError {
> ```

I didn't feel it is good. Before I wrote down that error, I wanted to add a note to ask for better idea but I didn't, because I always believe you all review code carefully and leave better ideas. Thanks.

hongweiliu17: I feel we may add the `errors` under `pkg`, then it can contain more errors and be used by e2e-test maybe.

dirgim: > But, I am just curious the code below is not using 'hard-coding' to list the possible error?

You are right, it is also using hard-coding, but for the `reason` field, as opposed to using the `message` field to identify what happened. 
The `reason` should stay consistent and is easier to keep track of, while error messages can change. The wording for messages could get updated or messages can be chained by joining multiple error messages to reflect a deeper underlying cause for what caused the error.

MartinBasti: We shouldn't put errors under pkg dir, it's used only internally and we should avoid more dependencies in e2e tests.

Each package in pkg should have separate set of errors, so they really works like isolated packages for 3rd party use, this is not the case

hongweiliu17: Updated the error, thanks for the energetic discussion ^-^.

controllers/snapshot/snapshot_adapter.go
```diff
@@ -42,6 +41,39 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
+const (
+	ReasonEnvironmentNotInNamespace = "EnvironmentNotInNamespace"
+	ReasonUnknownError              = "UnknownError"
+)
+
+type IntegrationError struct {
```
dirgim, line 49: I know we decided against putting the error definition in `pkg`, but we could instead place it in the `helpers` package so it can be available to other controllers, wdyt?

hongweiliu17: Errors is moved to errorhandlers.go of helpers pkg. Thanks.

hongweiliu17: Hold on, need some update.

hongweiliu17: Ready to review.

controllers/snapshot/snapshot_adapter.go
```diff
@@ -477,8 +459,20 @@ func (a *Adapter) ensureEphemeralEnvironmentForScenarioExists(integrationTestSce
 	environment, ok := a.getExistingEnvironmentForScenario(integrationTestScenario, allEnvironments)
 	if !ok {
 		// environment doesn't exist, we have to create a new one
-		environment, err = a.createEnvironmentForScenario(integrationTestScenario)
+		existingEnv := findEnvironmentFromAllEnvironmentsByName(allEnvironments, integrationTestScenario.Spec.Environment.Name)
+		if existingEnv == nil {
+			err := h.NewEnvironmentNotInNamespaceError(integrationTestScenario.Spec.Environment.Name, integrationTestScenario.Namespace)
+			testStatuses.UpdateTestStatusIfChanged(
+				integrationTestScenario.Name, intgteststat.IntegrationTestStatusEnvironmentProvisionError,
+				fmt.Sprintf("Creation of copied ephemeral environment failed: %s. Try again after creating environment.", err))
```
dirgim, line 467: I think that we probably don't need this print here. If we do, it should be done using `a.logger.Info`.

hongweiliu17: This is to add info to status annotation so user can see it.

dirgim: You are right, I misread this line, thanks.

dirgim (approved):
LGTM

MartinBasti (approved):

Josh-Everett (approved):

PR #386, opened 2023-10-26
WIP: fix(STONEINTG-666): upgrade controller-runtime to v0.16.3

dirgim:
* Upgrade controller-runtime to v0.16.3 along with related library upgrades
* Adjust the metrics manager options to align with the updated API
* Switch to using go version 1.20 as it's required by the above upgrade
* Temporarily use the docker.io/library/golang:1.20 image since the ubi9/go-tools hasn't yet released the new version


Dockerfile
```diff
@@ -1,5 +1,7 @@
 # Build the manager binary
-FROM registry.access.redhat.com/ubi9/go-toolset:1.19.13-4.1697647145 as builder
+FROM docker.io/library/golang:1.20@sha256:098d628490c97d4419ed44a23d893f37b764f3bea06e0827183e8af4120e19be as builder
+
+WORKDIR /workspace
```
MartinBasti, line 4: 2 Workdirs defined in this stage

dirgim: My bad, c/p error, fixed.

Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM registry.access.redhat.com/ubi9/go-toolset:1.19.13-4.1697647145 as builder
+FROM docker.io/library/golang:1.20@sha256:098d628490c97d4419ed44a23d893f37b764f3bea06e0827183e8af4120e19be as builder
```
maysunfaisal, line 2: There is a chance mentioning this image here may get rate limited https://www.docker.com/increase-rate-limits/

dirgim: Yes, we are probably not going to be using that image, it's there to help test this change until we find a way to use an image originating from Red Hat-approved repositories.

PR #387, opened 2023-10-26
fix(STONEINTG-640): limit ITS name length

MartinBasti:
We rely on labels which are limited by 63 character, thus we have to limit IntegrationTestScenario name to the same limits as labels have.


config/webhook/manifests.yaml
```diff
@@ -1,32 +1,3 @@
-
----
-apiVersion: admissionregistration.k8s.io/v1
```
MartinBasti, line 3: This is automatically generated, and kubebuilder somehow decided it's not needed to have Mutating webhook.

I tested and conversion still works.

If you think that it should stay there let me know, but FYI I tried very hard to keep mutating webhook here and failed on endpoint errors :D

config/webhook/manifests.yaml
```diff
@@ -52,6 +22,7 @@ webhooks:
     operations:
     - CREATE
     - UPDATE
+    - DELETE
     resources:
-    - applications
```
MartinBasti, line 56: Dunno why it was application before, I also don't know why validation webhook has been configured without any validation code.

api/v1beta1/integrationtestscenario_suite_test.go
```diff
@@ -56,35 +63,67 @@ var _ = BeforeSuite(func() {
...
+	var err error
+	// cfg is defined in this file globally.
+	cfg, err = testEnv.Start()
 	Expect(err).NotTo(HaveOccurred())
 	Expect(cfg).NotTo(BeNil())
 
 	scheme := runtime.NewScheme()
-	Expect(AddToScheme(scheme)).To(Succeed())
+	err = AddToScheme(scheme)
+	Expect(err).NotTo(HaveOccurred())
```
dheerajodha, line 81: This change does the same thing but in 2 lines instead. Should we revert it?

[docs links](https://onsi.github.io/ginkgo/#suite-setup-and-cleanup-beforesuite-and-aftersuite)

-------------------
image

MartinBasti: I copied autogenerated code, that's why.

MartinBasti: Reverted to original

config/webhook/manifests.yaml
dirgim: I think we should be fine with removing the mutating webhook, but @jsztuka please double check this.

dirgim (approved):
LGTM

api/v1beta1/integrationtestscenario_webhook.go
```diff
@@ -17,11 +17,42 @@ limitations under the License.
...
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
 )
 
 func (r *IntegrationTestScenario) SetupWebhookWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewWebhookManagedBy(mgr).
 		For(r).
 		Complete()
 }
+
+//+kubebuilder:webhook:path=/validate-appstudio-redhat-com-v1beta1-integrationtestscenario,mutating=false,failurePolicy=fail,sideEffects=None,groups=appstudio.redhat.com,resources=integrationtestscenarios,verbs=create;update;delete,versions=v1beta1,name=vintegrationtestscenario.kb.io,admissionReviewVersions=v1
```
jsztuka, line 33: `name=vintegrationtestscenario.kb.io`
Was this generated by kubebuilder?

dheerajodha (approved):

config/webhook/manifests.yaml
jsztuka: Mutating webhook should be capable of working properly even without manifests, wo yes, this should not interfere with its current behavior.

jsztuka (approved):
lgtm

PR #400, opened 2023-11-01
feat(EC-202): Proceed if Chains could not sign

lcarva:
This commit changes the controller to no longer wait for the PipelineRun to have the annotation `chains.tekton.dev/signed` be set to the value of `"true"`. Instead, it now only waits for the annotation to be set regardless of its value. This indicates the Chains controller is done processing the PipelineRun and the integration service should move on.

This prevents Chains failures from causing integration tests to never run. Instead, let the integration tests run, and verification tools, e.g. Enterprise Contract, report on unsigned artifacts.

### DIFF
--- a/tekton/predicates_test.go
+++ b/tekton/predicates_test.go
@@ -114,7 +114,7 @@ var _ = Describe("Predicates", func() {
 			Expect(instance.Update(contextEvent)).To(BeFalse())
 		})
 
-		It("should return false when an updated event is received for a succeeded PipelineRun and not signed", func() {
+		It("should return true when an updated event is received for a succeeded PipelineRun with failed chains signing", func() {
 			newPipelineRun.Status.SetCondition(&apis.Condition{
 				Type:   apis.ConditionSucceeded,
 				Status: "True",
@@ -126,8 +126,8 @@ var _ = Describe("Predicates", func() {
 			}
 			Expect(instance.Update(contextEvent)).To(BeFalse())
 
-			newPipelineRun.Annotations["chains.tekton.dev/signed"] = "false"
-			Expect(instance.Update(contextEvent)).To(BeFalse())
+			newPipelineRun.Annotations["chains.tekton.dev/signed"] = "failed"
+			Expect(instance.Update(contextEvent)).To(BeTrue())
 		})
 
 		It("should return false when an updated event is received for a succeeded and signed pipelineRun that was annotated with Snapshot", func() {
--- a/tekton/utils.go
+++ b/tekton/utils.go
@@ -107,11 +107,19 @@ func hasPipelineRunStateChangedToDeleting(objectOld, objectNew client.Object) bo
 	return false
 }
 
-// isPipelineRunSigned returns a boolean indicated whether the PipelineRun been signed
-// If the object passed to this function is not a PipelineRun, the function will return false.
-func isPipelineRunSigned(objectNew client.Object) bool {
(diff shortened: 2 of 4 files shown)


lcarva:
When reviewing this PR, please help determine if you agree that the modified behavior is indeed better than the previous one.

dirgim (approved):
LGTM

tekton/predicates_test.go
```diff
@@ -114,7 +114,7 @@ var _ = Describe("Predicates", func() {
 			Expect(instance.Update(contextEvent)).To(BeFalse())
 		})
 
-		It("should return false when an updated event is received for a succeeded PipelineRun and not signed", func() {
+		It("should return true when an updated event is received for a succeeded PipelineRun and not signed", func() {
```
MartinBasti, line 117: can we rephrase it to something like `when an updated event is received for a succeeded PipelineRun with failed chains signing` ?

MartinBasti:
Please update also description in diagram here:
https://github.com/redhat-appstudio/integration-service/blob/d7620c9eb233d1c406e721513ad9d1089cd6009b/docs/build_pipeline_controller.md?plain=1#L13

tekton/utils.go
```diff
@@ -107,11 +107,19 @@ func hasPipelineRunStateChangedToDeleting(objectOld, objectNew client.Object) bo
...
+func isChainsDoneWithPipelineRun(objectNew client.Object) bool {
 	if newPipelineRun, ok := objectNew.(*tektonv1beta1.PipelineRun); ok {
-		return metadata.HasAnnotationWithValue(newPipelineRun, PipelineRunChainsSignedAnnotation, "true")
+		// If the annotation value is set "true" it means Chains was able to sign. If it is set to
+		// "failed", it means something prevented Chains from signing, e.g. insufficient access. In
+		// either case, we want to proceed processing the PipelineRun instead of waiting
+		// indefinitely for a condition that may never happen, e.g. "failed" -> "true". Let
+		// downstream verification processes, e.g. Enterprise Contract, deal with the Chains
+		// failure.
+		return metadata.HasAnnotation(newPipelineRun, PipelineRunChainsSignedAnnotation)
```
dheerajodha, line 122: just to confirm: if we go by this approach, then we would only be checking if this annotation is present within the PLR even though it: 1) has no value, 2) is set to "true", or 3) is set to "failed".

Is it possible that this annotation could at some point be empty (or has no value) and cause this bug ([RHTAPBUGS-214](https://issues.redhat.com/browse/RHTAPBUGS-214)) again? Another part of me says it won't hit that bug because `Missing pipelinerun attestation` means _absence_ of this annotation.

lcarva: Looking at the Chains code base does, the Chains controller does not set that annotation to an empty value. So this would only happen if something else does that. (But if there's something else playing with that annotation we're already in trouble.)

You may see `Missing pipelinerun attestation` again if Chains does fail to process the PipelineRun, but not because it hasn't had a chance to do so. This is the correct behavior IMO.

tekton/predicates_test.go
lcarva: Done.

lcarva:
> Please update also description in diagram

Done!

MartinBasti (approved):

tekton/utils.go
dheerajodha: That makes sense, thanks for the explanation!

dheerajodha (approved):
LGTM

PR #408, opened 2023-11-07
feat(STONEINTG-540): manual re-run integration tests

MartinBasti:
This PR allows to trigger manual rerun of an integration test via snapshot annotation.


gitops/snapshot.go
```diff
@@ -658,3 +683,46 @@ func GetComponentSourceFromComponent(component *applicationapiv1alpha1.Component
...
 }
+
+// GetIntegrationTestRunLabelValue returns value of the label responsible for re-running tests
+func GetIntegrationTestRunLabelValue(snapshot applicationapiv1alpha1.Snapshot) (string, bool) {
+	labels := snapshot.GetLabels()
+	labelVal, ok := labels[SnaphostIntegrationTestRun]
+	return labelVal, ok
+}
+
+// TODO: take this function from operator-toolkit
```
MartinBasti, line 694: :eyes:

MartinBasti:
Back to draft: re-running integration tests of finished snapshot in ephemeral environments are causing deadlock, where no test is actually executed after deploying env and SEB is not removed.

controllers/snapshot/snapshot_adapter.go
```diff
@@ -75,6 +76,87 @@ func scenariosNamesToList(integrationTestScenarions *[]v1beta1.IntegrationTestSc
...
+					return controller.RequeueWithError(err)
+				}
+				return controller.StopProcessing()
+			}
+			return controller.RequeueWithError(fmt.Errorf("failed to ensure existence of the environment for scenario %s: %w", integrationTestScenario.Name, err))
+		}
+
+	} else {
+		integrationTestScenarioStatus, ok := testStatuses.GetScenarioStatus(integrationTestScenario.Name)
+		if ok && integrationTestScenarioStatus.TestPipelineRunName != "" {
```
dirgim, line 133: Do we need this check here since the status for that scenario will be reset `testStatuses.ResetStatus(scenarioName)` at this point?

MartinBasti: yes, in case that for unknown reason, label failed to be removed and reconciliation may re-trigger another test.

dirgim: I see your point, that's certainly possible, but wouldn't the `TestPipelineRunName` always be removed [here](https://github.com/redhat-appstudio/integration-service/pull/408/files#diff-eeed1ae6f738d926a3e659cd3b10bbeb1f2c2a05f7ab22d899fa6914059660c0R122) before we check this?

MartinBasti: yes /o\

dirgim: That's a tough case actually. What if we don't allow rerunning Integration tests that are still running?

controllers/binding/snapshotenvironmentbinding_adapter.go
```diff
@@ -75,11 +75,6 @@ func NewAdapter(snapshotEnvironmentBinding *applicationapiv1alpha1.SnapshotEnvir
 // EnsureIntegrationTestPipelineForScenarioExists is an operation that will ensure that the Integration test pipeline
 // associated with the Snapshot and the SnapshotEnvironmentBinding's IntegrationTestScenarios exist.
 func (a *Adapter) EnsureIntegrationTestPipelineForScenarioExists() (controller.OperationResult, error) {
-	if gitops.HaveAppStudioTestsFinished(a.snapshot) {
```
MartinBasti, line 78: This is my fix for the deadlock, I couldn't figure out what's the purpose of this gating.

dheerajodha: maybe because why would we want to create Integration PLRs for a snapshot that has already finished its testing?

MartinBasti: > maybe because why would we want to create Integration PLRs for a snapshot that has already finished its testing?

I'm re-running old tests in GH in order to figure out if it's regression, or race condition, or bug caused by something else, if new PRs are failing:D

controllers/snapshot/snapshot_adapter.go
MartinBasti: Updated and unittest added

controllers/binding/snapshotenvironmentbinding_adapter_test.go
```diff
@@ -410,33 +410,6 @@ var _ = Describe("Binding Adapter", Ordered, func() {
 
 	})
 
-	It("ensures the integrationTestPipelines are NOT created for a Snapshot that finished testing", func() {
```
MartinBasti, line 413: Surprisingly this test didn't fail yesterday

controllers/snapshot/snapshot_adapter.go
```diff
@@ -75,6 +76,91 @@ func scenariosNamesToList(integrationTestScenarions *[]v1beta1.IntegrationTestSc
...
+
+	a.logger.Info("Re-running integration test for scenario", "scenario", scenarioName)
+
+	testStatuses, err := gitops.NewSnapshotIntegrationTestStatusesFromSnapshot(a.snapshot)
+	if err != nil {
+		return controller.RequeueWithError(err)
+	}
+
+	integrationTestScenarioStatus, ok := testStatuses.GetScenarioStatus(integrationTestScenario.Name)
+	if ok && integrationTestScenarioStatus.Status == intgteststat.IntegrationTestStatusInProgress {
```
dirgim, line 109: Thinking about this, maybe we also want to include the pending state here? 
If our re-reconciliation upon failing to remove the label starts before the PLR starts, we might find the status still in pending?

dirgim: To clarify, I'm thinking about the ephemeral environment scenario here, the static example will be in the `InProgress` state.

gitops/snapshot.go
MartinBasti: Replaced with operator-toolkit

controllers/snapshot/snapshot_adapter.go
MartinBasti: That makes sense, added

controllers/snapshot/snapshot_adapter.go
```diff
@@ -75,6 +76,92 @@ func scenariosNamesToList(integrationTestScenarions *[]v1beta1.IntegrationTestSc
 	return &result
 }
 
+// EnsureRerunPipelineRunsExist is responsible for recreating integration test pipelines triggered by users
+func (a *Adapter) EnsureRerunPipelineRunsExist() (controller.OperationResult, error) {
+
+	scenarioName, ok := gitops.GetIntegrationTestRunLabelValue(*a.snapshot)
+	if !ok {
+		// no test rerun trigerred
```
hongweiliu17, line 84:
```suggestion
		// no test rerun triggered
```

gitops/snapshot.go
```diff
@@ -585,6 +588,26 @@ func HasSnapshotTestAnnotationChanged(objectOld, objectNew client.Object) bool {
 	return false
 }
 
+// HasSnapshotRerunLabelChanged returns a boolean indicating whether the Snapshot annotation for re-running
```
hongweiliu17, line 591:
```suggestion
// HasSnapshotRerunLabelChanged returns a boolean indicating whether the Snapshot label for re-running
```

gitops/snapshot.go
```diff
@@ -42,6 +42,9 @@ const (
 	// SnapshotTypeLabel contains the type of the Snapshot.
 	SnapshotTypeLabel = "test.appstudio.openshift.io/type"
 
+	// SnaphostIntegrationTestRun contains name of test we want to trigger run
+	SnaphostIntegrationTestRun = "test.appstudio.openshift.io/run"
```
hongweiliu17, line 46:
```suggestion
	SnapshotIntegrationTestRun = "test.appstudio.openshift.io/run"
```

loader/loader.go
```diff
@@ -443,3 +444,18 @@ func (l *loader) GetAutoReleasePlansForApplication(c client.Client, ctx context.
 
 	return &releasePlans.Items, nil
 }
+
+// GetScenario returns integration test scenario requested by name and namespace
+func (l *loader) GetScenario(c client.Client, ctx context.Context, name, namespace string) (*v1beta1.IntegrationTestScenario, error) {
+	scenario := &v1beta1.IntegrationTestScenario{}
```
hongweiliu17, line 450: We may replace these code in this function with the code like [this](https://github.com/redhat-appstudio/integration-service/blob/main/loader/loader.go#L118-L119) 
```
	scenario := &v1beta1.IntegrationTestScenario{}
	return scenario, toolkit.GetObject(name, namespace, c, ctx, scenario)
```

controllers/snapshot/snapshot_adapter.go
```diff
@@ -75,6 +76,92 @@ func scenariosNamesToList(integrationTestScenarions *[]v1beta1.IntegrationTestSc
...
+
+	a.logger.Info("Re-running integration test for scenario", "scenario", scenarioName)
+
+	testStatuses, err := gitops.NewSnapshotIntegrationTestStatusesFromSnapshot(a.snapshot)
+	if err != nil {
+		return controller.RequeueWithError(err)
+	}
+
+	integrationTestScenarioStatus, ok := testStatuses.GetScenarioStatus(integrationTestScenario.Name)
+	if ok && (integrationTestScenarioStatus.Status == intgteststat.IntegrationTestStatusInProgress ||
```
hongweiliu17, line 109: This might be a corner case: Is there a user case that a user wants to rerun an inProgress PLR or trigger PLR for a pending integration test scenario that is not triggered yet due to the error in other integration test scenarios?

MartinBasti: We have a separate story to run integration tests independently, so one test wouldn't be blocked by error in the other.
I'd not solve this here but rather wait for the other story to be fixed.

hongweiliu17: I am fine with it as long as it is known.

MartinBasti:
@hongweiliu17 PR updated with fixes, PTAL

hongweiliu17 (approved):
LGTM

PR #410, opened 2023-11-08
(STONEINTG-687): add finalizer to component controller

Josh-Everett:
Changes to predicate logic for component controller to react to update events as opposed to deletion events. Alongside the addition of a finalizer ensures that the component controller will be able to accomplish generation of new snapshot for remaining components before the resource is removed from the cluster. 

### DIFF
--- a/controllers/component/component_adapter.go
+++ b/controllers/component/component_adapter.go
@@ -19,6 +19,8 @@ package component
 import (
 	"context"
 
+	"fmt"
+
 	applicationapiv1alpha1 "github.com/redhat-appstudio/application-api/api/v1alpha1"
 	"github.com/redhat-appstudio/integration-service/gitops"
 	h "github.com/redhat-appstudio/integration-service/helpers"
@@ -27,6 +29,7 @@ import (
 	"github.com/redhat-appstudio/operator-toolkit/controller"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
 // Adapter holds the objects needed to reconcile a integration PipelineRun.
@@ -52,10 +55,25 @@ func NewAdapter(component *applicationapiv1alpha1.Component, application *applic
 	}
 }
 
+// EnsureComponentHasFinalizer is an operation that will ensure components
+// that have been created are assigned a finalizer
+func (a *Adapter) EnsureComponentHasFinalizer() (controller.OperationResult, error) {
+	if !isComponentMarkedForDeletion(a.component) {
+		if !controllerutil.ContainsFinalizer(a.component, h.ComponentFinalizer) {
+			err := h.AddFinalizerToComponent(a.client, a.logger, a.context, a.component, h.ComponentFinalizer)
+			if err != nil {
+				return controller.RequeueWithError(fmt.Errorf("failed to add the finalizer: %w", err))
+			}
+		}
+	}
+
+	return controller.ContinueProcessing()
+}
+
 // EnsureComponentIsCleanedUp is an operation that will ensure components
 // marked for deletion have a snapshot created without said component
 func (a *Adapter) EnsureComponentIsCleanedUp() (controller.OperationResult, error) {
-	if !hasComponentBeenDeleted(a.component) {
+	if !isComponentMarkedForDeletion(a.component) {
 		return controller.ContinueProcessing()
 	}
 
@@ -80,15 +98,16 @@ func (a *Adapter) EnsureComponentIsCleanedUp() (controller.OperationResult, erro
 		}
 	}
 
-	if len(snapshotComponents) == 0 {
-		a.logger.Info("Application has no available snapshot components for snapshot creation")
-		return controller.StopProcessing()
+	if len(snapshotComponents) != 0 {
+		_, err = a.createUpdatedSnapshot(&snapshotComponents)
+		if err != nil {
+			a.logger.Error(err, "Failed to create new snapshot after component deletion")
+			return controller.RequeueWithError(err)
+		}
 	}
-
-	_, err = a.createUpdatedSnapshot(&snapshotComponents)
+	err = h.RemoveFinalizerFromComponent(a.client, a.logger, a.context, a.component, h.ComponentFinalizer)
 	if err != nil {
-		a.logger.Error(err, "Failed to create new snapshot after component deletion")
-		return controller.RequeueWithError(err)
+		return controller.RequeueWithError(fmt.Errorf("failed to remove the finalizer: %w", err))
 	}
 
 	return controller.ContinueProcessing()
@@ -109,7 +128,7 @@ func (a *Adapter) createUpdatedSnapshot(snapshotComponents *[]applicationapiv1al
 
 	err := ctrl.SetControllerReference(a.application, snapshot, a.client.Scheme())
 	if err != nil {
-		a.logger.Error(err, "Failed to set controller refrence")
+		a.logger.Error(err, "Failed to set controller reference")
 		return nil, err
 	}
 
@@ -123,10 +142,22 @@ func (a *Adapter) createUpdatedSnapshot(snapshotComponents *[]applicationapiv1al
 	return snapshot, nil
 }
 
-func hasComponentBeenDeleted(object client.Object) bool {
-
+func isComponentMarkedForDeletion(object client.Object) bool {
 	if comp, ok := object.(*applicationapiv1alpha1.Component); ok {
 		return !comp.ObjectMeta.DeletionTimestamp.IsZero()
 	}
+	return false
+}
+
+// hasComponentChangedToDeleting returns a boolean indicating whether the Component was marked for deletion.
+// If the objects passed to this function is not a Component, the function will return false.
+func hasComponentChangedToDeleting(objectOld, objectNew client.Object) bool {
+	if oldComponent, ok := objectOld.(*applicationapiv1alpha1.Component); ok {
+		if newComponent, ok := objectNew.(*applicationapiv1alpha1.Component); ok {
+			return (oldComponent.GetDeletionTimestamp() == nil &&
+				newComponent.GetDeletionTimestamp() != nil)
+		}
+	}
+
 	return false
 }
--- a/controllers/component/component_controller.go
+++ b/controllers/component/component_controller.go
@@ -43,7 +43,7 @@ type Reconciler struct {
 func NewComponentReconciler(client client.Client, logger *logr.Logger, scheme *runtime.Scheme) *Reconciler {
 	return &Reconciler{
 		Client: client,
-		Log:    logger.WithName("integration pipeline"),
+		Log:    logger.WithName("component"),
 		Scheme: scheme,
 	}
 }
@@ -71,6 +71,11 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	var application *applicationapiv1alpha1.Application
 	application, err = loader.GetApplicationFromComponent(r.Client, ctx, component)
 	if err != nil {
+		if errors.IsNotFound(err) {
+			if err := helpers.RemoveFinalizerFromComponent(r.Client, logger, ctx, component, helpers.ComponentFinalizer); err != nil {
+				return ctrl.Result{}, err
+			}
+		}
 		logger.Error(err, "Failed to get Application from the component",
 			"Component.Name", component.Name)
 		return ctrl.Result{}, err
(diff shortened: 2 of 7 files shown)


controllers/component/component_adapter.go
```diff
@@ -55,6 +58,11 @@ func NewAdapter(component *applicationapiv1alpha1.Component, application *applic
 // EnsureComponentIsCleanedUp is an operation that will ensure components
 // marked for deletion have a snapshot created without said component
 func (a *Adapter) EnsureComponentIsCleanedUp() (controller.OperationResult, error) {
+	if !controllerutil.ContainsFinalizer(a.component, h.ComponentFinalizer) {
```
dirgim, line 61: I'm thinking that this might be too late to add a finalizer to a resource, since it's already being deleted.

An alternative solution might be to not try and `Get` the component from the cluster in the component_controller.go and instead use the `req.NamespacedName.Name` to extract the `Name` of the component and use that for the further logic in this adapter.

dirgim: Just realized that my alternative suggestion would be missing a way to find the `Application`, but my original point still stands.

Maybe we can also react to `Component` creation events and make sure that the finalizer is added then?

controllers/component/predicates.go
```diff
@@ -5,8 +5,29 @@ import (
...
 
+// ComponentCreatedPredicate returns a predicate which filters out
+// only components that have are the result of a create event
+func ComponentCreatedPredicate() predicate.Predicate {
+	return predicate.Funcs{
+		CreateFunc: func(createEvent event.CreateEvent) bool {
+			return true
+		},
+		DeleteFunc: func(deleteEvent event.DeleteEvent) bool {
+			return true
```
dirgim, line 16: While this doesn't hurt things currently, might be better to set it to `false` for clarity in the future.

controllers/component/component_adapter.go
```diff
@@ -52,6 +55,17 @@ func NewAdapter(component *applicationapiv1alpha1.Component, application *applic
 	}
 }
 
+// EnsureComponentHasFinalizer is an operation that will ensure components
+// that have been created are assigned a finalizer
+func (a *Adapter) EnsureComponentHasFinalizer() (controller.OperationResult, error) {
```
dirgim, line 60: Maybe we should check if the component is not being deleted (with something like `if hasComponentBeenDeleted(a.component)`) so we don't add a finalizer to such component unnecessarily in that case.

controllers/component/component_adapter.go
```diff
@@ -52,6 +55,17 @@ func NewAdapter(component *applicationapiv1alpha1.Component, application *applic
...
+	if !controllerutil.ContainsFinalizer(a.component, h.ComponentFinalizer) {
+		controllerutil.AddFinalizer(a.component, h.ComponentFinalizer)
```
dirgim, line 62: I don't think that this call updates(patches) the resource on the cluster, it only adds the finalizer to the object without applying it.
We should probably have a function like [AddFinalizerToPipelineRun](https://github.com/redhat-appstudio/integration-service/blob/c199d2673bc5c7dc9936798acce1ffd9bad6864b/helpers/integration.go#L468) for this.

Untested idea: maybe we can even make that function generic so we don't have duplicate versions for all different kinds of objects.

controllers/component/component_adapter.go
Josh-Everett: good idea, added that [here](https://github.com/redhat-appstudio/integration-service/pull/410/files#diff-0a268eab99b22ffc9ca9a74ef2752ed7dd05a822b3da8fe160ffee3a2ad1035eR61)

controllers/component/component_adapter.go
Josh-Everett: You're right. Should appropriately patch now in the new helper function in most recent pr.
The PLR and Component finalizer functions share near identical code, but I'm not sure the best way to go about turning the functions into one generic with both ` pipelineRun *tektonv1.PipelineRun`  and  `component *applicationapiv1alpha1.Component` types?

controllers/component/component_adapter.go
dirgim: Hm, looking at it now, I think the check should be inverse (`! hasComponentBeenDeleted(a.component)`) because currently it only adds the finalizer to components that begin to be deleted.

Josh-Everett: Good catch, thanks - fixed in most recent prs.

dirgim: Yup, thanks!

helpers/integration.go
```diff
@@ -476,6 +476,37 @@ func AddFinalizerToPipelineRun(adapterClient client.Client, logger IntegrationLo
 
 		logger.LogAuditEvent("Added Finalizer to the Integration PipelineRun", pipelineRun, LogActionUpdate, "finalizer", finalizer)
 	}
+	return nil
+}
+
+// AddFinalizerToComponent adds the finalizer to the component.
+// If finalizer was not added successfully, a non-nil error is returned.
+func AddFinalizerToComponent(adapterClient client.Client, logger IntegrationLogger, ctx context.Context, component *applicationapiv1alpha1.Component, finalizer string) error {
```
dirgim, line 484: We'll need [unit test coverage](https://app.codecov.io/gh/redhat-appstudio/integration-service/pull/410?src=pr&el=tree&utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=redhat-appstudio) for these functions - should be simple enough to add them to the [existing similar ones](https://github.com/redhat-appstudio/integration-service/blob/8038e5a25bd7ce574bf1ba9ab4298ab9c801d0b2/helpers/integration_test.go#L1042-L1057).

Josh-Everett: ack pr now includes [unit test](https://github.com/redhat-appstudio/integration-service/blob/9d75fa1ad8e2fe28fea5e9c327950e5a9f7a953b/helpers/integration_test.go#L1059), thanks for pointing me to the good reference.

dirgim: Thanks, looks good!

kasemAlem:
After investigating the artifacts and logs of [ci/prow/integration-service-e2e](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/redhat-appstudio_integration-service/410/pull-ci-redhat-appstudio-integration-service-main-integration-service-e2e/1747660006464950272) , for example , changes made in this PR lead to 3 components to be not deleted and looking into [components artifact](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/redhat-appstudio_integration-service/410/pull-ci-redhat-appstudio-integration-service-main-integration-service-e2e/1747660006464950272/artifacts/integration-service-e2e/redhat-appstudio-gather/artifacts/components.json) the finlizer was not removed as expected .

controllers/component/predicates.go
```diff
@@ -5,8 +5,29 @@ import (
...
+		},
+		DeleteFunc: func(deleteEvent event.DeleteEvent) bool {
+			return false
+		},
+		GenericFunc: func(genericEvent event.GenericEvent) bool {
+			return false
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return false
```
hongweiliu17, line 22: We may need to also `return true` here since we will check the `DeletionTimestamp`

hongweiliu17: Sorry, ignore this comment, this function is for create event only.

controllers/component/component_adapter.go
```diff
@@ -81,6 +100,10 @@ func (a *Adapter) EnsureComponentIsCleanedUp() (controller.OperationResult, erro
 	}
 
 	if len(snapshotComponents) == 0 {
+		err = h.RemoveFinalizerFromComponent(a.client, a.logger, a.context, a.component, h.ComponentFinalizer)
```
hongweiliu17, line 103: I think we should remove the finalizer from the component whether `len(snapshotComponents) == 0` or not here,  to make the component be able to be deleted, so we can combine this line and L117 by move this line out of thif `if`, how do you think of it?

controllers/component/predicates.go
```diff
@@ -19,7 +40,7 @@ func ComponentDeletedPredicate() predicate.Predicate {
 			return false
 		},
 		UpdateFunc: func(e event.UpdateEvent) bool {
```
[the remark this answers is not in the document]
Josh-Everett: yes - done [here](https://github.com/redhat-appstudio/integration-service/pull/410/files#diff-34d1d778ced71d374bfcf517373a0f1ad9f1ddba04d8333430f80cabc5d5e178R37)

controllers/component/component_adapter.go
```diff
@@ -124,9 +152,21 @@ func (a *Adapter) createUpdatedSnapshot(snapshotComponents *[]applicationapiv1al
 }
 
 func hasComponentBeenDeleted(object client.Object) bool {
```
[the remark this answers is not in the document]
Josh-Everett: see what you think of this renaming [here](https://github.com/redhat-appstudio/integration-service/pull/410/files#diff-0a268eab99b22ffc9ca9a74ef2752ed7dd05a822b3da8fe160ffee3a2ad1035eR155)

controllers/component/component_adapter.go
Josh-Everett: Yes I agree the code could be cleaned up @hongweiliu17 what are your thoughts on this refactor [here](https://github.com/redhat-appstudio/integration-service/pull/410/files#diff-0a268eab99b22ffc9ca9a74ef2752ed7dd05a822b3da8fe160ffee3a2ad1035eR102) . I think the prior code should have made a call to StopProcessing where it would Continue which I also fixed here.

hongweiliu17: Agree. Thanks.

controllers/component/component_controller.go
```diff
@@ -43,7 +43,7 @@ type Reconciler struct {
 func NewComponentReconciler(client client.Client, logger *logr.Logger, scheme *runtime.Scheme) *Reconciler {
 	return &Reconciler{
 		Client: client,
-		Log:    logger.WithName("integration pipeline"),
+		Log:    logger.WithName("component"),
 		Scheme: scheme,
 	}
 }
```
dheerajodha, line 49: The e2e tests are consistently failing with `context deadline exceeded` while deleting the components ([example](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/redhat-appstudio_integration-service/410/pull-ci-redhat-appstudio-integration-service-main-integration-service-e2e/1749627562767683584#1:build-log.txt%3A1076)), because as part of the [cleanup logic of our E2E tests](https://github.com/redhat-appstudio/e2e-tests/blob/cd8f045878b284a4a06ddb256ba24b088dff3714/tests/integration-service/integration.go#L408), we delete the application first and then it's component.

Due to this, right when the Application is deleted, and right when the [tests try to delete the component](https://github.com/redhat-appstudio/e2e-tests/blob/cd8f045878b284a4a06ddb256ba24b088dff3714/tests/integration-service/integration.go#L411), integration service picks up this component (since its marked for deletion) for reconciliation but consistently requeues the request with error `"Failed to get Application from the component"` right here:
https://github.com/redhat-appstudio/integration-service/blob/883f356190e853fb3653f2230acba875186df37f/controllers/component/component_controller.go#L74

In this case, I think we should remove the finalizer from the component if we cannot find its Application. It would be something similar to what we did [right here](https://github.com/redhat-appstudio/integration-service/blob/main/controllers/integrationpipeline/integrationpipeline_controller.go#L88-L100).

controllers/component/component_adapter.go
dheerajodha: I like the refactoring here. Based on the latest changes, does this mean that now we'll continue to reconcile a Component that has been marked for deletion and doesn't have a finalizer on it (so it will soon get deleted by K8s)? Wouldn't it cause a problem in the future if some other Ensure* function tries to process this component?

controllers/component/component_controller.go
Josh-Everett: Thanks Dheeraj, that is very helpful. I was looking at that code last night when trying to understand when/where I was missing a call to remove finalizer compared to other controllers. I couldn't understand the why so you saved me a bunch of investigating here.

controllers/component/component_adapter.go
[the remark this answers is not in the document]
dheerajodha: Looks good, thank you!

controllers/component/component_controller.go
dheerajodha: No problem at all, this was a little tricky anyway. I'm marking this thread as resolved since the e2e tests passed in the CI.

controllers/component/component_adapter.go
dheerajodha: I'm just thinking if removing `return controller.StopProcessing()` was the right decision, what do you think?

MartinBasti (approved):

controllers/component/component_adapter.go
Josh-Everett: I think this is similar to the way the PLR controller handles reconciliation around finalizer removal but maybe I am misunderstanding when it's correct to StopProcessing vs other options

dheerajodha: I won't fight for it because the possibility of getting into a problematic state is low for now.
After all, we'd need to have:
* an additional Ensure* function that gets called right after this one (`EnsureComponentIsCleanedUp`)
* a Component (without finalizer, with deletionTimestamp), which is about to be deleted in 0.000001 secs, that gets picked up for reconciliation by this new Ensure* function.

dheerajodha (approved):
Nice!
LGTM

kasemAlem (approved):

PR #412, opened 2023-11-08
feat(STONEINTG-619): use Tekton v0.49.0 and v1 APIs

dirgim:
Extracted first part of updates needed to support controller-runtime v0.16.3 ([STONEINTG-677](https://issues.redhat.com//browse/STONEINTG-677)) and latest version of release service APIs.

* Upgrade to Tekton v0.49.0 and use latest version of application-api
* Switch to using Tekton v1 APIs in Integration service
* Update Tekton TaskResult and PipelineResult references to use the updated Result API format
* Replace the remaining spec.Bundle references to use git or bundle resolvers where appropriate
* Remove loader/integration.go as it's not used in the integration-service any more and is causing unneeded churn for updating to Tekton v1 APIs
* Update unit tests with the changes


loader/integration.go
```diff
@@ -21,21 +21,21 @@
...
 // an error will be returned.
-func GetLatestPipelineRunForSnapshotAndScenario(adapterClient client.Client, ctx context.Context, loader ObjectLoader, snapshot *applicationapiv1alpha1.Snapshot, integrationTestScenario *v1beta1.IntegrationTestScenario) (*tektonv1beta1.PipelineRun, error) {
+func GetLatestPipelineRunForSnapshotAndScenario(adapterClient client.Client, ctx context.Context, loader ObjectLoader, snapshot *applicationapiv1alpha1.Snapshot, integrationTestScenario *v1beta1.IntegrationTestScenario) (*tektonv1.PipelineRun, error) {
 	integrationPipelineRuns, err := loader.GetAllPipelineRunsForSnapshotAndScenario(adapterClient, ctx, snapshot, integrationTestScenario)
 	if err != nil {
 		return nil, err
 	}
 
-	var latestIntegrationPipelineRun = &tektonv1beta1.PipelineRun{}
+	var latestIntegrationPipelineRun = &tektonv1.PipelineRun{}
```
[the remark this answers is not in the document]
dirgim: Removed the entire loader/integration.go as it's not used in the integration-service any more and is causing unneeded churn for updating to Tekton v1 APIs.

controllers/binding/snapshotenvironmentbinding_adapter.go
```diff
@@ -35,7 +35,7 @@ import (
 	h "github.com/redhat-appstudio/integration-service/helpers"
 	"github.com/redhat-appstudio/integration-service/loader"
 	"github.com/redhat-appstudio/integration-service/tekton"
-	pipeline "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
+	tektonv1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"
```
14rcole, line 38: Why do we need to update the import alias here?

dirgim: I did it to make the import standardized over our packages. In all other cases we were using tektonv1beta1. This also makes it easier to figure out which version of the library we are using when referencing objects.

14rcole (approved):
LGTM

gitops/binding_test.go
```diff
@@ -179,7 +179,6 @@ var _ = Describe("Gitops functions for managing Snapshots", Ordered, func() {
 		Expect(gitops.HaveAppStudioTestsSucceeded(hasSnapshot)).To(BeTrue())
 		bindingComponents := gitops.NewBindingComponents(hasSnapshot)
 		Expect(bindingComponents).NotTo(BeNil())
-		Expect(hasComp.Spec.Replicas).To(Equal(0))
```
MartinBasti, line 182: Why this change?

config/crd/bases/appstudio.redhat.com_integrationtestscenarios.yaml
```diff
@@ -108,8 +108,6 @@ spec:
                         required:
                         - deploymentTargetClaim
                         type: object
-                    required:
```
MartinBasti, line 111: Why this change?

MartinBasti (commented):
Do we need to use go 1.20 here?

config/crd/bases/appstudio.redhat.com_integrationtestscenarios.yaml
dirgim: This is a result of the [change on the application-api side](https://github.com/redhat-appstudio/application-api/commit/1df054f63121c3860fac87679e61e2cdd1c91a15) being applied here. The `env` field is not required anymore in the latest version of the Environment configuration, and the ITS API [references the same struct](https://github.com/redhat-appstudio/integration-service/blob/3def7addb6470aec18020efc92c02a79fe3a7cc1/api/v1beta1/integrationtestscenario_types.go#L57).

gitops/binding_test.go
dirgim: The replicas are not automatically set to 0 in the newer versions of application-api (see this [commit](https://github.com/redhat-appstudio/application-api/commit/4b8b7924dfaf9981d479c6b4ca9f68b6e27368eb)) and it is now not defined by default. 
This is correct behavior because the replicas field is only there to override argoCD configuration if users wish to.

dirgim:
> Do we need to use go 1.20 here?

It's only required by the controller-runtime v0.16.3 upgrade, so I decided to do that as part of that PR to keep this one focused on (mainly) the tekton-related changes.

dheerajodha (approved):

PR #418, opened 2023-11-13
fix(stoneintg-645): failed to get component for build pipeline

14rcole:
Update the component controller log message that is outputted when a build pipeline could not be found.  The new log message is outputted as 'INFO' rather than 'ERROR' and contains more information

### DIFF
--- a/controllers/buildpipeline/buildpipeline_controller.go
+++ b/controllers/buildpipeline/buildpipeline_controller.go
@@ -92,17 +92,20 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 				return ctrl.Result{}, err
 			}
 		}
-		return helpers.HandleLoaderError(logger, err, "Component", "PipelineRun")
+		return helpers.HandleLoaderError(logger, err, "component", "pipelineRun")
+	} else if component == nil {
+		// if both component and error are nil then the component label for the pipeline did not exist
+		// in this case we should stop reconciliation
+		logger.Info("Failed to  get component for build pipeline - component label does not exist", "name", pipelineRun.Name, "namespace", pipelineRun.Namespace)
+		return ctrl.Result{}, nil
 	}
 
 	application := &applicationapiv1alpha1.Application{}
-	if component != nil {
-		application, err = loader.GetApplicationFromComponent(r.Client, ctx, component)
-		if err != nil {
-			logger.Error(err, "Failed to get Application from Component",
-				"Component.Name ", component.Name, "Component.Namespace ", component.Namespace)
-			return ctrl.Result{}, err
-		}
+	application, err = loader.GetApplicationFromComponent(r.Client, ctx, component)
+	if err != nil {
+		logger.Error(err, "Failed to get Application from Component",
+			"Component.Name ", component.Name, "Component.Namespace ", component.Namespace)
+		return ctrl.Result{}, err
 	}
 
 	if application == nil {
--- a/controllers/component/component_controller.go
+++ b/controllers/component/component_controller.go
@@ -65,7 +65,6 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		if errors.IsNotFound(err) {
 			return ctrl.Result{}, nil
 		}
-
 		return ctrl.Result{}, err
 	}
 
--- a/controllers/buildpipeline/buildpipeline_adapter.go
+++ b/controllers/buildpipeline/buildpipeline_adapter.go
@@ -73,10 +73,6 @@ func (a *Adapter) EnsureSnapshotExists() (controller.OperationResult, error) {
 		return controller.ContinueProcessing()
 	}
 
-	if a.component == nil {
-		return a.removeFinalizerAndContinueProcessing()
-	}
-
 	isLatest, err := a.isLatestSucceededBuildPipelineRun()
 	if err != nil {
 		return controller.RequeueWithError(err)
--- a/controllers/buildpipeline/buildpipeline_adapter_test.go
+++ b/controllers/buildpipeline/buildpipeline_adapter_test.go
@@ -19,8 +19,9 @@ package buildpipeline
 import (
 	"bytes"
 	"reflect"
-	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"time"
+
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	"github.com/redhat-appstudio/integration-service/gitops"
 	"github.com/redhat-appstudio/integration-service/helpers"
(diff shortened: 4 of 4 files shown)


controllers/component/component_controller.go
```diff
@@ -61,11 +61,12 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	component := &applicationapiv1alpha1.Component{}
 	err := r.Get(ctx, req.NamespacedName, component)
 	if err != nil {
-		logger.Error(err, "Failed to get component for", "req", req.NamespacedName)
```
dirgim, line 64: I think that these changes should go into the [buildpipeline controller](https://github.com/redhat-appstudio/integration-service/blob/main/controllers/buildpipeline/buildpipeline_controller.go) because STONEINTG-645 is geared towards the missing component in the build pipeline data.

14rcole: I thought putting them in the component controller seemed strange - guess I should have voiced that concern.  That said, the ticket says "this error that's shown below is the expected state, because not every build pipeline on cluster belongs to RHTAP so they don't have a component associated with them via the component label."  It seems like the change should go [here](https://github.com/redhat-appstudio/integration-service/blob/41b7f7cdd0c5503ca5d013bad317ef09fa19c378/controllers/buildpipeline/buildpipeline_controller.go#L82-L87), correct?

dirgim: Yup, that should be the place for it.

14rcole: @MartinBasti Since you created the original ticket, do you have any objections?

MartinBasti: no objections, I agree, ticket is about build pipeline

controllers/buildpipeline/buildpipeline_controller.go
```diff
@@ -88,11 +88,15 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	})
 	if err != nil {
 		if errors.IsNotFound(err) {
+			logger.Info("Failed to get component for build pipeline - skipping processing, it's not an appstudio pipeline", "name", pipelineRun.Name, "namespace", pipelineRun.Namespace)
```
dirgim, line 91: Thinking about this, shouldn't the non-appstudio pipelineRuns be missing the component label in the first place?
In that case, the `IsNotFound` error shouldn't be possible for those pipelineRuns, the component should be set to nil [at this point](https://github.com/redhat-appstudio/integration-service/blob/eb807b310184c9b9a29a910140edfd473ef8fc9b/controllers/buildpipeline/buildpipeline_controller.go#L86) instead.

14rcole: That's true, we also handle that case [here](https://github.com/redhat-appstudio/integration-service/blob/eb807b310184c9b9a29a910140edfd473ef8fc9b/controllers/buildpipeline/buildpipeline_controller.go#L99) but if we want to we can do some error handling above instead of just failing silently

MartinBasti: I'm confused here, it seems that invalid component case is already handled well, and we cannot really say if that is caused by not being RHTAP pipeline or removal of component.

Then if component label is missing, we just continue processing it, there is just single condition that creation of snapshot is skipped. Why do we even process PLR without component and don't just stop it here in reconcile initialization?

dirgim: Agreed, I think we should just stop processing if the component label for the build pipelineRun is missing.

14rcole: Based on this conversation and the discussion on Slack from earlier today I've made the following changes
1. Stop reconciliation if `GetComponentFromPipelineRun()` returns `nil, nil`
2. Remove the check to verify that the component is non-nil before calling `GetApplicationFromComponent()`.  It's no longer needed
3. Replace the call to `helpers.HandleLoaderError()` with a logger call and a return. That function just checks for `errors.IsNotFound()` and logs/returns accordingly, which we do in the lines above already.
4. Remove the couple lines of dead code in the buildpipeline adapter that checks if the component is nil

This turned from a bugfix into more of a cleanup PR and I'll update the commit message accordingly

controllers/buildpipeline/buildpipeline_controller.go
```diff
@@ -92,17 +92,21 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 				return ctrl.Result{}, err
 			}
 		}
-		return helpers.HandleLoaderError(logger, err, "Component", "PipelineRun")
+		logger.Error(err, "Failed to get component for build pipeline", "name", pipelineRun.Name, "namespace", pipelineRun.Namespace)
+		return ctrl.Result{}, nil
```
hongweiliu17, line 96: Should we `return ctrl.Result{}, err` here, then reconcile the build PLR again later?

14rcole: We should.  I think on linke 92 we should be returning `ctrl.Result{}, nil` too. I don't think we want to re-reconcile if the component can't be found.  Do you agree?

dheerajodha: If we return `ctrl.Result{}, nil` on line 92, wouldn't we have a build PLR, with a finalizer attached to it, left hanging in the cluster?

dirgim: > If we return `ctrl.Result{}, nil` on line 92, wouldn't we have a build PLR, with a finalizer attached to it, left hanging in the cluster?

Agreed, on line 92 an error happened while trying to remove the finalizer, so our reconciliation didn't do everything it needed to do, hence the error being returned to requeue a subsequent reconciliation attempt.

dirgim: Tbh I'm in favor in leaving the `return helpers.HandleLoaderError(logger, err, "Component", "PipelineRun")` call here instead of making the current change, because that will handle the case when the component is not found as well as when there's a different error finding it.

dirgim: Right we do handle the removal of the finalizer above currently, but not the logging and returning `ctrl.Result{}, nil`.

We could add another `return ctrl.Result{}, nil` to line 93 to stop reconciliation and prevent requeuing when the component is noth there, but I don't think it's worth adding when `HandleLoaderError` would do the same thing.

14rcole: I see what you mean.  I've added back the call to `HandleLoaderError`

controllers/buildpipeline/buildpipeline_controller.go
dirgim: Thanks, I think this change will correctly handle the original issue in [STONEINTG-645](https://issues.redhat.com/browse/STONEINTG-645).

controllers/buildpipeline/buildpipeline_controller.go
```diff
@@ -92,17 +92,20 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 				return ctrl.Result{}, err
 			}
 		}
-		return helpers.HandleLoaderError(logger, err, "Component", "PipelineRun")
+		return helpers.HandleLoaderError(logger, err, "component", "pipelineRun")
+	} else if component == nil {
```
dirgim, line 96: Looks like we're missing coverage for this condition, it would probably be good to add an unit test that checks that we don't error out for a pipelineRun without the component label.

dirgim: Nevermind, we already have a test for that, I misread the codecov report.

MartinBasti (approved):

PR #421, opened 2023-11-14
fix(STONEINTG-580): handle invalid environments

Josh-Everett:
This patch adds a check for verifying the environemnt used for the creation of an ITS is a valid kubernetes object. When an invalid environment is found to be refrenced by the snapshot controller during the generation of an ITS, the snapshot controller will stop reconcilliation instead of requeuing with an error.

### Maintainers will complete the following section

### DIFF
--- a/controllers/snapshot/snapshot_adapter.go
+++ b/controllers/snapshot/snapshot_adapter.go
@@ -18,6 +18,7 @@ package snapshot
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"reflect"
 	"strings"
@@ -306,6 +307,19 @@ func (a *Adapter) EnsureCreationOfEphemeralEnvironments() (controller.OperationR
 				a.logger.Error(err, "environment doesn't exist in the same namespace as integrationScenario at all, try again after creating environment",
 					"integrationTestScenario.Name", integrationTestScenario.Name)
 				return controller.StopProcessing()
+			}
+			if clienterrors.IsInvalid(err) {
+
+				if !gitops.IsSnapshotMarkedAsFailed(a.snapshot) {
+					errKind, errName := GetDetailsFromStatusError(err)
+					a.logger.Error(err, fmt.Sprintf("Resource %s (%s) referenced by integrationTestScenario is invalid", errName, errKind), "integrationTestScenario.Name", integrationTestScenario.Name)
+					_, err = gitops.MarkSnapshotAsFailed(a.client, a.context, a.snapshot, fmt.Sprintf("Resource %s (%s) associated with integrationTestScenario %s is invalid: %s.", errName, errKind, integrationTestScenario.Name, err))
+					if err != nil {
+						a.logger.Error(err, "Failed to Update Snapshot status")
+						return controller.RequeueWithError(err)
+					}
+					return controller.StopProcessing()
+				}
 			}
 			return controller.RequeueWithError(fmt.Errorf("failed to ensure existence of the environment for scenario %s: %w", integrationTestScenario.Name, err))
 		}
@@ -879,4 +893,17 @@ func (a *Adapter) writeIntegrationTestStatusAtError(sits *intgteststat.SnapshotI
 func shouldScenarioRunInEphemeralEnv(scenario *v1beta1.IntegrationTestScenario) bool {
 	// non-empty environment defined in scenario resource means to run in ephemeral env
 	return !reflect.ValueOf(scenario.Spec.Environment).IsZero()
+}
+
+func GetDetailsFromStatusError(err error) (string, string) {
+	var (
+		errKind   string = "<unknown>"
+		errName   string = "<unknown>"
+		statusErr *clienterrors.StatusError
+	)
+	if ok := errors.As(err, &statusErr); ok {
+		errKind = statusErr.Status().Details.Kind
+		errName = statusErr.Status().Details.Name
+	}
+	return errKind, errName
 }
--- a/controllers/snapshot/snapshot_adapter_test.go
+++ b/controllers/snapshot/snapshot_adapter_test.go
@@ -1502,6 +1502,60 @@ var _ = Describe("Snapshot Adapter", Ordered, func() {
 
 	})
 
+	When("An environment defined in ITS is invalid", func() {
+		var (
+			buf bytes.Buffer
+		)
+
+		BeforeAll(func() {
+
+			log := helpers.IntegrationLogger{Logger: buflogr.NewWithBuffer(&buf)}
+			adapter = NewAdapter(hasSnapshotPR, hasApp, hasComp, log, loader.NewMockLoader(), k8sClient, ctx)
+			adapter.context = toolkit.GetMockedContext(ctx, []toolkit.MockData{
+				{
+					ContextKey: loader.ApplicationContextKey,
+					Resource:   hasApp,
+				},
+				{
+					ContextKey: loader.ComponentContextKey,
+					Resource:   hasComp,
+				},
+				{
+					ContextKey: loader.SnapshotContextKey,
+					Resource:   hasSnapshotPR,
+				},
+				{
+					ContextKey: loader.EnvironmentContextKey,
+					Resource:   env,
+				},
+				{
+					ContextKey: loader.SnapshotEnvironmentBindingContextKey,
+					Resource:   nil,
+				},
+				{
+					ContextKey: loader.ApplicationComponentsContextKey,
+					Resource:   []applicationapiv1alpha1.Component{*hasComp},
+				},
+				{
+					ContextKey: loader.AllIntegrationTestScenariosContextKey,
+					Resource:   []v1beta1.IntegrationTestScenario{*integrationTestScenario},
+				},
+				{
+					ContextKey: loader.DeploymentTargetClassContextKey,
+					Resource:   deploymentTargetClass,
+					Err:        errors.NewInvalid(deploymentTargetClass.GetObjectKind().GroupVersionKind().GroupKind(), "invalid environment", nil),
+				},
+			})
+		})
+
+		It("Ensure stopping processing when environment defined in ITS is invalid", func() {
+			result, err := adapter.EnsureCreationOfEphemeralEnvironments()
+			Expect(result.CancelRequest && !result.RequeueRequest && err == nil).To(BeTrue())
+			expectedLogEntry := "is invalid"
+			Expect(buf.String()).Should(ContainSubstring(expectedLogEntry))
+		})
+	})
+
 	Describe("EnsureRerunPipelineRunsExist", func() {
 
 		When("manual re-run of scenario using static env is trigerred", func() {


14rcole (approved):

controllers/snapshot/snapshot_adapter_test.go
```diff
@@ -1476,6 +1477,83 @@ var _ = Describe("Snapshot Adapter", Ordered, func() {
...
+					ContextKey: loader.ComponentContextKey,
+					Resource:   hasComp,
+				},
+				{
+					ContextKey: loader.SnapshotContextKey,
+					Resource:   hasSnapshotPR,
+				},
+				{
+					ContextKey: loader.EnvironmentContextKey,
+					Err:        errors.NewInvalid(invalidEnv.GetObjectKind().GroupVersionKind().GroupKind(), "invalid environment", nil),
```
dirgim, line 1527: With the current mocked error, the unit tests fails when calling the GetAllEnvironments [here](https://github.com/redhat-appstudio/integration-service/blob/5bd1ba4b8796e93ec2afeb9a3267738ff2156027/controllers/snapshot/snapshot_adapter.go#L290), because the `invalid` error happens and gets returned before the condition in this PR is checked.

dirgim (approved):
LGTM

controllers/snapshot/snapshot_adapter.go
```diff
@@ -306,6 +306,17 @@ func (a *Adapter) EnsureCreationOfEphemeralEnvironments() (controller.OperationR
 					"integrationTestScenario.Name", integrationTestScenario.Name)
 				return controller.StopProcessing()
 			}
+			if clienterrors.IsInvalid(err) {
+				a.logger.Error(err, "environment or deploymentTargetClass referenced by integrationScenario is invalid", "integrationTestScenario.Name", integrationTestScenario.Name)
+				if !gitops.IsSnapshotMarkedAsFailed(a.snapshot) {
+					_, err = gitops.MarkSnapshotAsFailed(a.client, a.context, a.snapshot,
+						fmt.Sprintf("The environment or deploymentTargetClass associated with snapshot %s is invalid.", a.snapshot.Spec.Application))
```
MartinBasti, line 313: Should we also propagate the error also here, to show users why it's invalid?

controllers/snapshot/snapshot_adapter.go
```diff
@@ -306,6 +306,17 @@ func (a *Adapter) EnsureCreationOfEphemeralEnvironments() (controller.OperationR
 					"integrationTestScenario.Name", integrationTestScenario.Name)
 				return controller.StopProcessing()
 			}
+			if clienterrors.IsInvalid(err) {
+				a.logger.Error(err, "environment or deploymentTargetClass referenced by integrationScenario is invalid", "integrationTestScenario.Name", integrationTestScenario.Name)
```
MartinBasti, line 310: We can provide what exactly is wrong, which kind and name

```diff
diff --git a/controllers/snapshot/snapshot_adapter.go b/controllers/snapshot/snapshot_adapter.go
index 86cc817..9518bb3 100644
--- a/controllers/snapshot/snapshot_adapter.go
+++ b/controllers/snapshot/snapshot_adapter.go
@@ -18,6 +18,7 @@ package snapshot
 
 import (
        "context"
+       "errors"
        "fmt"
        "reflect"
        "strings"
@@ -307,10 +308,20 @@ func (a *Adapter) EnsureCreationOfEphemeralEnvironments() (controller.OperationR
                                return controller.StopProcessing()
                        }
                        if clienterrors.IsInvalid(err) {
-                               a.logger.Error(err, "environment or deploymentTargetClass referenced by integrationScenario is invalid", "integrationTestScenario.Name", integrationTestScenario.Name)
+                               var (
+                                       errKind   string = "<unknown>"
+                                       errName   string = "<unknown>"
+                                       statusErr *clienterrors.StatusError
+                               )
+
+                               if ok := errors.As(err, &statusErr); ok {
+                                       errKind = statusErr.Status().Details.Kind
+                                       errName = statusErr.Status().Details.Name
+                               }
+                               a.logger.Error(err, fmt.Sprintf("Object %s (%s) referenced by integrationScenario is invalid", errName, errKind), "integrationTestScenario.Name", integrationTestScenario.Name)
                                if !gitops.IsSnapshotMarkedAsFailed(a.snapshot) {
                                        _, err = gitops.MarkSnapshotAsFailed(a.client, a.context, a.snapshot,
-                                               fmt.Sprintf("The environment or deploymentTargetClass associated with snapshot %s is invalid.", a.snapshot.Spec.Application))
+                                               fmt.Sprintf("Object %s (%s) associated with snapshot %s is invalid: %s.", errName, errKind, a.snapshot.Spec.Application, err))
                                        if err != nil {
                                                a.logger.Error(err, "Failed to Update Snapshot status")
                                        }

```

MartinBasti: s/Object/Resource

controllers/snapshot/snapshot_adapter.go
MartinBasti: `associated with snapshot` it's not associated by snapshot, but by integration test scenario (at least environment)

controllers/snapshot/snapshot_adapter.go
```diff
@@ -306,6 +307,27 @@ func (a *Adapter) EnsureCreationOfEphemeralEnvironments() (controller.OperationR
...
+						errKind   string = "<unknown>"
+						errName   string = "<unknown>"
+						statusErr *clienterrors.StatusError
+					)
+					if ok := errors.As(err, &statusErr); ok {
+						errKind = statusErr.Status().Details.Kind
+						errName = statusErr.Status().Details.Name
+					}
+					a.logger.Error(err, fmt.Sprintf("Resource %s (%s) referenced by integrationTestScenario is invalid", errName, errKind), "integrationTestScenario.Name", integrationTestScenario.Name)
+					_, err = gitops.MarkSnapshotAsFailed(a.client, a.context, a.snapshot, fmt.Sprintf("Object %s (%s) associated with snapshot %s is invalid: %s.", errName, errKind, a.snapshot.Spec.Application, err))
```
hongweiliu17, line 324: It might be ```suggestion
					_, err = gitops.MarkSnapshotAsFailed(a.client, a.context, a.snapshot, fmt.Sprintf("Object %s (%s) associated with snapshot %s is invalid: %s.", errName, errKind, a.snapshot.Name, err))
``` because it says `snapshot %s`

**Review by hongweiliu17 — APPROVED** [review]
LGTM

File: controllers/snapshot/snapshot_adapter.go [hunk 5]
```diff
@@ -306,6 +307,27 @@ func (a *Adapter) EnsureCreationOfEphemeralEnvironments() (controller.OperationR
 					"integrationTestScenario.Name", integrationTestScenario.Name)
 				return controller.StopProcessing()
 			}
+			if clienterrors.IsInvalid(err) {
+
+				//a.logger.Error(err, "environment or deploymentTargetClass referenced by integrationTestScenario is invalid", "integrationTestScenario.Name", integrationTestScenario.Name)
```
MartinBasti, line 312: This can be removed

controllers/snapshot/snapshot_adapter.go
MartinBasti:
```suggestion
					_, err = gitops.MarkSnapshotAsFailed(a.client, a.context, a.snapshot, fmt.Sprintf("Resource %s (%s) associated with snapshot %s is invalid: %s.", errName, errKind, a.snapshot.Spec.Application, err))
```

controllers/snapshot/snapshot_adapter.go
```diff
@@ -306,6 +307,27 @@ func (a *Adapter) EnsureCreationOfEphemeralEnvironments() (controller.OperationR
...
+					)
+					if ok := errors.As(err, &statusErr); ok {
+						errKind = statusErr.Status().Details.Kind
+						errName = statusErr.Status().Details.Name
+					}
+					a.logger.Error(err, fmt.Sprintf("Resource %s (%s) referenced by integrationTestScenario is invalid", errName, errKind), "integrationTestScenario.Name", integrationTestScenario.Name)
+					_, err = gitops.MarkSnapshotAsFailed(a.client, a.context, a.snapshot, fmt.Sprintf("Object %s (%s) associated with snapshot %s is invalid: %s.", errName, errKind, a.snapshot.Spec.Application, err))
+					if err != nil {
+						a.logger.Error(err, "Failed to Update Snapshot status")
+					}
```
dheerajodha, line 327: Do we want to call `controller.RequeueWithError(err)` here? or else, if there were an error while marking the snapshot as failed, the controller wouldn't retry and stop reconciling the snapshot. In the end, we could have snapshots in the cluster that aren't marked as failed yet, it could be misleading for the users.

controllers/snapshot/snapshot_adapter.go
Josh-Everett: Error propagation and correct info on relation to snapshot vs ITS in [b674ec](https://github.com/redhat-appstudio/integration-service/pull/421/commits/b674ecd242f54c9700dd14f99c04946a91f6af72)

controllers/snapshot/snapshot_adapter.go
Josh-Everett: This was useful, thank you. Should be fixed in [b674ec](https://github.com/redhat-appstudio/integration-service/pull/421/commits/b674ecd242f54c9700dd14f99c04946a91f6af72)

controllers/snapshot/snapshot_adapter.go
Josh-Everett: removed in [b674ec](https://github.com/redhat-appstudio/integration-service/pull/421/commits/b674ecd242f54c9700dd14f99c04946a91f6af72)

controllers/snapshot/snapshot_adapter.go
Josh-Everett: Ack yes you are correct, thanks for the good catch @dheerajodha. Updated in [b674ec](https://github.com/redhat-appstudio/integration-service/pull/421/commits/b674ecd242f54c9700dd14f99c04946a91f6af72)

controllers/snapshot/snapshot_adapter.go
```diff
@@ -307,6 +308,27 @@ func (a *Adapter) EnsureCreationOfEphemeralEnvironments() (controller.OperationR
...
+				if !gitops.IsSnapshotMarkedAsFailed(a.snapshot) {
+					var (
+						errKind   string = "<unknown>"
+						errName   string = "<unknown>"
+						statusErr *clienterrors.StatusError
+					)
+					if ok := errors.As(err, &statusErr); ok {
+						errKind = statusErr.Status().Details.Kind
+						errName = statusErr.Status().Details.Name
+					}
```
dirgim, line 322: nit/suggestion: this way of extracting basic info from errors seems like it could be useful in other places too, so it could be a good idea to extract it into a generic function.

MartinBasti: I'm not sure if this can be a generic function, it would work only for StatusError, but for better readability, we can introduce `GetDetailsFromStatusError` to return `Details`

controllers/snapshot/snapshot_adapter.go
```diff
@@ -307,6 +308,27 @@ func (a *Adapter) EnsureCreationOfEphemeralEnvironments() (controller.OperationR
...
+					a.logger.Error(err, fmt.Sprintf("Resource %s (%s) referenced by integrationTestScenario is invalid", errName, errKind), "integrationTestScenario.Name", integrationTestScenario.Name)
+					_, err = gitops.MarkSnapshotAsFailed(a.client, a.context, a.snapshot, fmt.Sprintf("Resource %s (%s) associated with snapshot %s is invalid: %s.", errName, errKind, a.snapshot.Name, err))
```
MartinBasti, line 324: We should inform user which ITS is invalid instead of snapshot, Status is recorded in snapshot, so user knows which snapshot is affected, but user doesn't know which ITS is it causing. Error log above provides more useful info than msg in snapshot IMHO.

Josh-Everett: ack yes good point, changed in most recent patch.

controllers/snapshot/snapshot_adapter.go
Josh-Everett: I refactored out that bit to make things a bit more readable, but not sure how I would implement a generic here. Im interested in your thoughts @dirgim & @MartinBasti

controllers/snapshot/snapshot_adapter.go
MartinBasti: I don't see it

controllers/snapshot/snapshot_adapter.go
dirgim: Yes, agreed.

controllers/snapshot/snapshot_adapter.go

controllers/snapshot/snapshot_adapter.go
Josh-Everett: Updated to reflect resource & a more clear error message that refers to the ITS as opposed to the snapshot [here](https://github.com/redhat-appstudio/integration-service/commit/d03da3f685c1c4b4d3e036780b3a2401a384bf03#diff-8e718c6bf06f754db7e9e94c6f74b5c6704941dff95d637289e3c4cc194c95a0R316)

14rcole (approved):

PR #436, opened 2023-11-21
fix(rhtapbugs-975): point test plr params to source repo

14rcole:
Previously the git url for a PR job was set to the target branch, causing errors when pulling the revision to be tested. This change sets the git URL in the integration PLR to the source branch git url instead

### DIFF
--- a/tekton/utils.go
+++ b/tekton/utils.go
@@ -158,26 +158,26 @@ func GetOutputImageDigest(object client.Object) (string, error) {
 	return "", fmt.Errorf("couldn't find the IMAGE_DIGEST TaskRun result")
 }
 
-// GetComponentSourceGitUrl returns a string containing the CHAINS-GIT_URL result value from a given PipelineRun.
+// GetComponentSourceGitUrl returns a string containing the git-url params value from a given PipelineRun.
 func GetComponentSourceGitUrl(object client.Object) (string, error) {
 	if pipelineRun, ok := object.(*tektonv1.PipelineRun); ok {
-		for _, pipelineResult := range pipelineRun.Status.Results {
-			if pipelineResult.Name == "CHAINS-GIT_URL" {
-				return pipelineResult.Value.StringVal, nil
+		for _, param := range pipelineRun.Spec.Params {
+			if param.Name == "git-url" {
+				return param.Value.StringVal, nil
 			}
 		}
 	}
-	return "", fmt.Errorf("couldn't find the CHAINS-GIT_URL PipelineRun result")
+	return "", fmt.Errorf("couldn't find the 'git-url' PipelineRun parameter")
 }
 
 // GetComponentSourceGitCommit returns a string containing the CHAINS-GIT_COMMIT result value from a given PipelineRun.
 func GetComponentSourceGitCommit(object client.Object) (string, error) {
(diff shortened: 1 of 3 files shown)


tekton/utils.go
```diff
@@ -161,9 +161,9 @@ func GetOutputImageDigest(object client.Object) (string, error) {
 // GetComponentSourceGitUrl returns a string containing the CHAINS-GIT_URL result value from a given PipelineRun.
 func GetComponentSourceGitUrl(object client.Object) (string, error) {
 	if pipelineRun, ok := object.(*tektonv1.PipelineRun); ok {
-		for _, pipelineResult := range pipelineRun.Status.Results {
-			if pipelineResult.Name == "CHAINS-GIT_URL" {
-				return pipelineResult.Value.StringVal, nil
+		for _, param := range pipelineRun.Spec.Params {
```
dirgim, line 164: I'm not sure that we want to do make this change for all types of build pipelineRuns.
If we don't, we should have a separate function that fetches the source git URL and is only used when the build pipelineRuns is running for a `pull_request` event.

MartinBasti: IMO it's safe to use git-url in push events, as it's supposed to be source url. In case of push events source and target URL in PAC are the same value.

dirgim: Okay, in that case I think we should be fine as long as we document that the integration service now depends on thhe git-url parameter being present on the build pipelineRun. I think at least [this section](https://redhat-appstudio.github.io/book/book/integration-service.html#image-extraction-details) in the book repo should be updated.

tekton/utils_test.go
```diff
@@ -88,14 +94,14 @@ var _ = Describe("Utils", func() {
...
 		git_url, _ := tekton.GetComponentSourceGitUrl(pipelineRun)
-		if git_url != "https://github.com/devfile-samples/devfile-sample-java-springboot-basic" {
-			Fail(fmt.Sprintf("Expected git_url is https://github.com/devfile-samples/devfile-sample-java-springboot-basic, but got %s", git_url))
+		if git_url != "https://github.com/upstream-user/devfile-sample-java-springboot-basic" {
+			Fail(fmt.Sprintf("Expected git_url is https://github.com/upstream-user/devfile-sample-java-springboot-basic, but got %s", git_url))
 		}
 		klog.Infoln("Got expected git_url")
 	})
 
 	It("can return err when can't find result for CHAINS-GIT_URL", func() {
```
dirgim, line 103: I think this description should be updated to say that we can't find the `git-url` pipeline parameter.

tekton/utils.go
14rcole: Can do.  I'm also going to update [this function](https://github.com/redhat-appstudio/integration-service/blob/main/tekton/utils.go#L174) to pull from `Spec.Params`.  I'm bringing up another cluster to make there's a param called `revision` or `commit` or something similar.  I think it's best to get both the source git url and commit from a similar place. It'll make the code easier to understand.

14rcole: [Book PR](https://github.com/redhat-appstudio/book/pull/151)

14rcole: It is now, thanks for the catch. My last push failed.

dirgim (approved):
LGTM

PR #437, opened 2023-11-21
fix(STONEINTG-689): skip update if *StatusAdapter fails

MartinBasti:
If we fail to create object with *StatusAdapter, we should skip update of PR status. This error is usually non-recoverable it only retriggers reconciliation and blocks other updates.

## Maintainers will complete the following section

- [ ] Commit messages are descriptive enough ([hints](https://www.freecodecamp.org/news/how-to-write-better-git-commit-messages/))
- [ ] Code coverage from testing does not decrease and new code is covered ([check the PR coverage on codecov](https://app.codecov.io/gh/redhat-appstudio/integration-service/pulls))
- [ ] [Controllers diagrams](https://github.com/redhat-appstudio/integration-service/tree/main/docs) are updated when PR changes controllers code  (if applicable)

PR #438, opened 2023-11-22
fix(STONEINTG-689): add PR last update snapshot annotation

jsztuka:

### DIFF
--- a/controllers/statusreport/statusreport_adapter.go
+++ b/controllers/statusreport/statusreport_adapter.go
@@ -19,6 +19,7 @@ package statusreport
 import (
 	"context"
 	"fmt"
+
 	"github.com/redhat-appstudio/integration-service/api/v1beta1"
 	"github.com/redhat-appstudio/integration-service/gitops"
 	"github.com/redhat-appstudio/integration-service/metrics"
--- a/gitops/snapshot.go
+++ b/gitops/snapshot.go
@@ -57,6 +57,9 @@ const (
 	// SnapshotTestScenarioLabel contains json data with test results of the particular snapshot
 	SnapshotTestsStatusAnnotation = "test.appstudio.openshift.io/status"
 
+	// SnapshotPRLastUpdate contains timestamp of last time PR was updated
+	SnapshotPRLastUpdate = "test.appstudio.openshift.io/pr-last-update"
+
 	// BuildPipelineRunPrefix contains the build pipeline run related labels and annotations
 	BuildPipelineRunPrefix = "build.appstudio"
 
@@ -708,5 +711,27 @@ func RemoveIntegrationTestRerunLabel(adapterClient client.Client, ctx context.Co
 		return fmt.Errorf("failed to patch snapshot: %w", err)
 	}
 
+	return nil
+}
+
+func GetLatestUpdateTime(snapshot *applicationapiv1alpha1.Snapshot) (time.Time, error) {
+	latestUpdateTime := snapshot.GetAnnotations()[SnapshotPRLastUpdate]
+	if latestUpdateTime == "" {
+		return time.Time{}, nil
+	}
+	t := time.Time{}
+	err := t.UnmarshalText([]byte(latestUpdateTime))
+	if err != nil {
+		return t, fmt.Errorf("failed to unmarshal time annotation: %w", err)
+	}
+	return t, nil
+}
+
+func SetLatestUpdateTime(snapshot *applicationapiv1alpha1.Snapshot, t time.Time) error {
+	byteVal, _ := t.MarshalText()
+	err := metadata.SetAnnotation(snapshot, SnapshotPRLastUpdate, string(byteVal))
+	if err != nil {
+		return fmt.Errorf("failed to set annotation for snapshot %w", err)
+	}
 	return nil
 }
--- a/status/reporters.go
+++ b/status/reporters.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"fmt"
 	"strconv"
+	"time"
 
 	"github.com/go-logr/logr"
 	ghapi "github.com/google/go-github/v45/github"
@@ -236,19 +237,13 @@ func (cru *CheckRunStatusUpdater) UpdateStatus(ctx context.Context, integrationT
 		}
 	} else {
 		cru.logger.Info("found existing checkrun", "existingCheckRun", existingCheckrun)
-		if cru.ghClient.IsUpdateNeeded(existingCheckrun, checkRun) {
-			cru.logger.Info("found existing check run with the same ExternalID but different conclusion/status, updating checkrun for scenario test status of snapshot",
-				"snapshot.NameSpace", cru.snapshot.Namespace, "snapshot.Name", cru.snapshot.Name, "scenarioName", integrationTestStatusDetail.ScenarioName, "checkrun.ExternalID", checkRun.ExternalID)
-			err = cru.ghClient.UpdateCheckRun(ctx, *existingCheckrun.ID, checkRun)
-			if err != nil {
-				cru.logger.Error(err, "failed to update checkrun",
-					"checkRun", checkRun)
-				return err
-			}
-		} else {
-			cru.logger.Info("found existing check run with the same ExternalID and conclusion/status, no need to update checkrun for scenario test status of snapshot",
-				"snapshot.NameSpace", cru.snapshot.Namespace, "snapshot.Name", cru.snapshot.Name, "scenarioName", integrationTestStatusDetail.ScenarioName, "checkrun.ExternalID", checkRun.ExternalID)
+		err = cru.ghClient.UpdateCheckRun(ctx, *existingCheckrun.ID, checkRun)
+		if err != nil {
+			cru.logger.Error(err, "failed to update checkrun",
+				"checkRun", checkRun)
+			return err
 		}
+
 	}
 	return nil
 }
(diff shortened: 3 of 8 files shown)


controllers/statusreport/statusreport_adapter.go
```diff
@@ -19,6 +19,7 @@ package statusreport
 import (
 	"context"
 	"fmt"
+
```
jsztuka, line 22: IDE loves to add spaces for no reason whatsoever, cant remove it :(

gitops/snapshot.go
```diff
@@ -703,3 +706,21 @@ func RemoveIntegrationTestRerunLabel(adapterClient client.Client, ctx context.Co
 
 	return nil
 }
+
+func GetLatestUpdateTime(snapshot *applicationapiv1alpha1.Snapshot) (time.Time, error) {
+	//specific time format choosen, regular format errors out on parsing
+	dateString := "2005/01/02/15:04:05"
+	latestUpdateTime := snapshot.GetAnnotations()[SnapshotPRLastUpdate]
+	if latestUpdateTime == "" {
+		err := metadata.SetAnnotation(snapshot, SnapshotPRLastUpdate, time.Now().String())
```
MartinBasti, line 715: Why Get method should update something?
This is against SRP and it doesn't belong here, setting annotation must be done somewhere else.

status/reporters.go
```diff
@@ -480,6 +480,21 @@ func (r *GitHubReporter) ReportStatusForSnapshot(k8sClient client.Client, ctx co
...
+				return err
+			}
+			timeDiff := latestUpdateTime.Sub(integrationTestStatusDetail.LastUpdateTime)
+			if timeDiff < 0 {
+				err := metadata.SetAnnotation(snapshot, gitops.SnapshotPRLastUpdate, integrationTestStatusDetail.LastUpdateTime.String())
+				if err != nil {
+					logger.Error(err, "failed to set annotation for snapshot",
+						"snapshot.NameSpace", snapshot.Namespace, "snapshot.Name", snapshot.Name)
+					return err
+				}
```
MartinBasti, line 496: What's the purpose if this? It has no effect to code execution, it just update timestamp and still executes reporting into GH

MartinBasti (changes requested):

status/reporters.go
jsztuka: Updated

status/reporters.go
```diff
@@ -480,6 +488,14 @@ func (r *GitHubReporter) ReportStatusForSnapshot(k8sClient client.Client, ctx co
 
 		for _, integrationTestStatusDetail := range integrationTestStatusDetails {
 			integrationTestStatusDetail := *integrationTestStatusDetail // G601
+			if latestUpdateTime.Before(integrationTestStatusDetail.LastUpdateTime) {
+				logger.Info("IntegrationTest: %w contains new changes", integrationTestStatusDetail.ScenarioName)
+				if newLatestUpdateTime.Before(integrationTestStatusDetail.LastUpdateTime) {
+					newLatestUpdateTime = integrationTestStatusDetail.LastUpdateTime
+				}
+			} else {
+				logger.Info("IntegrationTest: %w contains no new changes", integrationTestStatusDetail.ScenarioName)
```
dirgim, line 497: Should we call a `continue` after this or something similar in order to skip executing creating/updating a checkRun?

jsztuka: Thats what I had there in a first place, it invalid memory address or nil pointer dereference in unit tests. I thought that changes in https://github.com/redhat-appstudio/integration-service/pull/437 should take care of that.

status/reporters.go
```diff
@@ -452,6 +452,14 @@ func (r *GitHubReporter) ReportStatusForSnapshot(k8sClient client.Client, ctx co
 		return fmt.Errorf("sha label not found %q", gitops.PipelineAsCodeSHALabel)
 	}
 	integrationTestStatusDetails := statuses.GetStatuses()
+	latestUpdateTime, _ := gitops.GetLatestUpdateTime(snapshot)
```
MartinBasti, line 455: Maybe we should at least to log err in error log when it fail to fetch timestamp

MartinBasti: it's actually there, this line just miss err variable
```suggestion
	latestUpdateTime, err := gitops.GetLatestUpdateTime(snapshot)
```

status/reporters.go
```diff
@@ -452,6 +452,14 @@ func (r *GitHubReporter) ReportStatusForSnapshot(k8sClient client.Client, ctx co
...
+	newLatestUpdateTime := latestUpdateTime
+	if err != nil {
+		logger.Error(err, "failed to get latest update annotation for snapshot",
+			"snapshot.NameSpace", snapshot.Namespace, "snapshot.Name", snapshot.Name)
+		return err
```
MartinBasti, line 460: Maybe instead returning error here, we could just use 0 timestamp and let it selfheal

status/reporters.go
MartinBasti: if we don't put `continue` here, we don't even need this PR

status/reporters.go
jsztuka: Do you agree we keep the error msg? or we want to log it as info?

status/reporters.go
MartinBasti: Also we shouldn't log this, it will flood logs with the same msg

status/reporters_test.go
```diff
@@ -557,13 +558,13 @@ var _ = Describe("GitHubReporter", func() {
 
 		It("reports snapshot tests status via CheckRuns", func() {
 			// Create an pending CheckRun
-			hasSnapshot.Annotations["test.appstudio.openshift.io/status"] = "[{\"scenario\":\"scenario1\",\"status\":\"Pending\",\"lastUpdateTime\":\"2023-08-26T17:57:49+02:00\",\"details\":\"pending\"}]"
+			hasSnapshot.Annotations["test.appstudio.openshift.io/status"] = "[{\"scenario\":\"scenario1\",\"status\":\"Pending\",\"lastUpdateTime\":\"2023-08-26T17:57:49Z\",\"details\":\"pending\"}]"
```
MartinBasti, line 561: Why this change?

status/reporters.go
MartinBasti: error sounds fine

status/reporters_test.go
MartinBasti: I got it, because it was not updating due timezone mismatch

MartinBasti (approved):

gitops/snapshot.go
```diff
@@ -703,3 +706,25 @@ func RemoveIntegrationTestRerunLabel(adapterClient client.Client, ctx context.Co
 
 	return nil
 }
+
+func GetLatestUpdateTime(snapshot *applicationapiv1alpha1.Snapshot) (time.Time, error) {
```
dirgim, line 710: While we do exercise these functions somewhat in the status/reporters unit tests, I think we'll also need some dedicated unit tests for them to have enough code coverage.

gitops/snapshot.go
```diff
@@ -703,3 +706,25 @@ func RemoveIntegrationTestRerunLabel(adapterClient client.Client, ctx context.Co
...
+	}
+	t := time.Time{}
+	err := t.UnmarshalText([]byte(latestUpdateTime))
+	if err != nil {
+		return t, fmt.Errorf("failed to unmarshal time annotation: %w", err)
+	}
+	return t, nil
+}
+
+func SetLatestUpdateTime(snapshot *applicationapiv1alpha1.Snapshot, t time.Time) error {
```
dirgim, line 723: Same here, I think we'll also need some dedicated unit tests for this to have enough code coverage.

dirgim:
nit: the commit message could do with a description of the changes.

MartinBasti (changes requested):
`IsUpdateNeeded` we probably don't need this anymore, as update detection changed.

MartinBasti:
Here is a potential race condition, because creating of lastUpstaded timestamp entry experiences delay between creation and writing the value into k8s, in some cases, we may skip update in case that another test result has been updated around the same time.

Solution would be to keep records of last update per scenario as JSON in the new annotation. Not sure how much will this race condition happen though.

gitops/snapshot.go
jsztuka: Added for two cases within existing unit tests, would this coverage suffice?

dirgim: I was thinking more along the lines of adding unit tests to the [gitops/snapshot_test.go](https://github.com/redhat-appstudio/integration-service/blob/main/gitops/snapshot_test.go) to satisfy codecov.

MartinBasti:
> `IsUpdateNeeded` we probably don't need this anymore, as update detection changed.

Still valid

jsztuka:
> > `IsUpdateNeeded` we probably don't need this anymore, as update detection changed.
> 
> Still valid

So even after refactoring there: https://github.com/redhat-appstudio/integration-service/pull/448/ we want to remove it?

jsztuka:
> `IsUpdateNeeded` we probably don't need this anymore, as update detection changed.

removed

jsztuka:
IsUpdateNeeded removed since update detection changed. - Do we want to include that information in commit msg?

status/reporters.go
```diff
@@ -701,10 +704,20 @@ func (r *GitHubReporter) ReportStatusForSnapshot(k8sClient client.Client, ctx co
 	}
 
 	for _, integrationTestStatusDetail := range integrationTestStatusDetails {
+		if latestUpdateTime.Before(integrationTestStatusDetail.LastUpdateTime) {
+			logger.Info("IntegrationTest: %w contains new changes", integrationTestStatusDetail.ScenarioName)
```
MartinBasti, line 708:
```suggestion
			logger.Info("Integration Test contains new status updates", "scenario.Name" integrationTestStatusDetail.ScenarioName)
```

jsztuka: updated

MartinBasti (approved):

status/reporters.go
```diff
@@ -480,6 +489,15 @@ func (r *GitHubReporter) ReportStatusForSnapshot(k8sClient client.Client, ctx co
 
 		for _, integrationTestStatusDetail := range integrationTestStatusDetails {
 			integrationTestStatusDetail := *integrationTestStatusDetail // G601
+			if latestUpdateTime.Before(integrationTestStatusDetail.LastUpdateTime) {
```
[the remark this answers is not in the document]
dirgim: Resolving this to unblock merging, #448 has been merged.

PR #440, opened 2023-11-23
fix(STONEINTG-693): stop collisions in buildpipeline adapter tests

dirgim:
* Don't start the controller manager in the controller test suite because that causes it to react to creation events in adapter unit tests and causes collisions when it attempts to add a finalizer to all newy created build pipelineRuns
* Wait for the buildPipeline run resource on the control plane to be fully updated and its finalizer removed before ending finalizer tests to prevent create actions in beforeEach from running into conflicts
* Reduce update actions for build pipelineRuns where they don't require testing resources on the control plane

### DIFF
--- a/controllers/buildpipeline/buildpipeline_adapter_test.go
+++ b/controllers/buildpipeline/buildpipeline_adapter_test.go
@@ -19,6 +19,7 @@ package buildpipeline
 import (
 	"bytes"
 	"reflect"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"time"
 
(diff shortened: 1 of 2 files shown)


dirgim:
I tested this by throttling resources on my local machine to the point where the unit tests were running into the reported issue almost every time, and then applying these changes and running the unit tests 10 times in a row successfully.
I also ran the [Check sources test](https://github.com/redhat-appstudio/integration-service/actions/runs/6968229998/job/18962222454) on this PR 5 times in a row to confirm that it also functions the same in the gitHub actions environment.

hongweiliu17 (approved):

dheerajodha (approved):

PR #443, opened 2023-11-26
fix(STONEINTG-646):  trigger snapshot reconcile for added its

kasemAlem:
This PR to fix race condition when adding another IntegrationTestScenario to an application that already running former one, the new one will bot be triggered. 

to fix it, we identify the untriggered ITS  and then we add Re-run label to snapshot and this will trigger the snapshot reconcile .

more details in Jira [STONEINTG-646](https://issues.redhat.com/browse/STONEINTG-646)

### DIFF
--- a/controllers/statusreport/statusreport_adapter.go
+++ b/controllers/statusreport/statusreport_adapter.go
@@ -121,6 +121,18 @@ func (a *Adapter) EnsureSnapshotFinishedAllTests() (controller.OperationResult, 
 	if !allIntegrationTestsFinished {
 		a.logger.Info("Not all required Integration PipelineRuns finished",
 			"snapshot.Name", a.snapshot.Name)
+
+		// If for the snapshot there is an IntegrationTestScenario that is not triggered, it will add run labebl to snapshot
+		integrationTestScenarioNotTriggered := a.findUntriggeredIntegrationTestFromStatus(integrationTestScenarios, testStatuses)
+		if integrationTestScenarioNotTriggered != "" {
+			a.logger.Info("Detected an integrationTestScenario was not triggered, applying snapshot reconcilation",
+				"integrationTestScenario.Name", integrationTestScenarioNotTriggered)
+			if err = gitops.AddIntegrationTestRerunLabel(a.client, a.context, a.snapshot, integrationTestScenarioNotTriggered); err != nil {
+				return controller.RequeueWithError(err)
+			}
+
+		}
+
 		return controller.ContinueProcessing()
 	}
 
@@ -298,4 +310,17 @@ func (a *Adapter) getComponentSourceFromSnapshotComponent(snapshot *applicationa
 		}
 	}
 	return nil, fmt.Errorf("couldn't find the requested component source info in the given Snapshot")
+}
+
+// findUntriggeredIntegrationTestFromStatus returns name of integrationTestScenario that is not triggered yet.
+func (a *Adapter) findUntriggeredIntegrationTestFromStatus(integrationTestScenarios *[]v1beta1.IntegrationTestScenario, testStatuses *intgteststat.SnapshotIntegrationTestStatuses) string {
+	for _, integrationTestScenario := range *integrationTestScenarios {
+		integrationTestScenario := integrationTestScenario // G601
+		_, ok := testStatuses.GetScenarioStatus(integrationTestScenario.Name)
+		if !ok {
+			return integrationTestScenario.Name
+		}
+
+	}
+	return ""
 }
--- a/controllers/statusreport/statusreport_adapter_test.go
+++ b/controllers/statusreport/statusreport_adapter_test.go
@@ -350,6 +350,62 @@ var _ = Describe("Snapshot Adapter", Ordered, func() {
 			Expect(buf.String()).Should(ContainSubstring(expectedLogEntry))
 		})
 
+		It("testing function findUntriggeredIntegrationTestFromStatus ", func() {
+
+			integrationTestScenarioTest := &v1beta1.IntegrationTestScenario{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "example-pass-test",
+					Namespace: "default",
+
+					Labels: map[string]string{
+						"test.appstudio.openshift.io/optional": "false",
+					},
+				},
+				Spec: v1beta1.IntegrationTestScenarioSpec{
+					Application: hasApp.Name,
+					ResolverRef: v1beta1.ResolverRef{
+						Resolver: "git",
+						Params: []v1beta1.ResolverParameter{
+							{
+								Name:  "url",
+								Value: "https://github.com/redhat-appstudio/integration-examples.git",
+							},
+							{
+								Name:  "revision",
+								Value: "main",
+							},
+							{
+								Name:  "pathInRepo",
+								Value: "pipelineruns/integration_pipelinerun_pass.yaml",
+							},
+						},
+					},
+					Environment: v1beta1.TestEnvironment{
+						Name: "envname",
+						Type: "POC",
+						Configuration: &applicationapiv1alpha1.EnvironmentConfiguration{
+							Env: []applicationapiv1alpha1.EnvVarPair{},
+						},
+					},
+				},
+			}
+
+			// Check when all integrationTestScenarion exist in testStatuses of the snapshot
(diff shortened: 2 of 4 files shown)


controllers/statusreport/statusreport_adapter.go
```diff
@@ -298,3 +310,16 @@ func (a *Adapter) getComponentSourceFromSnapshotComponent(snapshot *applicationa
 	}
 	return nil, fmt.Errorf("couldn't find the requested component source info in the given Snapshot")
 }
+
+// determineIfAllIntegrationTestsNotTriggered returns name of integrationTestScenario that is not triggered yet.
+func (a *Adapter) determineIfAllIntegrationTestsNotTriggered(integrationTestScenarios *[]v1beta1.IntegrationTestScenario, testStatuses *intgteststat.SnapshotIntegrationTestStatuses) string {
+	for _, integrationTestScenario := range *integrationTestScenarios {
+		integrationTestScenario := integrationTestScenario // G601
+		_, ok := testStatuses.GetScenarioStatus(integrationTestScenario.Name)
```
jsztuka, line 318: Don't we want to handle errors in case they occur there?

jsztuka:
Its probably out of scope, but don't we want to cover new functions with unit tests also?

controllers/statusreport/statusreport_adapter.go
```diff
@@ -109,6 +110,17 @@ func (a *Adapter) EnsureSnapshotFinishedAllTests() (controller.OperationResult,
...
+	// If for the snapshot there is an IntegrationTestScenario that is not triggered, it will add run labebl to snapshot
+	integrationTestScenarioNotTriggered := a.determineIfAllIntegrationTestsNotTriggered(integrationTestScenarios, testStatuses)
+	if integrationTestScenarioNotTriggered != "" {
+		a.logger.Info("Detected an integrationTestScenario was not triggered, applying snapshot reconcilation",
+			"integrationTestScenario.Name", integrationTestScenarioNotTriggered)
+		if err = gitops.AddIntegrationTestRerunLabel(a.client, a.context, a.snapshot, integrationTestScenarioNotTriggered); err != nil {
+			return controller.RequeueWithError(err)
+		}
+
+	}
```
dirgim, line 122: I'm thinking that  moving this logic to be inside the condition when not all integration tests finished ([here](https://github.com/redhat-appstudio/integration-service/blob/dd5a5e7abc4e575de00d5b568b95a8f74a13a50d/controllers/statusreport/statusreport_adapter.go#L120) currently) would make sense since there's no need to check this if we determine that everything passed.

controllers/statusreport/statusreport_adapter.go
```diff
@@ -109,6 +110,17 @@ func (a *Adapter) EnsureSnapshotFinishedAllTests() (controller.OperationResult,
 		return controller.RequeueWithError(err)
 	}
 
+	// If for the snapshot there is an IntegrationTestScenario that is not triggered, it will add run labebl to snapshot
+	integrationTestScenarioNotTriggered := a.determineIfAllIntegrationTestsNotTriggered(integrationTestScenarios, testStatuses)
```
MartinBasti, line 114: We don't react to creation events here, so hopefully, it's enough to assume here, that this will only trigger new tests when status changed, and not randomly for any reconciliation for all old snapshots.

Just making note so others can check if I'm not mistaken.

MartinBasti: on the other hand, some defensive programming wouldn't hurt, if we change predicates in future

controllers/statusreport/statusreport_adapter.go
dirgim: I think it wouldn't make sense to add it there as it would hide the purpose of the `determineIfAllIntegrationTestsFinishedAndPassed` function (by making it do more than expected) and would it make harder to maintain in the future. 

That's why I proposed adding it [here](https://github.com/redhat-appstudio/integration-service/blob/dd5a5e7abc4e575de00d5b568b95a8f74a13a50d/controllers/statusreport/statusreport_adapter.go#L120) because at that point we know that not all scenarios finished and can then run the new logic to find missing pipelineRuns in a place that's more visible.

kasemAlem: agree, 
moved it to [here](https://github.com/redhat-appstudio/integration-service/blob/dd5a5e7abc4e575de00d5b568b95a8f74a13a50d/controllers/statusreport/statusreport_adapter.go#L120)

dirgim: Thanks!

gitops/snapshot.go
```diff
@@ -710,3 +710,20 @@ func RemoveIntegrationTestRerunLabel(adapterClient client.Client, ctx context.Co
 
 	return nil
 }
+
+// AddIntegrationTestRerunLabel adding re-run label to snapshot
+func AddIntegrationTestRerunLabel(adapterClient client.Client, ctx context.Context, snapshot *applicationapiv1alpha1.Snapshot, integrationTestScenarioName string) error {
```
MartinBasti, line 715: This function should have unittest

kasemAlem: added one

controllers/statusreport/statusreport_adapter.go
```diff
@@ -298,3 +311,16 @@ func (a *Adapter) getComponentSourceFromSnapshotComponent(snapshot *applicationa
 	}
 	return nil, fmt.Errorf("couldn't find the requested component source info in the given Snapshot")
 }
+
+// determineIfAllIntegrationTestsNotTriggered returns name of integrationTestScenario that is not triggered yet.
+func (a *Adapter) determineIfAllIntegrationTestsNotTriggered(integrationTestScenarios *[]v1beta1.IntegrationTestScenario, testStatuses *intgteststat.SnapshotIntegrationTestStatuses) string {
```
dirgim, line 316: nit: we might consider a different name for this function for more clarity, something like `findMissingIntegrationTestFromStatus` or similar. Not a hard requirement from me though.

kasemAlem: how about `findUntriggeredIntegrationTestFromStatus` ?

dirgim: Sure, that looks good to me too.

controllers/statusreport/statusreport_adapter.go
```diff
@@ -298,3 +311,16 @@ func (a *Adapter) getComponentSourceFromSnapshotComponent(snapshot *applicationa
 	}
 	return nil, fmt.Errorf("couldn't find the requested component source info in the given Snapshot")
 }
+
+// findUntriggeredIntegrationTestFromStatus returns name of integrationTestScenario that is not triggered yet.
+func (a *Adapter) findUntriggeredIntegrationTestFromStatus(integrationTestScenarios *[]v1beta1.IntegrationTestScenario, testStatuses *intgteststat.SnapshotIntegrationTestStatuses) string {
```
MartinBasti, line 316: Also this function should have unittest

controllers/statusreport/statusreport_adapter.go
dirgim: Thanks!

controllers/statusreport/statusreport_adapter_test.go
```diff
@@ -350,6 +350,63 @@ var _ = Describe("Snapshot Adapter", Ordered, func() {
...
+					Environment: v1beta1.TestEnvironment{
+						Name: "envname",
+						Type: "POC",
+						Configuration: &applicationapiv1alpha1.EnvironmentConfiguration{
+							Env: []applicationapiv1alpha1.EnvVarPair{},
+						},
+					},
+				},
+			}
+			Expect(k8sClient.Create(ctx, integrationTestScenarioTest)).Should(Succeed())
```
dirgim, line 392: We probably don't need to create this scenario on the control plane since it gets added to the list anyways (current [line 403](https://github.com/redhat-appstudio/integration-service/pull/443/files#diff-7b702c5210316ec43c95daf87f77f35c26270e1f634b5e1afe66d012d86550dbR403)).
If I'm missing something and we do need to create it, then it also needs to be deleted at the end so it doesn't get left behind.

dirgim: Thanks!

Josh-Everett (approved):

MartinBasti (approved):

PR #455, opened 2023-12-01
fix(STONEINTG-692): ignore restored snapshots

MartinBasti:
We have to ignore restored snaphots because currently it's not restoring status fields and we may redeploy, retest and release old snapshots after restoring resources from backup.


controllers/statusreport/statusreport_controller.go
```diff
@@ -101,7 +108,10 @@ func SetupController(manager ctrl.Manager, log *logr.Logger) error {
 func setupControllerWithManager(manager ctrl.Manager, controller *Reconciler) error {
 	return ctrl.NewControllerManagedBy(manager).
 		For(&applicationapiv1alpha1.Snapshot{}).
-		WithEventFilter(predicate.Or(
-			gitops.SnapshotTestAnnotationChangePredicate())).
+		WithEventFilter(
+			predicate.And(
+				toolkitpredicates.IgnoreBackups{},
```
MartinBasti, line 113: Technically we don't need this predicate here as we don't react to creation in other predicates. But better safe than sorry, LMKWDYT

controllers/statusreport/statusreport_controller_test.go
```diff
@@ -17,8 +17,9 @@ limitations under the License.
 package statusreport
 
 import (
-	"k8s.io/apimachinery/pkg/api/errors"
```
MartinBasti, line 20: blame IDE, I cannot get rid of it, but is fixed it, stdlib imports first

controllers/statusreport/statusreport_controller.go
dirgim: I agree, it's better to have it here in case something changes in the future.

dirgim (approved):
LGTM

PR #489, opened 2023-12-21
fix(stoneintg-730): Add build PLR annotation for snapshot status

14rcole:
Add an annotation to the build pipelineRuns that shows the current status of the snapshot associated (or that will be associated) with that build pipeline.  This will allow users to more easily find out if build pipelines are not being finalized because the snapshot quota has been reached

### DIFF
--- a/controllers/buildpipeline/buildpipeline_adapter.go
+++ b/controllers/buildpipeline/buildpipeline_adapter.go
@@ -18,6 +18,7 @@ package buildpipeline
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"strconv"
 	"strings"
@@ -64,6 +65,13 @@ func NewAdapter(pipelineRun *tektonv1.PipelineRun, component *applicationapiv1al
 // EnsureSnapshotExists is an operation that will ensure that a pipeline Snapshot associated
 // to the build PipelineRun being processed exists. Otherwise, it will create a new pipeline Snapshot.
 func (a *Adapter) EnsureSnapshotExists() (controller.OperationResult, error) {
+	var err error
+	defer func() {
+		err = a.annotateBuildPipelineRunWithCreateSnapshotAnnotation(err)
+		if err != nil {
+			a.logger.Error(err, "Could not add create snapshot annotation to build pipelineRun", h.CreateSnapshotAnnotationName, a.pipelineRun)
+		}
+	}()
 	if !h.HasPipelineRunSucceeded(a.pipelineRun) {
 		if h.HasPipelineRunFinished(a.pipelineRun) || a.pipelineRun.GetDeletionTimestamp() != nil {
 			// The pipeline run  has failed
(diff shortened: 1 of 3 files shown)


controllers/buildpipeline/buildpipeline_adapter.go
```diff
@@ -277,20 +281,45 @@ func (a *Adapter) getSucceededBuildPipelineRunsForComponent(component *applicati
...
+func (a *Adapter) annotateBuildPipelineRunWithSnapshot(pipelineRun *tektonv1.PipelineRun, snapshot *applicationapiv1alpha1.Snapshot) (*tektonv1.PipelineRun, error) {
+	pipelineRun, err := a.annotateBuildPipelineRun(pipelineRun, tekton.SnapshotNameLabel, snapshot.Name)
+	if err != nil {
+		a.logger.LogAuditEvent("Updated build pipelineRun", pipelineRun, h.LogActionUpdate,
+			"snapshot.Name", snapshot.Name)
+	}
+	return pipelineRun, err
+}
+
+func (a *Adapter) annotateBuildPipelineRunWithCreateSnapshotAnnotation(status string, err error) {
```
MartinBasti, line 305: Why  we don't return error state here? We can at least log it in defer func when something fails

helpers/build.go
```diff
@@ -0,0 +1,3 @@
+package helpers
+
+const CreateSnapshotAnnotationName = "test.appstudio.openshift.io/succeed-create-snapshot"
```
MartinBasti, line 3:
```suggestion
const CreateSnapshotAnnotationName = "test.appstudio.openshift.io/create-snapshot-status"
```

`suceeed` sounds confusing given it can also report error

controllers/buildpipeline/buildpipeline_adapter.go
14rcole: I didn't return an error because I thought it would be thrown away in defer.  Logging is a good point, I'll update that.

helpers/build.go
dirgim: If I'm not wrong, I'm still seeing `test.appstudio.openshift.io/succeed-create-snapshot` on this line

controllers/buildpipeline/buildpipeline_adapter.go
```diff
@@ -64,6 +65,9 @@ func NewAdapter(pipelineRun *tektonv1.PipelineRun, component *applicationapiv1al
 // EnsureSnapshotExists is an operation that will ensure that a pipeline Snapshot associated
 // to the build PipelineRun being processed exists. Otherwise, it will create a new pipeline Snapshot.
 func (a *Adapter) EnsureSnapshotExists() (controller.OperationResult, error) {
+	var result string
```
dirgim, line 68: If I'm not missing something, I can't see where this value is set in the adapter. Could this variable (and the `status` parameter itself) be dropped and the `status` be set directly within the `annotateBuildPipelineRunWithCreateSnapshotAnnotation` function since that's where we figure out if something went wrong with Snapshot creation?

14rcole: Yep, done

dirgim (approved):
LGTM

MartinBasti (approved):

kasemAlem (approved):

jsztuka (approved):
lgtm

controllers/buildpipeline/buildpipeline_adapter.go
```diff
@@ -277,20 +285,49 @@ func (a *Adapter) getSucceededBuildPipelineRunsForComponent(component *applicati
...
 	}
-	a.logger.LogAuditEvent("Updated build pipelineRun", pipelineRun, h.LogActionUpdate,
-		"snapshot.Name", snapshot.Name)
 	return pipelineRun, nil
 }
 
+func (a *Adapter) annotateBuildPipelineRunWithSnapshot(pipelineRun *tektonv1.PipelineRun, snapshot *applicationapiv1alpha1.Snapshot) (*tektonv1.PipelineRun, error) {
+	pipelineRun, err := a.annotateBuildPipelineRun(pipelineRun, tekton.SnapshotNameLabel, snapshot.Name)
+	if err != nil {
+		a.logger.LogAuditEvent("Updated build pipelineRun", pipelineRun, h.LogActionUpdate,
```
hongweiliu17, line 303: When `err != nil`, it is a failure, we should not log this successful message.

hongweiliu17 (changes requested):
Only one comment for log message when `err != nil`, others look good to me.

hongweiliu17 (approved):
LGTM.

14rcole:
Note: The EC tests keep failing and the PLRs are getting cleaned up while I'm debugging.  I'm going to re-test tomorrow and add finalizers to all of the PLRs created for the tests.  Will also be sure to clean them up before closing the ticket.

hongweiliu17:
> Note: The EC tests keep failing and the PLRs are getting cleaned up while I'm debugging. I'm going to re-test tomorrow and add finalizers to all of the PLRs created for the tests. Will also be sure to clean them up before closing the ticket.

I think the failure is caused by [RHTAPBUGS-1099](https://issues.redhat.com/browse/RHTAPBUGS-1099)

sonam1412 (approved):

PR #495, opened 2023-12-29
feat(STONEINTG-703): reset snapshot status conditions when rerunning test

hongweiliu17:

### DIFF
--- a/controllers/snapshot/snapshot_adapter.go
+++ b/controllers/snapshot/snapshot_adapter.go
@@ -156,6 +156,11 @@ func (a *Adapter) EnsureRerunPipelineRunsExist() (controller.OperationResult, er
 		return controller.RequeueWithError(err)
 	}
 
+	if err = gitops.ResetSnapshotStatusConditions(a.client, a.context, a.snapshot, "Integration test is being rerun for snapshot"); err != nil {
+		a.logger.Error(err, "Failed to reset snapshot status conditions")
+		return controller.RequeueWithError(err)
+	}
+
 	if err = gitops.RemoveIntegrationTestRerunLabel(a.client, a.context, a.snapshot); err != nil {
 		return controller.RequeueWithError(err)
 	}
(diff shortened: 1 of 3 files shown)


controllers/snapshot/snapshot_adapter.go
```diff
@@ -156,6 +156,11 @@ func (a *Adapter) EnsureRerunPipelineRunsExist() (controller.OperationResult, er
 		return controller.RequeueWithError(err)
 	}
 
+	if err = gitops.ResetSnapshotStatusConditions(a.client, a.context, a.snapshot, "Snapshot is being rerun"); err != nil {
```
dheerajodha, line 159:
```suggestion
	if err = gitops.ResetSnapshotStatusConditions(a.client, a.context, a.snapshot, fmt.Sprintf("Snapshot is being retested by the IntegrationTestScenario: %s", integrationTestScenario.Name)); err != nil {
```
I'm thinking about providing the user with more details about which scenario is currently retesting the snapshot. But then if the user re-runs another scenario simultaneously, we'd need to add this new scenario's name in this message as well -- it would make things complicated, wouldn't it?

On second thought, the users re-ran the scenario by themselves, so they know which scenario(s) is retesting their snapshot, so it's not a necessity to add the scenario's name here. Wdyt?

hongweiliu17: > I'm thinking about providing the user with more details about which scenario is currently retesting the snapshot. But then if the user re-runs another scenario simultaneously, we'd need to add this new scenario's name in this message as well -- it would make things complicated, wouldn't it?

It is not a bad idea to provide this scenario name to status condition. And yes, if user rerun another test scenario, providing scenario name in status condition will be some complicated.
> 
> On second thought, the users re-ran the scenario by themselves, so they know which scenario(s) is retesting their snapshot, so it's not a necessity to add the scenario's name here. Wdyt?

Yes, user should have known the scenario name.

But let me add more info to message to make it clearer.

dheerajodha (approved):
LGTM

dirgim (approved):
LGTM

kasemAlem (approved):

Josh-Everett (approved):

PR #501, opened 2024-01-04
feat: add additional buckets to sli 2b histogram

14rcole:
I currently maintain a spreadsheet tracking the latency between snapshot creation time and PLR start time for scenarios that use ephemeral environments.  So far, five snapshots have taken 15-30 seconds and one has taken 30+ seconds.  This commit adds two new buckets: 30-45 and 45-60.  This will allow us to get a better idea of just how long ephemeral environments are taking to be created so that we can more accurately create an SLO for this SLI.

### DIFF
--- a/metrics/integration.go
+++ b/metrics/integration.go
@@ -45,7 +45,7 @@ var (
 		prometheus.HistogramOpts{
 			Name:    "integration_svc_snapshot_created_to_pipelinerun_with_ephemeral_env_started_seconds",
 			Help:    "The duration measures the time elapsed between the creation of a snapshot resource and the initiation of an integration PipelineRun for pipelines operating within an ephemeral environment.",
-			Buckets: []float64{0.05, 0.1, 0.5, 1, 2, 3, 4, 5, 10, 15, 30},
+			Buckets: []float64{1, 5, 10, 15, 20, 25, 30, 35, 45, 60, 120, 240, 300},
 		},
 	)
 


metrics/integration.go
```diff
@@ -37,7 +37,7 @@ var (
 		prometheus.HistogramOpts{
 			Name:    "integration_svc_snapshot_created_to_pipelinerun_with_static_env_started_seconds",
 			Help:    "Time duration from the moment the snapshot resource was created till a integration pipelineRun is started in a static environment",
-			Buckets: []float64{0.05, 0.1, 0.5, 1, 2, 3, 4, 5, 10, 15, 30},
+			Buckets: []float64{0.05, 0.1, 0.5, 1, 2, 3, 4, 5, 10, 15, 30, 45, 60},
```
MartinBasti, line 40: This is SLI 2A; 2B is for ephemeral envs

Also for ephemeral envs, we should include numbers up to limit  https://github.com/redhat-appstudio/integration-service/blob/c199d2673bc5c7dc9936798acce1ffd9bad6864b/controllers/binding/snapshotenvironmentbinding_adapter.go#L43 or generaly make gaps between numbers bigger, I don't think that we can cound with deploments under 1s at all

MartinBasti: Line 48 should be updated in this PR

14rcole: Ack, I'll also remove the first few buckets

metrics/integration.go
```diff
@@ -45,7 +45,7 @@ var (
 		prometheus.HistogramOpts{
 			Name:    "integration_svc_snapshot_created_to_pipelinerun_with_ephemeral_env_started_seconds",
 			Help:    "The duration measures the time elapsed between the creation of a snapshot resource and the initiation of an integration PipelineRun for pipelines operating within an ephemeral environment.",
-			Buckets: []float64{0.05, 0.1, 0.5, 1, 2, 3, 4, 5, 10, 15, 30},
+			Buckets: []float64{1, 3, 5, 10, 15, 20, 25, 30, 35, 45, 60},
```
MartinBasti, line 48: I still think we should include buckets up to 300s limit
```suggestion
			Buckets: []float64{1, 5, 10, 15, 30, 45, 60, 120, 240, 300},
```

14rcole: We can do that, but I want to keep in the 20, 25, and 35s buckets.  Most of the ephemeral environments have taken 15-30s so far and I'd like to know a bit more about what's going on in there.  For example, it's possible that we had 5 provisionings take 29s and 1 take 31s but its also possible we had 5 take 16s and 1 take 100s. There's no reason we need to have 11 buckets, right?

MartinBasti: yeah, fine by me, I just wanted to extend it to full period

MartinBasti (approved):

dirgim (approved):
LGTM

jsztuka (approved):

PR #502, opened 2024-01-05
fix(STONEINTG-737): cleanup related resources after ITS is deleted

dirgim:
* If the integrationTestScenario has related ephemeral environments, ensure that those are cleaned up when it is deleted
* Add a finalizer to IntegrationTestScenarios in order to be able to correctly handle the resource cleanup
* Add utility functions to handle the above functionalities
* Update diagrams with the new behavior


controllers/scenario/predicates.go
```diff
@@ -21,15 +21,15 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
 )
 
-// IntegrationPipelineRunStartedPredicate returns a predicate which filters out
-// only created integration scenarios
+// IntegrationScenarioCreatedPredicate returns a predicate which filters out
```
jsztuka, line 24: I feel that the functionname is little bit misleading, can we consider something like: 
_IntegrationScenarioCreatedAndDeletedPredicate_ ?

dirgim: Tbh I was thinking of removing the predicate altogether as we're responding to basically all types of events now.

dirgim: Ack, updated

MartinBasti: Do we really want to react to generic events?

helpers/integration.go
```diff
@@ -478,3 +479,35 @@ func AddFinalizerToPipelineRun(adapterClient client.Client, logger IntegrationLo
...
 }
+
+// RemoveFinalizerFromScenario removes the finalizer from the IntegrationTestScenario.
+// If finalizer was not removed successfully, a non-nil error is returned.
+func RemoveFinalizerFromScenario(adapterClient client.Client, logger IntegrationLogger, ctx context.Context, scenario *v1beta1.IntegrationTestScenario, finalizer string) error {
+	patch := client.MergeFrom(scenario.DeepCopy())
+	if ok := controllerutil.RemoveFinalizer(scenario, finalizer); ok {
+		err := adapterClient.Patch(ctx, scenario, patch)
+		if err != nil {
+			return fmt.Errorf("error occurred while patching the updated inte after finalizer removal: %w", err)
```
hongweiliu17, line 490:
```suggestion
			return fmt.Errorf("error occurred while patching the updated IntegrationTestScenario after finalizer removal: %w", err)
```

hongweiliu17 (approved):
Only one comment for log message. Others looks to me.

controllers/scenario/predicates.go
dirgim: I'm fine with reacting to all events for this owned resource and making sure it's validated tbh

helpers/integration.go
dirgim: Good catch, thanks. Fixed it now.

MartinBasti (approved):

PR #513, opened 2024-01-11
feat(STONE-738): Copy labels+annotations to composite snapshot

jsztuka:

### DIFF
--- a/gitops/snapshot.go
+++ b/gitops/snapshot.go
@@ -775,3 +775,33 @@ func ResetSnapshotStatusConditions(adapterClient client.Client, ctx context.Cont
 
 	return nil
 }
+
+// CopySnapshotLabelsAndAnnotation coppies labels and annotations from build pipelineRun or tested snapshot
+// into regular or composite snapshot
+func CopySnapshotLabelsAndAnnotation(application *applicationapiv1alpha1.Application, snapshot *applicationapiv1alpha1.Snapshot, componentName string, source *metav1.ObjectMeta, prefix string, isComposite bool) {
+
+	if snapshot.Labels == nil {
+		snapshot.Labels = map[string]string{}
+	}
+
+	if snapshot.Annotations == nil {
+		snapshot.Annotations = map[string]string{}
+	}
+	if !isComposite {
+		snapshot.Labels[SnapshotTypeLabel] = SnapshotComponentType
+	} else {
+		snapshot.Labels[SnapshotTypeLabel] = SnapshotCompositeType
+	}
+
+	snapshot.Labels[SnapshotComponentLabel] = componentName
+	snapshot.Labels[ApplicationNameLabel] = application.Name
+
+	// Copy PAC annotations/labels from source(tested snapshot or pipelinerun) to snapshot.
+	_ = metadata.CopyLabelsWithPrefixReplacement(source, &snapshot.ObjectMeta, "pipelinesascode.tekton.dev", PipelinesAsCodePrefix)
+	_ = metadata.CopyAnnotationsWithPrefixReplacement(source, &snapshot.ObjectMeta, "pipelinesascode.tekton.dev", PipelinesAsCodePrefix)
+
+	// Copy labels and annotations prefixed with defined prefix
+	_ = metadata.CopyLabelsByPrefix(source, &snapshot.ObjectMeta, prefix)
+	_ = metadata.CopyAnnotationsByPrefix(source, &snapshot.ObjectMeta, prefix)
+
+}
--- a/controllers/buildpipeline/buildpipeline_adapter.go
+++ b/controllers/buildpipeline/buildpipeline_adapter.go
@@ -206,27 +206,14 @@ func (a *Adapter) prepareSnapshotForPipelineRun(pipelineRun *tektonv1.PipelineRu
 		return nil, err
 	}
 
-	if snapshot.Labels == nil {
-		snapshot.Labels = make(map[string]string)
-	}
-	snapshot.Labels[gitops.SnapshotTypeLabel] = gitops.SnapshotComponentType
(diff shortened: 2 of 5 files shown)


jsztuka:
Do we want to revert changes from go.mod & go.sum?

gitops/snapshot.go
```diff
@@ -775,3 +775,23 @@ func ResetSnapshotStatusConditions(adapterClient client.Client, ctx context.Cont
 
 	return nil
 }
+
+func CopySnapshotLabelsAndAnnotation(application *applicationapiv1alpha1.Application, snapshot *applicationapiv1alpha1.Snapshot, componentName string, source metav1.ObjectMeta, destination metav1.ObjectMeta, prefix string) (*applicationapiv1alpha1.Snapshot, error) {
```
MartinBasti, line 779: missing docstring

you are passing snapshot as pointer, you don't need to return the same pointer

gitops/snapshot.go
```diff
@@ -775,3 +775,23 @@ func ResetSnapshotStatusConditions(adapterClient client.Client, ctx context.Cont
...
+
+	if snapshot.Labels == nil {
+		snapshot.Labels = make(map[string]string)
```
MartinBasti, line 782: Wouldn't just this work instead of make() ?
```suggestion
		snapshot.Labels = map[string]string{}
```

gitops/snapshot.go
```diff
@@ -775,3 +775,23 @@ func ResetSnapshotStatusConditions(adapterClient client.Client, ctx context.Cont
 
 	return nil
 }
+
+func CopySnapshotLabelsAndAnnotation(application *applicationapiv1alpha1.Application, snapshot *applicationapiv1alpha1.Snapshot, componentName string, source metav1.ObjectMeta, destination metav1.ObjectMeta, prefix string) (*applicationapiv1alpha1.Snapshot, error) {
+
+	if snapshot.Labels == nil {
```
MartinBasti, line 781: Shouldn't be annotations handled in the same way?

gitops/snapshot.go
```diff
@@ -775,3 +775,23 @@ func ResetSnapshotStatusConditions(adapterClient client.Client, ctx context.Cont
...
+
+	// Copy PAC annotations/labels from source(tested snapshot or pipelinerun) to snapshot.
+	_ = metadata.CopyLabelsWithPrefixReplacement(&source, &destination, "pipelinesascode.tekton.dev", PipelinesAsCodePrefix)
+	_ = metadata.CopyAnnotationsWithPrefixReplacement(&source, &destination, "pipelinesascode.tekton.dev", PipelinesAsCodePrefix)
+
+	// Copy labels and annotations prefixed with defined prefix
+	_ = metadata.CopyLabelsByPrefix(&source, &destination, prefix)
+	_ = metadata.CopyAnnotationsByPrefix(&source, &destination, prefix)
+
+	return snapshot, nil
```
MartinBasti, line 796: Nothing returns error in this function, does it even make sense to declare error in return types?

MartinBasti (changes requested):

gitops/snapshot.go
jsztuka: I agree, added case for annotations also.

gitops/snapshot.go
jsztuka: Added docstring, removed returned statements.

gitops/snapshot.go
jsztuka: I think so, updated.

gitops/snapshot.go
jsztuka: No need for return types there, removed.

gitops/snapshot.go
```diff
@@ -775,3 +775,33 @@ func ResetSnapshotStatusConditions(adapterClient client.Client, ctx context.Cont
 
 	return nil
 }
+
+// CopySnapshotLabelsAndAnnotation coppies labels and annotations from build pipelineRun or tested snapshot
+// into regular or composite snapshot
+func CopySnapshotLabelsAndAnnotation(application *applicationapiv1alpha1.Application, snapshot *applicationapiv1alpha1.Snapshot, componentName string, source *metav1.ObjectMeta, destination *metav1.ObjectMeta, prefix string, isComposite bool) {
```
MartinBasti, line 781: but `destination` is clear here, it's always `snapshot` why do we need to list it explicitly as a parameter?

gitops/snapshot.go
```diff
@@ -775,3 +775,33 @@ func ResetSnapshotStatusConditions(adapterClient client.Client, ctx context.Cont
 
 	return nil
 }
+
+// CopySnapshotLabelsAndAnnotation coppies labels and annotations from build pipelineRun or tested snapshot
+// into regular or composite snapshot
+func CopySnapshotLabelsAndAnnotation(application *applicationapiv1alpha1.Application, snapshot *applicationapiv1alpha1.Snapshot, componentName string, source *metav1.ObjectMeta, prefix string, isComposite bool) {
```
dirgim, line 781: I think we'll need [unit test coverage](https://app.codecov.io/gh/redhat-appstudio/integration-service/pull/513) for this new function.

jsztuka: Added unit tests for covering newly implemented function.

dirgim: Now the [ginkgo linter is complaining](https://github.com/redhat-appstudio/integration-service/actions/runs/7528084698/job/20489547180?pr=513) - it's about the `Expect(err).To(BeNil())` calls that should be changed to `Expect(err).ToNot(HaveOccurred())`.

MartinBasti (approved):

Josh-Everett (approved):

PR #514, opened 2024-01-11
fix(RHTAPBUGS-1091): react to deleted running build PLR

MartinBasti:
Tekton keeps deleting PLRs in running status. Integration-service must react to deletion event, even if PLR is still in running status to prevent blocking of deletion by removing finalizer.

### DIFF
--- a/controllers/buildpipeline/buildpipeline_adapter.go
+++ b/controllers/buildpipeline/buildpipeline_adapter.go
@@ -65,8 +65,9 @@ func NewAdapter(pipelineRun *tektonv1.PipelineRun, component *applicationapiv1al
 // to the build PipelineRun being processed exists. Otherwise, it will create a new pipeline Snapshot.
 func (a *Adapter) EnsureSnapshotExists() (controller.OperationResult, error) {
 	if !h.HasPipelineRunSucceeded(a.pipelineRun) {
-		if h.HasPipelineRunFinished(a.pipelineRun) {
+		if h.HasPipelineRunFinished(a.pipelineRun) || a.pipelineRun.GetDeletionTimestamp() != nil {
 			// The pipeline run  has failed
+			// OR pipeline has been deleted but it's still in running state (tekton bug/feature?)
 			return a.removeFinalizerAndContinueProcessing()
 		}
 		// The build pipeline run has not finished yet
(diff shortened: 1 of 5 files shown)


Josh-Everett (approved):
lgtm

dheerajodha (approved):
LGTM

nit: would be nice if you could update the commit title and body by replacing "PLRs" with "build PLRs" to reflect that this is specific to Build PLRs

MartinBasti:
> LGTM
> 
> nit: would be nice if you could update the commit title and body by replacing "PLRs" with "build PLRs" to reflect that this is specific to Build PLRs

updated

dheerajodha:
Should work this time

/test integration-service-e2e

controllers/buildpipeline/buildpipeline_adapter.go
```diff
@@ -65,8 +65,9 @@ func NewAdapter(pipelineRun *tektonv1.PipelineRun, component *applicationapiv1al
 // to the build PipelineRun being processed exists. Otherwise, it will create a new pipeline Snapshot.
 func (a *Adapter) EnsureSnapshotExists() (controller.OperationResult, error) {
 	if !h.HasPipelineRunSucceeded(a.pipelineRun) {
-		if h.HasPipelineRunFinished(a.pipelineRun) {
+		if h.HasPipelineRunFinished(a.pipelineRun) || a.pipelineRun.GetDeletionTimestamp() != nil {
```
hongweiliu17, line 68: If one PLR works well but has not finished, is it possible the prune mechanism tries to delete it? If yes, an in-progess PLR might be deleted by this code as well. I mean when does the prune mechanism delete the PLR? Only after the PLR finishes?

MartinBasti: It's the purpose of this PR to delete running PLR marked for deletion. If PLR is deleted tekton doesn't change status to non-running, it will be stuck forever in running status. Se we have to "unlock" even running pipeline.

dheerajodha: makes me think about the purpose of having a finalizer on build PLRs in the first place, it was to ensure that the integration service can create snapshots of successful builds. By this change, we'd allow the pruner to delete the PLR (while it's running) whenever it wants.

Now with this change, the finalizer would only be useful when the pruner tries to delete a successful build PLR for which the integration service hasn't created a Snapshot yet. IIUC

hongweiliu17: If the time between the PLR is started and decided to be deleted is long enough, the change can make sure the build PLR can be deleted correctly.

hongweiliu17 (approved):
LGTM

PR #517, opened 2024-01-15
refactor(MarkSnapshot*): don't return the same pointer

MartinBasti:
MarkSnapshot* functiond were returning the same pointer as they received in snapshot paramter. There is no need to return pointer if we pass pointer as paramereter.

Given the pointer was returned it could puzzle devs to expect that it's a newly allocated memory instead. This old-new pointer was even used inconsistently in code, it's just safer to remove it and make clear that original pointer is being used for data updates.


dirgim (approved):
LGTM

hongweiliu17 (approved):
LGTM

PR #523, opened 2024-01-18
fix(stoneintg-753): requeue release creation until timeout

14rcole:
Prior to this commit the integration service would requeue release creation indefinitely if it failed.  If a release quota was hit then releases would be backed up then be created all at once and overwhelm user workspaces with potentially no-longer-relevant releases.  This change sets a timeout of three hours for requeued releases, after which creating a release for that snapshot will fail.

### DIFF
--- a/controllers/snapshot/snapshot_adapter.go
+++ b/controllers/snapshot/snapshot_adapter.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"reflect"
 	"strings"
+	"time"
 
 	clienterrors "k8s.io/apimachinery/pkg/api/errors"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -42,6 +43,8 @@ import (
 	tektonv1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
+
+const RetryReleaseTimeout = time.Duration(3 * time.Hour)
 
 // configuration options for scenario
 type ScenarioOptions struct {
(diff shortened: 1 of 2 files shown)


MartinBasti (changes requested):
LGTM except this small issue with not using constants

controllers/snapshot/snapshot_adapter.go
```diff
@@ -934,3 +935,14 @@ func GetDetailsFromStatusError(err error) (string, string) {
 	}
 	return errKind, errName
 }
+
```
jsztuka, line 938: Do we want to add docstring here?

14rcole: Yes, thank you

controllers/snapshot/snapshot_adapter.go
```diff
@@ -934,3 +935,14 @@ func GetDetailsFromStatusError(err error) (string, string) {
...
+func (a *Adapter) RequeueIfYoungerThanThreshold(retErr error) (controller.OperationResult, error) {
+	timeout := time.Duration(3 * time.Hour)
```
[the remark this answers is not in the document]
MartinBasti: sorry I mean it as a module constant, not just const. Something that could be exported into unittest

controllers/snapshot/snapshot_adapter.go
```diff
@@ -435,9 +436,9 @@ func (a *Adapter) EnsureAllReleasesExist() (controller.OperationResult, error) {
 		if er != nil {
 			a.logger.Error(er, "Failed to mark snapshot integration status as invalid",
 				"snapshot.Name", a.snapshot.Name)
-			return controller.RequeueWithError(errors.Join(err, er))
+			return a.RequeueIfYoungerThanThreshold(errors.Join(err, er))
```
14rcole, line 439: It seems like this line and the other three like it are the cause of the "Duplicated Lines" error in Sonarcloud.  Is there any way to get around this? It doesn't look like I can combine the lines in any way.

MartinBasti: I'd just ignore it

controllers/snapshot/snapshot_adapter.go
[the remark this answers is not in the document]
14rcole: Got it, thanks for the clarification

MartinBasti (approved):

Josh-Everett (approved):

PR #534, opened 2024-01-23
chore(RHTAPWATCH-729): get namespace release snapshots

yftacherzog:
Find all snapshots that are part of a release

### DIFF
--- a/cmd/snapshotgc/snapshotgc.go
+++ b/cmd/snapshotgc/snapshotgc.go
@@ -1,0 +1,59 @@
+package main
+
+import (
+	"context"
+
+	"github.com/go-logr/logr"
+	applicationapiv1alpha1 "github.com/redhat-appstudio/application-api/api/v1alpha1"
+	releasev1alpha1 "github.com/redhat-appstudio/release-service/api/v1alpha1"
+	"k8s.io/apimachinery/pkg/runtime"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var (
+	scheme = runtime.NewScheme()
+)
+
+func init() {
+	utilruntime.Must(applicationapiv1alpha1.AddToScheme(scheme))
+	utilruntime.Must(clientgoscheme.AddToScheme(scheme))
+	utilruntime.Must(releasev1alpha1.AddToScheme(scheme))
+}
+
+// Stores pointers to resources to which the snapshot is associated
+type snapshotData struct {
+	// environmentBinding applicationapiv1alpha1.SnapshotEnvironmentBinding
+	release releasev1alpha1.Release
+}
+
+// Gets a map to allow to tell with direct lookup if a snapshot is associated with
+// a release resource
+func getSnapshotsForNSReleases(
+	cl client.Client,
+	snapToData map[string]snapshotData,
+	namespace string,
+	logger logr.Logger,
+) (map[string]snapshotData, error) {
+	releases := &releasev1alpha1.ReleaseList{}
+	err := cl.List(
+		context.Background(),
+		releases,
+		&client.ListOptions{Namespace: namespace},
+	)
+	if err != nil {
+		logger.Error(err, "Failed to list releases")
+		return nil, err
+	}
+
+	for _, release := range releases.Items {
+		data, ok := snapToData[release.Spec.Snapshot]
+		if !ok {
+			data = snapshotData{}
+		}
+		data.release = release
+		snapToData[release.Spec.Snapshot] = data
+	}
+	return snapToData, nil
+}
--- a/cmd/snapshotgc/snapshotgc_test.go
+++ b/cmd/snapshotgc/snapshotgc_test.go
@@ -1,0 +1,123 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/go-logr/logr"
+	"github.com/go-logr/zapr"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	releasev1alpha1 "github.com/redhat-appstudio/release-service/api/v1alpha1"
+	"go.uber.org/zap"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+var _ = Describe("Test garbage collection for snapshots", func() {
+	var logger logr.Logger
+
+	BeforeEach(func() {
+		zapLog, err := zap.NewDevelopment()
+		if err != nil {
+			panic(fmt.Sprintf("who watches the watchmen (%v)?", err))
+		}
+		logger = zapr.NewLogger(zapLog)
+	})
+
+	Describe("Test getSnapshotsForNSReleases", func() {
+		It("Finds no snapshot when there are no releases", func() {
+
+			cl := fake.NewClientBuilder().
+				WithScheme(scheme).
+				WithLists(
+					&releasev1alpha1.ReleaseList{
+						Items: []releasev1alpha1.Release{},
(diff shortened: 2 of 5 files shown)


cmd/snapshotgc/snapshotgc.go
```diff
@@ -0,0 +1,61 @@
...
+type snapshotData struct {
+	snapshot applicationapiv1alpha1.Snapshot
+	// environmentBinding applicationapiv1alpha1.SnapshotEnvironmentBinding
+	release releasev1alpha1.Release
+}
+
+func getSnapshotsForReleases(
+	cl client.Client,
+	snapToData map[string]snapshotData,
+) map[string]snapshotData {
```
MartinBasti, line 35: `map[string]snapshotData`

I'd like to know little bit more about this structure, why it's uses `namespace|name` as a key, will it be used by future function for lookups?

cmd/snapshotgc/snapshotgc.go
```diff
@@ -0,0 +1,61 @@
...
+	releases := &releasev1alpha1.ReleaseList{}
+	err := cl.List(context.Background(), releases)
```
MartinBasti, line 37: I don't like that we are fetching all snapshots at once. It generates a big load on api and also we have to store a lot of results in memory.

Why we don't go namespace by namespace, one by one, process one namespace, free the memory and continue with another. It will also simplify code it won't need to keep complex key in map

cmd/snapshotgc/snapshotgc.go
```diff
@@ -0,0 +1,61 @@
...
+	}
+
+	for _, release := range releases.Items {
+		snap := &applicationapiv1alpha1.Snapshot{}
+		err := cl.Get(context.Background(), types.NamespacedName{
+			Name:      release.Spec.Snapshot,
+			Namespace: release.Namespace,
+		}, snap)
+		if err != nil {
+			fmt.Printf("Failed Getting snapshot: %v\n", err)
```
MartinBasti, line 50: do we miss `continue` here?

cmd/snapshotgc/snapshotgc.go
```diff
@@ -0,0 +1,61 @@
...
+		}
+
+		data, ok := snapToData[snap.Namespace+"|"+snap.Name]
```
MartinBasti, line 53: This is to handle the same snapshot released in multiple releases?

cmd/snapshotgc/snapshotgc.go
```diff
@@ -0,0 +1,61 @@
...
+	releasev1alpha1.AddToScheme(scheme)
+}
+
+type snapshotData struct {
+	snapshot applicationapiv1alpha1.Snapshot
+	// environmentBinding applicationapiv1alpha1.SnapshotEnvironmentBinding
+	release releasev1alpha1.Release
+}
+
+func getSnapshotsForReleases(
```
MartinBasti, line 32: types and functions are missing docstrings

MartinBasti (changes requested):

cmd/snapshotgc/snapshotgc.go
yftacherzog: > why it's uses namespace|name as a key
I was hoping to use the snapshot itself as key, but I was not able to make it work (not comparable or something like that), so I opted for a string, and this combination is something I expect to be unique.
I'm open for other suggestions.

> will it be used by future function for lookups?
Yes.

cmd/snapshotgc/snapshotgc.go
yftacherzog: I'll look into that. I was considering that initially, but had something in mind that made me prefer looking at it as a flat list.

One thing in favor of the current approach is the number of API calls.
Suppose that we go namespace by namespace, it means that we'll have to fetch releases (and bindings later) for each namespace (also namespaces that don't have them). 
Do you expect memory to be more of an issue than the total number of calls?

cmd/snapshotgc/snapshotgc.go
yftacherzog: yes. Fixed.

cmd/snapshotgc/snapshotgc.go
yftacherzog: That snapshot data struct can hold references to both bindings and release. later on, we're going to have another (all too similar) function that will collect bindings. We're checking here if the snapshot was added already for bindings (or something else).

yftacherzog:
@MartinBasti any idea why is it failing on kube-linter? Is it related to this PR?

MartinBasti:
> @MartinBasti any idea why is it failing on kube-linter? Is it related to this PR?

It's not related, just a new check has been added there today, I'll fix in separate PR, feel free to ignore

cmd/snapshotgc/snapshotgc.go
MartinBasti: True is somewhere in the middle, you don;t want many calls, but you also don't want huge calls, where you trigger OOM attack on server by listing the whole world.

But you are afraid of too many api calls, why do you then fetch snapshots one by one instead of getting all snapshots and then process it.

The middle road for me would be:
Per namespace: 
- fetch all snapshots at once
- fetch all releases at once
- fetch all SEBs (bindings) at once
Do magic and find which snapshots should be deleted, and delete them


Integration service in prod is currently using 650MB of memory, where the majority is caused by caches of CRs. 600MB sounds quite a lot for me for pruner.

yftacherzog: ok. I changed the implementation.

- Going namespace by namespace
- Storing only the snapshot name at the first step
- Will fetch the snapshot itself only later in the process, and only for those which are not associated with release/binding.

cmd/snapshotgc/snapshotgc.go
yftacherzog: The logic will be per-namespace, so I changed it to be only the snapshot name.

cmd/snapshotgc/snapshotgc.go
yftacherzog: @MartinBasti let me know if you're ok with the change. I'll apply the same approach for coming PRs.

MartinBasti: Looks good to me, thank you

go.mod
```diff
@@ -38,6 +38,7 @@ require (
 	github.com/cespare/xxhash/v2 v2.2.0 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/emicklei/go-restful/v3 v3.11.0 // indirect
+	github.com/evanphx/json-patch v5.6.0+incompatible // indirect
```
MartinBasti, line 41: Why do we need this change?

yftacherzog: No idea. To be able to run the tests locally, I had to run `go mod tidy` and it added that.
Not sure how to avoid it.

MartinBasti (approved):

MartinBasti (approved):

cmd/snapshotgc/snapshotgc_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+var _ = Describe("Test garbage collection for snapshots", func() {
+	var logger logr.Logger
+
+	BeforeEach(func() {
+		zapLog, err := zap.NewDevelopment()
+		if err != nil {
+			panic(fmt.Sprintf("who watches the watchmen (%v)?", err))
+		}
+		logger = zapr.NewLogger(zapLog)
```
jsztuka, line 24: What is the reason behind using two different loggers? 
Can't we use just one or the another?

yftacherzog: For creating the logger, I followed this: https://github.com/go-logr/zapr?tab=readme-ov-file#zapr-zap

I honestly don't know why it's created this way. 

Per @MartinBasti request, I used this library as it implements the `logr` interface, that is suppose to help us keeping consistency between the script's logs and the controller's logs.

jsztuka (approved):
lgtm

PR #535, opened 2024-01-23
fix(kube-lint): fix liveness-port check

MartinBasti:
Image doesn't have exposed ports for liveness and readiness probes.

Adding EXPOSE into image. Also changing tag of the image to latest as we don't use `next` anymore.

This also creates a chicken egg problem, where we have to build latest image first to make effect on kube-lint, sorry future generations, but this is the best effort/outcome ratio (once we move this check copletely into konflux, we can use currently built image there).

### DIFF
--- a/.github/.kube-linter-config.yaml
+++ b/.github/.kube-linter-config.yaml
@@ -5,4 +5,5 @@ checks:
   include: [ ]
   # exclude explicitly excludes checks, by name. exclude has the highest priority: if a check is
   # in exclude, then it is not considered, even if it is in include as well.
-  exclude: [ ]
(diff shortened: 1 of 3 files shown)


MartinBasti:
kill me, nothing is easy
```
check: latest-tag, remediation: Use a container image with a specific tag other than latest.
```

MartinBasti:
As I mentioned, this must be merged first to update latest image to have effect on kube-linter (or may not work at all and new fix would be needed :D)

kasemAlem (approved):

Josh-Everett (approved):
lgtm

PR #537, opened 2024-01-24
chore(RHTAPWATCH-729): get namespace Bindings snapshots

yftacherzog:
Find all snapshots that are part of a SnapshotEnvironmentBinding

### DIFF
--- a/cmd/snapshotgc/snapshotgc.go
+++ b/cmd/snapshotgc/snapshotgc.go
@@ -24,8 +24,8 @@ func init() {
 
 // Stores pointers to resources to which the snapshot is associated
 type snapshotData struct {
-	// environmentBinding applicationapiv1alpha1.SnapshotEnvironmentBinding
-	release releasev1alpha1.Release
+	environmentBinding applicationapiv1alpha1.SnapshotEnvironmentBinding
+	release            releasev1alpha1.Release
 }
 
 // Gets a map to allow to tell with direct lookup if a snapshot is associated with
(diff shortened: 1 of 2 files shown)


cmd/snapshotgc/snapshotgc.go
```diff
@@ -57,3 +57,33 @@ func getSnapshotsForNSReleases(
...
+
+	for _, bind := range binds.Items {
+		data, ok := snapToData[bind.Spec.Snapshot]
+		if !ok {
+			data = snapshotData{}
+		}
+		data.environmentBinding = bind
+		snapToData[bind.Spec.Snapshot] = data
+	}
+	return snapToData, nil
```
MartinBasti, line 88: Actually we don't have to return `snapToData` it's passed by reference so we don't have to return the same ref as is in param.

MartinBasti (approved):
Looks good, excetp the one comment, but I'm fine with fixing it globally in follow up PR

cmd/snapshotgc/snapshotgc.go
yftacherzog: I don't like the practice of relying on side effects. I prefer returning the result directly. It also makes testing more straightforward.

MartinBasti: What side effects? Reference/pointer parameters are well documented base concept of golang. 
You must hate go-client that uses only references to update objects.

What is no nice with such function is, that because developer received a return value, the one may expect that the original value is unchanged and still work with it as independent copy, but that's not true, it's the same reference.

I don't get how unit-testing is more straightforward.

yftacherzog: > You must hate go-client that uses only references to update objects.

Indeed I do!

> What is no nice with such function is, that because developer received a return value, the one may expect that the original value is unchanged and still work with it as independent copy, but that's not true, it's the same reference.

That is indeed a downside.

> I don't get how unit-testing is more straightforward.

It's much easier to mock something that just returns a value instead of modifying its inputs.

MartinBasti: ah mocking is always fun

as I said, not a merge blocker for me

14rcole (approved):

jsztuka (approved):
lgtm

hongweiliu17 (approved):
LGTM

PR #481, opened 2023-12-18
feat(STONEINTG-725): retry for recoverable error when creating relase

hongweiliu17:
when creating release for snapshot
* retry for recoverable error

Signed-off-by: Hongwei Liu <hongliu@redhat.com>

## Maintainers will complete the following section

- [ ] Commit messages are descriptive enough ([hints](https://www.freecodecamp.org/news/how-to-write-better-git-commit-messages/))
- [ ] Code coverage from testing does not decrease and new code is covered ([check the PR coverage on codecov](https://app.codecov.io/gh/redhat-appstudio/integration-service/pulls))
- [ ] [Controllers diagrams](https://github.com/redhat-appstudio/integration-service/tree/main/docs) are updated when PR changes controllers code  (if applicable)

PR #542, opened 2024-01-26
fix(RHTAPBUGS-1119): don't block build PLR removal

MartinBasti:
When we cannot create a snapshot, we shouldn't block build PLR with finalizers, but rather let users to restart build.

### DIFF
--- a/controllers/buildpipeline/buildpipeline_adapter.go
+++ b/controllers/buildpipeline/buildpipeline_adapter.go
@@ -31,6 +31,7 @@ import (
 	"github.com/redhat-appstudio/integration-service/tekton"
 	"github.com/redhat-appstudio/operator-toolkit/controller"
 	tektonv1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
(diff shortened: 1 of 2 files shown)


kasemAlem (approved):

controllers/buildpipeline/buildpipeline_adapter.go
```diff
@@ -132,7 +132,8 @@ func (a *Adapter) EnsureSnapshotExists() (controller.OperationResult, error) {
 	err = a.client.Create(a.context, expectedSnapshot)
 	if err != nil {
 		a.logger.Error(err, "Failed to create Snapshot")
-		return controller.RequeueWithError(err)
+		// we cannot create a snapshot and we don't want to block resources, user has to retry
+		return a.removeFinalizerAndStopProcessing()
```
hongweiliu17, line 136: We need to check if the `err` is about the reached snapshot quota then remove the finalizer, as I wanted to do in https://github.com/redhat-appstudio/integration-service/pull/481#discussion_r1429759860, I think.
The retry for other err is still needed here.

hongweiliu17 (changes requested):

controllers/buildpipeline/buildpipeline_adapter.go
MartinBasti: updated

Josh-Everett (approved):

controllers/buildpipeline/buildpipeline_adapter.go
```diff
@@ -132,6 +133,10 @@ func (a *Adapter) EnsureSnapshotExists() (controller.OperationResult, error) {
 	err = a.client.Create(a.context, expectedSnapshot)
 	if err != nil {
 		a.logger.Error(err, "Failed to create Snapshot")
+		if errors.IsForbidden(err) {
+			// we cannot create a snapshot and we don't want to block resources, user has to retry
```
dheerajodha, line 137: nit: can we specify an example (like: we hit the snapshot quota) of why we weren't able to create a snapshot? Because we are requeuing for other kinds of errors
```suggestion
			// we cannot create a snapshot (possibly because the snapshot quota is hit) and we don't want to block resources, user has to retry
```

dheerajodha (approved):
LGTM

hongweiliu17 (approved):
LGTM

PR #545, opened 2024-02-01
fix(STONEINTG-779): fixes for annotating build PLR

hongweiliu17:
* remove finalizer after annotating build PLR
* annotate build PLR with successful only when the real snapshot is created
* requeue when failing to annotate build PLR or remove finalizer
* a few other small fixes

### DIFF
--- a/controllers/buildpipeline/buildpipeline_adapter.go
+++ b/controllers/buildpipeline/buildpipeline_adapter.go
@@ -63,19 +63,33 @@ func NewAdapter(pipelineRun *tektonv1.PipelineRun, component *applicationapiv1al
 
 // EnsureSnapshotExists is an operation that will ensure that a pipeline Snapshot associated
 // to the build PipelineRun being processed exists. Otherwise, it will create a new pipeline Snapshot.
-func (a *Adapter) EnsureSnapshotExists() (controller.OperationResult, error) {
-	var err error
+func (a *Adapter) EnsureSnapshotExists() (result controller.OperationResult, err error) {
+	// a marker if we should remove finalizer from build PLR
+	var canRemoveFinalizer bool
+
 	defer func() {
 		err = tekton.AnnotateBuildPipelineRunWithCreateSnapshotAnnotation(a.context, a.pipelineRun, a.client, err)
 		if err != nil {
 			a.logger.Error(err, "Could not add create snapshot annotation to build pipelineRun", h.CreateSnapshotAnnotationName, a.pipelineRun)
+			// requeue when err is not nil
+			result, err = controller.RequeueWithError(err)
+		}
+
+		if canRemoveFinalizer {
+			err = h.RemoveFinalizerFromPipelineRun(a.client, a.logger, a.context, a.pipelineRun, h.IntegrationPipelineRunFinalizer)
+			if err != nil {
+				a.logger.Error(err, "Failed to remove finalizer from build pipelineRun")
+				// requeue when err is not nil
+				result, err = controller.RequeueWithError(err)
+			}
 		}
 	}()
 	if !h.HasPipelineRunSucceeded(a.pipelineRun) {
 		if h.HasPipelineRunFinished(a.pipelineRun) || a.pipelineRun.GetDeletionTimestamp() != nil {
-			// The pipeline run  has failed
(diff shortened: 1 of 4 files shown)


controllers/buildpipeline/buildpipeline_adapter.go
```diff
@@ -65,17 +65,28 @@ func NewAdapter(pipelineRun *tektonv1.PipelineRun, component *applicationapiv1al
...
+	var canRemoveFinalizer bool
+
 	defer func() {
 		err = tekton.AnnotateBuildPipelineRunWithCreateSnapshotAnnotation(a.context, a.pipelineRun, a.client, err)
 		if err != nil {
 			a.logger.Error(err, "Could not add create snapshot annotation to build pipelineRun", h.CreateSnapshotAnnotationName, a.pipelineRun)
 		}
+
+		if canRemoveFinalizer {
+			err := h.RemoveFinalizerFromPipelineRun(a.client, a.logger, a.context, a.pipelineRun, h.IntegrationPipelineRunFinalizer)
```
MartinBasti, line 78: because this is in defer function, if we fail to remove finalizer, it may stuck there because we won't retry

MartinBasti: I somehow feel that this defer function is not the right solution. I don't like the magic with global variables.

we should probably explicitly call the function to update annotation on places where it makes sense

hongweiliu17: Yes, you are right. I realized it when I saw that annotation** in defer before, so I can't ignore it...
Maybe we have to remove the `defer`. Do you have any idea?

14rcole: Right now we're running into an increasing number of things we may need to do when the ensure functions exit.  We may want to have a general function that handles all parts of the cleanup that just gets called everywhere.  That way we're not dealing with 6 or 7 lines of repeated code everywhere we exit an Ensure function.

hongweiliu17: I would still use defer but rewrite the returned operationResult

controllers/buildpipeline/buildpipeline_adapter.go
```diff
@@ -63,19 +63,41 @@ func NewAdapter(pipelineRun *tektonv1.PipelineRun, component *applicationapiv1al
...
 	defer func() {
 		err = tekton.AnnotateBuildPipelineRunWithCreateSnapshotAnnotation(a.context, a.pipelineRun, a.client, err)
 		if err != nil {
 			a.logger.Error(err, "Could not add create snapshot annotation to build pipelineRun", h.CreateSnapshotAnnotationName, a.pipelineRun)
+			// requeue when err is not nil
+			result = controller.OperationResult{
+				RequeueDelay:   0,
+				RequeueRequest: true,
+				CancelRequest:  false,
+			}
```
MartinBasti, line 79: Logic looks good but I'm curios if we could use this to make readability better?
```suggestion
			result, err = controller.RequeueWithError(err)
```

kasemAlem (approved):

controllers/buildpipeline/buildpipeline_adapter.go
hongweiliu17: Updated. Thanks.

MartinBasti (approved):

tekton/build_pipeline.go
```diff
@@ -46,6 +46,10 @@ func AnnotateBuildPipelineRunWithCreateSnapshotAnnotation(ctx context.Context, p
 	message := ""
 	status := ""
 	if ensureSnapshotExistsErr == nil {
+		if !metadata.HasAnnotation(pipelineRun, SnapshotNameLabel) {
+			// do nothing fo in progress build PLR
```
dirgim, line 50: nit: `do nothing fo in progress build PLR` -> `do nothing for in progress build PLR`

dirgim (approved):
LGTM aside the nit

tekton/build_pipeline.go
hongweiliu17: Thanks! Updated.

PR #549, opened 2024-02-01
fix(STONEINTG-752): don't reconcile unnecessary Integration PLRs

dheerajodha:
* Currently, we are processing Intg PLRs that are already finished, are marked for deletion, and don't contain a finalizer.
* This causes a problem when a user deletes the TaskRun(s) of this PLR, and then deletes it.
* Because our service tries to reconcile it as it just got marked for deletion. And then hits this bug when we aren't able to fetch all/some of its TaskRuns.

### DIFF
--- a/controllers/integrationpipeline/integrationpipeline_adapter.go
+++ b/controllers/integrationpipeline/integrationpipeline_adapter.go
@@ -19,8 +19,9 @@ package integrationpipeline
 import (
 	"context"
 	"fmt"
-	"k8s.io/apimachinery/pkg/api/errors"
 	"strings"
+
+	"k8s.io/apimachinery/pkg/api/errors"
 
 	applicationapiv1alpha1 "github.com/redhat-appstudio/application-api/api/v1alpha1"
 	"github.com/redhat-appstudio/integration-service/gitops"
@@ -166,6 +167,20 @@ func GetIntegrationPipelineRunStatus(adapterClient client.Client, ctx context.Co
 		} else {
 			return intgteststat.IntegrationTestStatusInProgress, fmt.Sprintf("Integration test is running as pipeline run '%s'", pipelineRun.Name), nil
 		}
+	}
+
+	taskRuns, err := h.GetAllTaskRunsWithMatchingPipelineLabel(adapterClient, ctx, pipelineRun)
+	if err != nil {
+		return intgteststat.IntegrationTestStatusTestInvalid, fmt.Sprintf("Unable to get all the TaskRun(s) related to the pipelineRun '%s'", pipelineRun.Name), err
+	}
+
+	taskRunsInClusterCount := len(taskRuns.Items)
+	taskRunsInChildRefCount := len(pipelineRun.Status.ChildReferences)
+
+	if taskRunsInClusterCount != taskRunsInChildRefCount {
+		return intgteststat.IntegrationTestStatusTestInvalid, fmt.Sprintf("Failed to determine status of pipelinerun '%s'"+
+			", due to mismatch in TaskRuns present in cluster (%v) and those referenced within childReferences (%v)",
+			pipelineRun.Name, taskRunsInClusterCount, taskRunsInChildRefCount), nil
 	}
 
 	outcome, err := h.GetIntegrationPipelineRunOutcome(adapterClient, ctx, pipelineRun)
--- a/controllers/integrationpipeline/integrationpipeline_adapter.go
+++ b/controllers/integrationpipeline/integrationpipeline_adapter.go
@@ -85,7 +85,7 @@ func (a *Adapter) EnsureStatusReportedInSnapshot() (controller.OperationResult, 
 			return err
 		}
 
-		pipelinerunStatus, detail, err = GetIntegrationPipelineRunStatus(a.client, a.context, a.pipelineRun)
+		pipelinerunStatus, detail, err = a.GetIntegrationPipelineRunStatus(a.client, a.context, a.pipelineRun)
 		if err != nil {
 			return err
 		}
@@ -158,7 +158,7 @@ func (a *Adapter) EnsureEphemeralEnvironmentsCleanedUp() (controller.OperationRe
 }
 
 // GetIntegrationPipelineRunStatus checks the Tekton results for a given PipelineRun and returns status of test.
-func GetIntegrationPipelineRunStatus(adapterClient client.Client, ctx context.Context, pipelineRun *tektonv1.PipelineRun) (intgteststat.IntegrationTestStatus, string, error) {
+func (a *Adapter) GetIntegrationPipelineRunStatus(adapterClient client.Client, ctx context.Context, pipelineRun *tektonv1.PipelineRun) (intgteststat.IntegrationTestStatus, string, error) {
 	// Check if the pipelineRun finished from the condition of status
 	if !h.HasPipelineRunFinished(pipelineRun) {
 		// Mark the pipelineRun's status as "Deleted" if its not finished yet and is marked for deletion (with a non-nil deletionTimestamp)
@@ -169,12 +169,12 @@ func GetIntegrationPipelineRunStatus(adapterClient client.Client, ctx context.Co
 		}
 	}
 
-	taskRuns, err := h.GetAllTaskRunsWithMatchingPipelineLabel(adapterClient, ctx, pipelineRun)
+	taskRuns, err := a.loader.GetAllTaskRunsWithMatchingPipelineLabel(adapterClient, ctx, pipelineRun)
 	if err != nil {
 		return intgteststat.IntegrationTestStatusTestInvalid, fmt.Sprintf("Unable to get all the TaskRun(s) related to the pipelineRun '%s'", pipelineRun.Name), err
 	}
 
-	taskRunsInClusterCount := len(taskRuns.Items)
+	taskRunsInClusterCount := len(*taskRuns)
 	taskRunsInChildRefCount := len(pipelineRun.Status.ChildReferences)
 
 	if taskRunsInClusterCount != taskRunsInChildRefCount {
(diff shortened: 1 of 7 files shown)


dheerajodha:
[RHTAPBUGS-1096](https://issues.redhat.com/browse/RHTAPBUGS-1096)
/test integration-service-e2e

controllers/integrationpipeline/integrationpipeline_controller.go
```diff
@@ -85,6 +87,13 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return ctrl.Result{}, err
 	}
 
+	if helpers.HasPipelineRunFinished(pipelineRun) && pipelineRun.GetDeletionTimestamp() != nil &&
```
MartinBasti, line 90: Why we don't handle the issue on the place where it occurs instead of adding triple condition into reconciler, that may not even catch all issues?

dheerajodha: I was thinking about adding the fix at the place where it occurs but it felt like a bandage for an injury we're not supposed to have in the first place. I wondered if we even need to process these kinds of PLRs caught by the above `if` condition and the answer was no. But I'm curious about what you think about this.

MartinBasti: sorry couldn't find time earlier,  but it looks much better

helpers/integration.go
```diff
@@ -356,6 +356,25 @@ func GetAllChildTaskRunsForPipelineRun(adapterClient client.Client, ctx context.
 	return taskRuns, nil
 }
 
+// GetAllTaskRunsWithMatchingPipelineLabel finds all Child TaskRuns
+// whose "tekton.dev/pipeline" label points to the given PipelineRun
+func GetAllTaskRunsWithMatchingPipelineLabel(adapterClient client.Client, ctx context.Context, pipelineRun *tektonv1.PipelineRun) (*tektonv1.TaskRunList, error) {
```
dirgim, line 361: I think this function would be better placed in the [loader](https://github.com/redhat-appstudio/integration-service/blob/main/loader/loader.go) package.

controllers/integrationpipeline/integrationpipeline_adapter.go
```diff
@@ -168,6 +169,20 @@ func GetIntegrationPipelineRunStatus(adapterClient client.Client, ctx context.Co
 		}
 	}
 
+	taskRuns, err := h.GetAllTaskRunsWithMatchingPipelineLabel(adapterClient, ctx, pipelineRun)
+	if err != nil {
+		return -1, "", err
```
14rcole, line 174: I don't know if returning `-1` in place of a blank constant is the right approach.  If you've seen it done that way elsewhere let me know, I could be wrong here.  But I would just return `intgteststat.IntegrationTestStatusTestInvalid`.  Alternatively you could add a new status for `IntegrationTestStatusUnknown` with a comment that that's to be used for edge cases like this one where we can't properly determine the status.

dheerajodha: That's a good point.
It's used [here](https://github.com/redhat-appstudio/integration-service/blob/a054d6cb69d24bedb7e51d23000fae8db64d2d6b/pkg/integrationteststatus/integration_test_status.go#L136) when we create a new instance of `IntegrationTestStatusDetail`type.

Currently, when we return `-1, "", err` (non-nil error) from `GetIntegrationPipelineRunStatus()` function, it gets requeued by the controller [here](https://github.com/redhat-appstudio/integration-service/blob/a054d6cb69d24bedb7e51d23000fae8db64d2d6b/controllers/integrationpipeline/integrationpipeline_adapter.go#L102) and we never get to make this constant visible to the users by calling `WriteIntegrationTestStatusesIntoSnapshot()`. This is why I went with `-1`, although I agree with you that it's not the best solution. Wdyt?

helpers/integration.go
dheerajodha: Thanks for letting me know, I've made the required changes for this move.

controllers/integrationpipeline/integrationpipeline_adapter.go
dirgim: My vote is for returning `intgteststat.IntegrationTestStatusTestInvalid` since the issue is very similar to not being able to find taskRuns referenced in childReferences.

dheerajodha: I've updated the code to return `intgteststat.IntegrationTestStatusTestInvalid`. Once Ryan comes back and thinks otherwise, I will make the necessary updates to the code after we collectively reach an agreement.

helpers/integration.go
dheerajodha: Oops, done! PTAL

controllers/integrationpipeline/integrationpipeline_adapter.go
```diff
@@ -168,6 +169,20 @@ func GetIntegrationPipelineRunStatus(adapterClient client.Client, ctx context.Co
 		}
 	}
 
+	taskRuns, err := a.loader.GetAllTaskRunsWithMatchingPipelineLabel(adapterClient, ctx, pipelineRun)
+	if err != nil {
```
MartinBasti, line 173: I'm maybe too much of rebel, how about just failing without retries?


```diff
diff --git a/controllers/integrationpipeline/integrationpipeline_adapter.go b/controllers/integrationpipeline/integrationpipeline_adapter.go
index 04a2fcf..2ba9032 100644
--- a/controllers/integrationpipeline/integrationpipeline_adapter.go
+++ b/controllers/integrationpipeline/integrationpipeline_adapter.go
@@ -170,7 +170,7 @@ func GetIntegrationPipelineRunStatus(adapterClient client.Client, ctx context.Co
 
        outcome, err := h.GetIntegrationPipelineRunOutcome(adapterClient, ctx, pipelineRun)
        if err != nil {
-               return intgteststat.IntegrationTestStatusTestFail, "", fmt.Errorf("failed to evaluate inegration test results: %w", err)
+               return intgteststat.IntegrationTestStatusTestInvalid, fmt.Sprintf("failed to evaluate inegration test results: %s", err), nil
        }
 
        if !outcome.HasPipelineRunPassedTesting() {
```

dheerajodha: I agree with you that you're too much of a rebel

dheerajodha: This is interesting, thanks for the suggestion! I'm now in the process of deploying this suggested change on a cluster to see how it looks like and then post here for comparison to decide which one to go with.

loader/loader.go
```diff
@@ -58,6 +58,7 @@ type ObjectLoader interface {
 	GetAutoReleasePlansForApplication(c client.Client, ctx context.Context, application *applicationapiv1alpha1.Application) (*[]releasev1alpha1.ReleasePlan, error)
 	GetScenario(c client.Client, ctx context.Context, name, namespace string) (*v1beta1.IntegrationTestScenario, error)
 	GetAllEnvironmentsForScenario(c client.Client, ctx context.Context, integrationTestScenario *v1beta1.IntegrationTestScenario) (*[]applicationapiv1alpha1.Environment, error)
+	GetAllTaskRunsWithMatchingPipelineLabel(c client.Client, ctx context.Context, pipelineRun *tektonv1.PipelineRun) (*[]tektonv1.TaskRun, error)
```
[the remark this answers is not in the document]
dheerajodha: I saw [this comment](https://github.com/redhat-appstudio/integration-service/pull/557#discussion_r1484191542) by Kruno, should we handle this as part of [STONEINTG-774](https://issues.redhat.com/browse/STONEINTG-774)? wdyt?

controllers/integrationpipeline/integrationpipeline_adapter.go
dheerajodha: So, this is how the status looks like as part of the current logic within this PR:
image

dheerajodha: And this is how the status looks like as part of the changes suggested within this thread:
image

dheerajodha: Honestly, both of these messages look descriptive enough to clearly explain why the status is marked as Invalid. Compared to Approach 2, Approach 1's implementation makes some extra calls and duplicates some logic. I'm leaning towards the 2nd approach. What do you all think?

MartinBasti: the main issue is, that the second approach won't retry in any case

dheerajodha: Oh right, hmm that's not good...

dheerajodha: Generally, not retrying for all errors, doesn't sound good.
Looking at the current logic, we would miss out on retrying [this error](https://github.com/redhat-appstudio/integration-service/blob/main/helpers/integration.go#L161-L165):
```
error while compiling json data for schema validation ....
```

MartinBasti: yes, but wrong example, that's unrecoverable error that would need code change, so retrying won't solve anything in that case

I'm fine with merging it as is

dirgim: +1 on merging as is as discussed on standup.

dirgim (approved):
LGTM, just needs a rebase

dheerajodha:
FTR, now when we try to reproduce this bug, this is how the test's status gets updated within the Snapshot:
image

MartinBasti (approved):

PR #551, opened 2024-02-05
fix(STONEINTG-687): stop reconciliation when app is not found

MartinBasti:
When application was removed, we cannot do much and should stop reconciliation instead of retrying indefinitely.

### DIFF
--- a/controllers/component/component_controller.go
+++ b/controllers/component/component_controller.go
@@ -76,9 +76,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 				return ctrl.Result{}, err
 			}
 		}
-		logger.Error(err, "Failed to get Application from the component",
-			"Component.Name", component.Name)
-		return ctrl.Result{}, err
+		return helpers.HandleLoaderError(logger, err, "Application", "Component")
 	}
 
 	if application == nil {
--- a/controllers/component/component_controller_test.go
+++ b/controllers/component/component_controller_test.go
@@ -121,12 +121,12 @@ var _ = Describe("ComponentController", func() {
 		}).Should(BeNil())
 	})
 
-	It("can fail when Reconcile fails to prepare the adapter when Application is not found", func() {
+	It("Should stop Reconciliation when Application is not found", func() {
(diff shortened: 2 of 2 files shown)


dirgim (approved):
LGTM

sonam1412 (approved):
lgtm

kasemAlem (approved):

hongweiliu17 (approved):
LGTM

dheerajodha (approved):

controllers/component/component_controller.go
```diff
@@ -75,6 +75,8 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 			if err := helpers.RemoveFinalizerFromComponent(r.Client, logger, ctx, component, helpers.ComponentFinalizer); err != nil {
 				return ctrl.Result{}, err
 			}
+			logger.Error(err, "Application referenced in the component not found. Stopping reconcilation of the request.")
+			return ctrl.Result{}, nil
```
dheerajodha, line 79: To be consistent, can we instead call `HandleLoaderError` ([like this](https://github.com/redhat-appstudio/integration-service/blob/01731e8b4213486d457dcd9063a8915d5179f19d/controllers/integrationpipeline/integrationpipeline_controller.go#L99))? so that we can remove the `logger.Error` calls on Lines 78 and 81.

dheerajodha (changes requested):

controllers/component/component_controller.go
MartinBasti: oh nice, updated, thanks!

dheerajodha: @MartinBasti Can you also change the names of the resources to this instead?

`return helpers.HandleLoaderError(logger, err, "Application", "Component")`

MartinBasti: yes, I'm stupid I wanted to do it but my attention went somewhere else

I shouldn't be allowed to write code on mondays

dheerajodha: :laughing:  Thanks for the fix!

dheerajodha (approved):

Josh-Everett (approved):

PR #557, opened 2024-02-06
fix(STONEINTG-798): use labels to find Snapshots for build pipelineRun

dirgim:
* Stop finding matching Snapshots by comparing the contents of existing ones when determining if a new Snapshot needs to be created for a given build pipelineRun
* Switch to using the Snapshot annotation on the build pipelineRun, and also attempt to find Snapshots that are labeled for that PLR as a backup
* Add GetAllSnapshotsForBuildPipelineRun loader function to facilitate the above

### DIFF
--- a/controllers/buildpipeline/buildpipeline_adapter.go
+++ b/controllers/buildpipeline/buildpipeline_adapter.go
@@ -108,6 +108,36 @@ func (a *Adapter) EnsureSnapshotExists() (result controller.OperationResult, err
 		return controller.ContinueProcessing()
 	}
 
+	if _, found := a.pipelineRun.ObjectMeta.Annotations[tekton.SnapshotNameLabel]; found {
+		a.logger.Info("The build pipelineRun is already associated with existing Snapshot via annotation",
+			"snapshot.Name", a.pipelineRun.ObjectMeta.Annotations[tekton.SnapshotNameLabel])
+		canRemoveFinalizer = true
+		return controller.ContinueProcessing()
+	}
+
+	existingSnapshots, err := a.loader.GetAllSnapshotsForBuildPipelineRun(a.client, a.context, a.pipelineRun)
+	if err != nil {
+		a.logger.Error(err, "Failed to fetch Snapshots for the build pipelineRun")
+		return controller.RequeueWithError(err)
+	}
+	if len(*existingSnapshots) > 0 {
+		if len(*existingSnapshots) == 1 {
+			existingSnapshot := (*existingSnapshots)[0]
+			a.logger.Info("There is an existing Snapshot associated with this build pipelineRun, but the pipelineRun is not yet annotated",
+				"snapshot.Name", existingSnapshot.Name)
+			err := a.annotateBuildPipelineRunWithSnapshot(a.pipelineRun, &existingSnapshot)
+			if err != nil {
+				a.logger.Error(err, "Failed to update the build pipelineRun with snapshot name",
+					"pipelineRun.Name", a.pipelineRun.Name)
+				return controller.RequeueWithError(err)
+			}
+		} else {
+			a.logger.Info("The build pipelineRun is already associated with more than one existing Snapshot")
+		}
+		canRemoveFinalizer = true
+		return controller.ContinueProcessing()
+	}
+
 	expectedSnapshot, err := a.prepareSnapshotForPipelineRun(a.pipelineRun, a.component, a.application)
 	if err != nil {
 		// If PipelineRun result returns cusomized error update PLR annotation and exit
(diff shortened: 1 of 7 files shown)


controllers/buildpipeline/buildpipeline_adapter.go
```diff
@@ -108,6 +108,36 @@ func (a *Adapter) EnsureSnapshotExists() (result controller.OperationResult, err
...
+			return controller.RequeueWithError(err)
+		}
+		if len(*existingSnapshots) > 0 {
+			if len(*existingSnapshots) == 1 {
+				existingSnapshot := (*existingSnapshots)[0]
+				a.logger.Info("There is an existing Snapshot associated with this build pipelineRun, but the pipelineRun is not yet annotated",
+					"snapshot.Name", existingSnapshot.Name)
+				err := a.annotateBuildPipelineRunWithSnapshot(a.pipelineRun, &existingSnapshot)
+				if err != nil {
+					a.logger.Error(err, "Failed to update the build pipelineRun with new annotations",
```
MartinBasti, line 129:
```suggestion
					a.logger.Error(err, "Failed to update the build pipelineRun with snapshot name",
```

controllers/buildpipeline/buildpipeline_adapter.go
```diff
@@ -108,6 +108,36 @@ func (a *Adapter) EnsureSnapshotExists() (result controller.OperationResult, err
 		return controller.ContinueProcessing()
 	}
 
+	if _, found := a.pipelineRun.ObjectMeta.Annotations[tekton.SnapshotNameLabel]; found {
+		a.logger.Info("The build pipelineRun is already associated with existing Snapshot via annotation",
+			"snapshot.Name", a.pipelineRun.ObjectMeta.Annotations[tekton.SnapshotNameLabel])
+		canRemoveFinalizer = true
+		return controller.ContinueProcessing()
+	} else {
```
MartinBasti, line 116: We can make it more readble by not using `else` as `if` is returning from function

```
if annotated {
  return
}

existingSnaphsots ...
```

controllers/buildpipeline/buildpipeline_adapter.go
dirgim: Ack, updated.

controllers/buildpipeline/buildpipeline_adapter.go
dirgim: Agreed, updated.

MartinBasti (approved):

controllers/buildpipeline/buildpipeline_adapter.go
```diff
@@ -108,6 +108,36 @@ func (a *Adapter) EnsureSnapshotExists() (result controller.OperationResult, err
 		return controller.ContinueProcessing()
 	}
 
+	if _, found := a.pipelineRun.ObjectMeta.Annotations[tekton.SnapshotNameLabel]; found {
```
dheerajodha, line 111: what if this annotation points to a non-existent snapshot?

dirgim: I thought about that too, but the Snapshot could have been pruned or deleted by the user. The annotation is there to tell us that a single Snapshot was already created for that PLR, not that the Snapshot should still exist.

I think that we shouldn't create a new Snapshot after we created it once, otherwise we could start recreating Snapshots for old(er) pipelineRuns which I think is unexpected behavior.

loader/loader.go
```diff
@@ -58,6 +58,7 @@ type ObjectLoader interface {
 	GetAutoReleasePlansForApplication(c client.Client, ctx context.Context, application *applicationapiv1alpha1.Application) (*[]releasev1alpha1.ReleasePlan, error)
 	GetScenario(c client.Client, ctx context.Context, name, namespace string) (*v1beta1.IntegrationTestScenario, error)
 	GetAllEnvironmentsForScenario(c client.Client, ctx context.Context, integrationTestScenario *v1beta1.IntegrationTestScenario) (*[]applicationapiv1alpha1.Environment, error)
+	GetAllSnapshotsForBuildPipelineRun(c client.Client, ctx context.Context, pipelineRun *tektonv1.PipelineRun) (*[]applicationapiv1alpha1.Snapshot, error)
```
jsztuka, line 61: You gonna hate me, but can we have ctx as a first param? According to https://pkg.go.dev/context ?

dirgim: I was thinking that we should handle that as part of [STONEINTG-774](https://issues.redhat.com/browse/STONEINTG-774) for consistency.

controllers/buildpipeline/buildpipeline_adapter.go
```diff
@@ -108,6 +108,36 @@ func (a *Adapter) EnsureSnapshotExists() (result controller.OperationResult, err
...
+			a.logger.Info("There is an existing Snapshot associated with this build pipelineRun, but the pipelineRun is not yet annotated",
+				"snapshot.Name", existingSnapshot.Name)
+			err := a.annotateBuildPipelineRunWithSnapshot(a.pipelineRun, &existingSnapshot)
+			if err != nil {
+				a.logger.Error(err, "Failed to update the build pipelineRun with snapshot name",
+					"pipelineRun.Name", a.pipelineRun.Name)
+				return controller.RequeueWithError(err)
+			}
+		} else {
+			a.logger.Info("The build pipelineRun is already associated with more than one existing Snapshot")
```
dheerajodha, line 135: When would we come inside this else condition? (or is it possible to have more than 1 Snapshots for a build PLR based on the current logic?)

dirgim: It shouldn't be possible aside for possible edge cases caused by extreme API latency. I thought it would be good to cover (and log) that remote possibility just in case.

dheerajodha (approved):
Nice! LGTM

sonam1412 (approved):
lgtm

PR #562, opened 2024-02-08
chore(RHTAPWATCH-778): add snapshotgc cronjob

yftacherzog:
Adding a cronjob and associated rbac resources for running the snapshots garbage collector

### DIFF
--- a/config/snapshotgc/snapshotgc.yaml
+++ b/config/snapshotgc/snapshotgc.yaml
@@ -1,0 +1,32 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: snapshot-garbage-collector
+spec:
+  schedule: "0 5 * * *" # every day at 5AM UTC
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+            - name: test-gc
+              image: >-
+                quay.io/redhat-appstudio/integration-service:latest
+              command:
+                - /snapshotgc
+                - --zap-log-level=debug
+                - --pr-snapshots-to-keep=100
+                - --non-pr-snapshots-to-keep=700
+              imagePullPolicy: Always
+              resources:
+                requests:
+                  cpu: 1000m
+                  memory: 500Mi
+                limits:
+                  cpu: 1000m
+                  memory: 500Mi
+              securityContext:
+                readOnlyRootFilesystem: true
+                runAsNonRoot: true
+          restartPolicy: Never
+          serviceAccountName: snapshot-garbage-collector
--- a/config/rbac/kustomization.yaml
+++ b/config/rbac/kustomization.yaml
@@ -25,3 +25,6 @@ resources:
 - prometheus_service_account.yaml
 - prometheus_viewer_role.yaml
 - prometheus_role_binding.yaml
+
+# Snapshot garbage collector
+- snapshotgc_rbac.yaml
--- a/config/rbac/snapshotgc_rbac.yaml
+++ b/config/rbac/snapshotgc_rbac.yaml
@@ -1,0 +1,47 @@
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: snapshot-garbage-collector
+  namespace: system
+---
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: snapshot-garbage-collector
+rules:
+- apiGroups:
+  - appstudio.redhat.com
+  resources:
+  - releases
+  - snapshotenvironmentbindings
+  verbs:
+  - get
+  - list
+- apiGroups:
+  - appstudio.redhat.com
+  resources:
+  - snapshots
+  verbs:
+  - get
+  - list
+  - delete
+- apiGroups:
+  - ''
+  resources:
+  - namespaces
+  verbs:
+  - list
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: snapshot-garbage-collector
+subjects:
+- kind: ServiceAccount
+  name: snapshot-garbage-collector
+  namespace: system
+roleRef:
+  kind: ClusterRole
+  name: snapshot-garbage-collector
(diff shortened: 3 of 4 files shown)


config/snapshotgc/snapshotgc.yaml
```diff
@@ -0,0 +1,32 @@
...
+        spec:
+          containers:
+            - name: test-gc
+              image: >-
+                quay.io/redhat-user-workloads/rhtap-integration-tenant/integration-service/integration-service@sha256:7e803efa5aa9561d5869096d3642106d0dae1567831e1ed6f2b0c7ac3e37fdcb
+              command:
+                - /snapshotgc
+                - --zap-log-level=debug
+                - --pr-snapshots-to-keep=1
+                - --non-pr-snapshots-to-keep=1
```
yftacherzog, line 19: @MartinBasti what do you think we should set for those?

yftacherzog: Of course. We decided on `M` and `N` :)

MartinBasti: totally not 1 :D

I'd suggest to start with something big and then lowering it

config/snapshotgc/snapshotgc.yaml
```diff
@@ -0,0 +1,32 @@
...
+spec:
+  schedule: "* * * * *" # FIXME
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          containers:
+            - name: test-gc
+              image: >-
+                quay.io/redhat-user-workloads/rhtap-integration-tenant/integration-service/integration-service@sha256:7e803efa5aa9561d5869096d3642106d0dae1567831e1ed6f2b0c7ac3e37fdcb
```
MartinBasti, line 14: Probably kustomization in infra deployments should set this into digest, for dev purposes we should keep here `latest`

config/snapshotgc/snapshotgc.yaml
MartinBasti: also I think that it would be better to make it configurable, so for keep  limit we don't have to change integration service codebase everytime. Maybe use envvars and make them configurable via kustomize in infra-deploymets

config/snapshotgc/snapshotgc.yaml
yftacherzog: ok. I'll try to figure this out.

config/snapshotgc/snapshotgc.yaml
MartinBasti: 1024

MartinBasti: but quota is per both PR and push snapshots together 1024 

Let's start with 700 for push snapshots and 100 for PR snapshots, PR snapshots are basically throwaway

config/default/kustomization.yaml
```diff
@@ -16,6 +16,7 @@ bases:
 - ../crd
 - ../rbac
 - ../manager
+- ../snapshotgc
```
MartinBasti, line 19: can we not include it as default kustomization? I'd like to test it on stage longer and not promote it into production quickly (respectively, not blocking prod promotions of other features).

Once it's well tested, we can make it part of default kustomization

yftacherzog: Do you want to have there also the rbac-related stuff, or we keep those under rbac as part of the defaults?

MartinBasti: rbac is fine as is, it won't break anything if created :D

config/snapshotgc/snapshotgc.yaml
```diff
@@ -0,0 +1,32 @@
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: snapshot-garbage-collector
+spec:
+  schedule: "0 11 * * *" # every day at 11AM UTC
```
dirgim, line 6: Suggestion: maybe we can run this at a lower-traffic time, somewhere between 0-6AM UTC.

yftacherzog: done

config/snapshotgc/snapshotgc.yaml
yftacherzog: I corrected the image name and confirmed in dev that image tag is getting updated with the existing mechanism on the [lines here](https://github.com/redhat-appstudio/infra-deployments/blob/main/components/integration/development/kustomization.yaml#L7-L10).

The change in infra-deployment that will deploy the cronjob is [here](https://github.com/yftacherzog/infra-deployments/commit/cf498e9e6e107905d5dd834ef1b471f1779023ad) (I didn't create a PR yet).

config/snapshotgc/snapshotgc.yaml
MartinBasti: There is still this part, how to change limits easier than updating integration-service and do full release round

yftacherzog: Do you have anything in mind?

MartinBasti: To unblock testing, I can waive it now, but it's something I'd like to resolve in next PRs

MartinBasti (approved):

config/snapshotgc/snapshotgc.yaml
MartinBasti: I was suggesting env vars

yftacherzog: How would you change the value of the env var without having a release cycle?

MartinBasti: we can patch it using kustomize in infra deplyomets, without touching integration-service repo

MartinBasti: https://stackoverflow.com/questions/69828954/add-environment-variable-to-kubernetes-statefulset-with-kustomize something like this

yftacherzog: I see. Alright.

dirgim (approved):
LGTM

yftacherzog:
@MartinBasti can we have this one merged?

PR #565, opened 2024-02-12
fix(STONEINTG-751): add backoff threshold for GH authentication

Josh-Everett:
Adds a backoff timer to continue reconciliation when auth fails for github status reporting. Will allow reconciliation to continue without extraneous error logging after threshold reached.

### DIFF
--- a/controllers/statusreport/statusreport_adapter.go
+++ b/controllers/statusreport/statusreport_adapter.go
@@ -19,11 +19,11 @@ package statusreport
 import (
 	"context"
 	"fmt"
-
 	"github.com/redhat-appstudio/integration-service/api/v1beta1"
 	"github.com/redhat-appstudio/integration-service/gitops"
 	"github.com/redhat-appstudio/integration-service/metrics"
 	"github.com/redhat-appstudio/operator-toolkit/metadata"
+	"time"
 
 	applicationapiv1alpha1 "github.com/redhat-appstudio/application-api/api/v1alpha1"
 	"github.com/redhat-appstudio/integration-service/helpers"
@@ -34,6 +34,8 @@ import (
 	"github.com/redhat-appstudio/operator-toolkit/controller"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
+
+const SnapshotRetryTimeout = time.Duration(3 * time.Hour)
 
 // Adapter holds the objects needed to reconcile a snapshot's test status report.
 type Adapter struct {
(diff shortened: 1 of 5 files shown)


controllers/statusreport/statusreport_adapter.go
```diff
@@ -74,9 +77,7 @@ func (a *Adapter) EnsureSnapshotTestStatusReportedToGitHub() (controller.Operati
 
 	for _, reporter := range reporters {
 		if err := reporter.ReportStatusForSnapshot(a.client, a.context, &a.logger, a.snapshot); err != nil {
-			a.logger.Error(err, "failed to report test status to github for snapshot",
```
dirgim, line 77: I think it would be good to keep the log error here so we can pick up on the issue when debugging.

controllers/statusreport/statusreport_adapter.go
```diff
@@ -322,3 +325,17 @@ func (a *Adapter) findUntriggeredIntegrationTestFromStatus(integrationTestScenar
 	}
 	return ""
 }
+
+// RequeueIfYoungerThanThreshold checks if the adapter' snapshot is younger than the threshold defined
+// in the function.  If it is, the function returns an operation result instructing the reconciler
+// to requeue the object and the error message passed to the function.  If not, the function returns
+// an operation result instructing the reconciler NOT to requeue the object.
+func (a *Adapter) RequeueIfYoungerThanThreshold(retErr error) (controller.OperationResult, error) {
```
dirgim, line 333: Non-blocking suggestion: this function could be useful to other controllers too, so it might not be bad to move it to a package outside this adapter so it's more easily available elsewhere.

dheerajodha: +1 on moving this function to a common location so it can be used both by statusreport, snapshot ([as here](https://github.com/redhat-appstudio/integration-service/blob/037767a3f1f57fb6a1a1108dc01795b4e48bbabc/controllers/snapshot/snapshot_adapter.go#L942)), and other controllers.

MartinBasti: Also it can be generic for any object, not just Snapshot

helpers/integration.go
```diff
@@ -546,3 +548,14 @@ func AddFinalizerToScenario(adapterClient client.Client, logger IntegrationLogge
...
+
+func IsSnapshotfYoungerThanThreshold(snapshot *applicationapiv1alpha1.Snapshot) bool {
+	snapshotCreationTime := snapshot.GetCreationTimestamp().Time
+	durationSinceSnapshotCreation := time.Since(snapshotCreationTime)
+
+	if durationSinceSnapshotCreation < SnapshotRetryTimeout {
+		return true
+	} else {
+		return false
+	}
```
MartinBasti, line 560:
```suggestion
	return durationSinceSnapshotCreation < SnapshotRetryTimeout

```

dirgim:
Checks are failing mainly because of the (now) unused `time` import [here](https://github.com/redhat-appstudio/integration-service/pull/565/files#diff-8e718c6bf06f754db7e9e94c6f74b5c6704941dff95d637289e3c4cc194c95a0R25).

helpers/integration.go
Josh-Everett: Yes agreed, not sure what I was thinking here.

controllers/statusreport/statusreport_adapter.go
Josh-Everett: updated to be generic in most recent patch

MartinBasti (approved):

helpers/integration.go
```diff
@@ -546,3 +547,10 @@ func AddFinalizerToScenario(adapterClient client.Client, logger IntegrationLogge
 
 	return nil
 }
+
+func IsObjectYoungerThanThreshold(obj metav1.Object, threshold time.Duration) bool {
```
dirgim, line 551: Let's add a simple unit test for this function just in case.

hongweiliu17: Agree to cover the more code in UT.

dirgim (approved):
LGTM

dirgim:
Looks like the PR needs a rebase on latest main before merging.

PR #573, opened 2024-02-15
refactor(STONEINTG-769): prepare for new reporter

MartinBasti:
Refactor code to allow to add new reporter easier.

Changes:
* `Status` is reponsible to decide when and what should be reported into git providers
* Status is also repsonsible to fetch test results and generate markdonw output instead of reporter. reporter will just report it. In case of need this can be updated to templating in future, but reporter shouldn't be repsonsible for fetching task results.
* `Reporter` is responsible to only pass data in correct format, and map values to values accepted by git provider (for example status value)
* Only one reporter is returned, we cannot support multiple reporters (PR/MR is always from the single provider)
* Interfaces and mocking for unittesting implemented.

## Todo
- [x] Fix commented tests
- [x] Test coverage


status/format.go
```diff
@@ -66,26 +66,10 @@ func FormatSummary(taskRuns []*helpers.TaskRun) (string, error) {
 	return buf.String(), nil
 }
 
-// FormatCommentForFinishedPipelineRun builds a markdown comment for a list of finished integration TaskRuns
-func FormatCommentForFinishedPipelineRun(title string, results []*helpers.TaskRun) (string, error) {
```
jsztuka, line 70: Is this no longer needed or we are formating comments different way?

MartinBasti: It's formatted in the same way without duplication caused by this function

status/status.go
```diff
@@ -16,70 +16,188 @@ limitations under the License.
...
 	}
 }
 
-// NewAdapter constructs an Adapter with optional params, if specified.
-func NewAdapter(logger logr.Logger, k8sClient client.Client, opts ...AdapterOption) *Adapter {
-	adapter := Adapter{
-		logger:         logger,
-		k8sClient:      k8sClient,
-		githubReporter: NewGitHubReporter(logger, k8sClient),
+// Get reporter which is able to process snapshot, nil means no suitable reporter found
```
dirgim, line 83: Tiny nit: should start with `GetReporter`

status/status.go
```diff
@@ -16,70 +16,188 @@ limitations under the License.
...
+	if githubReporter.Detect(snapshot) {
+		return githubReporter
 	}
 
-	for _, opt := range opts {
-		opt(&adapter)
+	return nil
+}
+
+// Report status of all integration tests into Pull Request
```
dirgim, line 93: Tiny nit: should start with `ReportSnapshotStatus`

dirgim:
Another tiny nit: the commit message has some typos, e.g. `reponsible` -> `responsible`, `repsonsible` -> `responsible`, and `markdonw` -> `markdown`.

status/status_test.go
```diff
@@ -18,43 +18,410 @@ package status_test
...
+	})
+
+	It("doesn't report anything when there are not test results", func() {
+
+		mockReporter.EXPECT().Initialize(gomock.Any(), gomock.Any()).Times(0)   // without test results reporter shouldn't be initialized
+		mockReporter.EXPECT().ReportStatus(gomock.Any(), gomock.Any()).Times(0) // without test results reported shouldn't report status
+
+		st := status.NewStatus(logr.Discard(), nil)
+		err := st.ReportSnapshotStatus(context.Background(), mockReporter, githubSnapshot)
+		Expect(err).To(Succeed())
```
dirgim, line 326: non-blocking suggestion: While `Succeed` works here just fine, I think using `Expect(err).NotTo(HaveOccurred())` is easier to read when we're checking an err variable.

status/reporter_github.go
```diff
@@ -52,14 +49,14 @@ type StatusUpdater interface {
...
 	// Update status of PR
-	UpdateStatus(ctx context.Context, integrationTestStatusDetail intgteststat.IntegrationTestStatusDetail) error
+	UpdateStatus(ctx context.Context, report TestReport) error
 }
 
 // CheckRunStatusUpdater updates PR status using CheckRuns (when application integration is enabled in repo)
 type CheckRunStatusUpdater struct {
 	ghClient          github.ClientInterface
 	k8sClient         client.Client
-	logger            *helpers.IntegrationLogger
```
dirgim, line 62: Just curious about this change, is it because we don't gain anything by using `IntegrationLogger` or is there something more behind it?

MartinBasti: I didn't like that helpers are crossimported everywhere and we don't need anything from IntegrationLogger here. 
Application name is already in underlying logger

dirgim: Fair enough, thanks for clarifying.

status/status.go
MartinBasti: udpated

status/status.go
MartinBasti: updated

status/status_test.go
MartinBasti: updated

dirgim: I'm still seeing it on this line, but can see it updated on other lines.

MartinBasti: updated everywhere

dirgim: Thanks!

dirgim (approved):
LGTM

hongweiliu17 (approved):
LGTM

PR #578, opened 2024-02-19
fix: simplify unittest for GetAllTaskRunsWithMatchingPipelineLabel

dirgim:
* The current version of unittests is flaky when run in GitHub actions because of latency when creating the TaskRun
* See [this failed run](https://github.com/redhat-appstudio/integration-service/actions/runs/7957738200/job/21721416963) for example

### DIFF
--- a/loader/loader_test.go
+++ b/loader/loader_test.go
@@ -409,6 +409,17 @@ var _ = Describe("Loader", Ordered, func() {
 			},
 		}
 		Expect(k8sClient.Create(ctx, hasBinding)).Should(Succeed())
+
+		taskRunSample = &tektonv1.TaskRun{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "taskrun-sample",
+				Namespace: "default",
+				Labels: map[string]string{
+					"tekton.dev/pipeline": buildPipelineRun.Name,
+				},
+			},
+		}
+		Expect(k8sClient.Create(ctx, taskRunSample)).Should(Succeed())
 	})
 
 	AfterAll(func() {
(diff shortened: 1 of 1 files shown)


loader/loader_test.go
```diff
@@ -763,36 +782,4 @@ var _ = Describe("Loader", Ordered, func() {
 			Expect(fetchedScenario.Spec).To(Equal(integrationTestScenario.Spec))
 		})
 	})
-
-	When("there's no TaskRuns present in the cluster that are associated with the given pipelineRun", func() {
-		It("ensures that GetAllTaskRunsWithMatchingPipelineLabel returns 0 TaskRuns", func() {
-			taskRuns, err := loader.GetAllTaskRunsWithMatchingPipelineLabel(k8sClient, ctx, integrationPipelineRun)
-			Expect(err).ToNot(HaveOccurred())
-			Expect(*taskRuns).To(BeEmpty())
-		})
```
dheerajodha, line 772: should this be re-added or this one's flaky too?

dheerajodha (commented):
Thanks for the PR!

loader/loader_test.go
dirgim: I think that needing this test causes flakiness because we need to either delete or create the taskRun, which then runs into latency issues when run in lower-resource environments.
Since [the case when there is a missing taskRun](https://github.com/redhat-appstudio/integration-service/blob/60be0fd0b8a656cf924431ba30bccdc85ad297b9/controllers/integrationpipeline/integrationpipeline_adapter_test.go#L1115-L1127) is already covered in adapter unit tests, I think this should be enough.

dheerajodha: Thanks! makes sense

dheerajodha:
Filed [KONFLUX-2210](https://issues.redhat.com/browse/KONFLUX-2210) to improve the failure message posted by the ci-helper app in the [latest CI failure](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/redhat-appstudio_integration-service/578/pull-ci-redhat-appstudio-integration-service-main-integration-service-e2e/1759523872912183296).

MartinBasti (approved):

PR #582, opened 2024-02-20
fix(STONEINTG-776): [unit-test] point the Snapshot to a non-existent Application

dheerajodha:
Before this commit:

* Currently, this unit test fails intermittently because it first deletes the Application and then reconciles the related Snapshot, in the hope that the controller will stop the reconciliation because of the missing App and mark the Snapshot as Invalid.
* But since, when the App gets deleted, Snapshot also gets queued for deletion. So, this introduces a race condition where we're trying to reconcile a Snapshot, that's marked for deletion, hoping it gets successfully reconciled before the deletion.

After this commit:

* Instead of deleting the App and reconciling the Snapshot, we are pointing the Snapshot to a non-existent App to replicate the same behaviour.
* This eliminates the race condition.

### DIFF
--- a/controllers/snapshot/snapshot_controller_test.go
+++ b/controllers/snapshot/snapshot_controller_test.go
@@ -194,15 +194,8 @@ var _ = Describe("SnapshotController", func() {
 	})
 
 	It("Does not return an error if the application cannot be found", func() {
-		err := k8sClient.Delete(ctx, hasApp)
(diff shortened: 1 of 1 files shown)


dirgim (approved):
LGTM

hongweiliu17 (approved):
LGTM

PR #586, opened 2024-02-20
fix(STONEINTG-752): fetch TaskRuns using "pipelineRun" label instead

dheerajodha:
Before this commit

* We were marking an Integration PLR as "Invalid" if the number of TaskRuns mentioned in its '.status.childReferences' field is NOT same as the number of TaskRuns fetched from the cluster using "tekton.dev/pipeline" label.
* This caused a bug because the label "tekton.dev/pipeline" points to the pipeline, that stays unique on every run.
* So, it can happen that we get 0 TaskRuns returned from cluster using this label, but the TaskRun would still be present in the cluster.

After this commit

* This changes the label to "tekton.dev/pipelineRun" to fetch the TaskRuns related to a PLR using that PLR's name.


dirgim (approved):
LGTM

PR #591, opened 2024-02-22
fix(RHTAPWATCH-847): support gitlab MR event type labels

yftacherzog:
The current implementation only took github PR event type into consideration when classifying snapshots to garbage collect. This change adds support to the equivalent labels set by gitlab.

### DIFF
--- a/cmd/snapshotgc/snapshotgc.go
+++ b/cmd/snapshotgc/snapshotgc.go
@@ -231,7 +231,10 @@ func getSnapshotsForRemoval(
 
 	for _, snap := range snapshots {
 		label, found := snap.GetLabels()["pac.test.appstudio.openshift.io/event-type"]
-		if found && label == "pull_request" {
+		if found && (label == "pull_request" ||
+			label == "Merge Request" ||
+			label == "Merge_Request" ||
+			label == "Note") {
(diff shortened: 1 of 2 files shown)


cmd/snapshotgc/snapshotgc.go
```diff
@@ -231,7 +231,10 @@ func getSnapshotsForRemoval(
 
 	for _, snap := range snapshots {
 		label, found := snap.GetLabels()["pac.test.appstudio.openshift.io/event-type"]
-		if found && label == "pull_request" {
+		if found && (label == "pull_request" ||
+			label == "Merge Request" ||
+			label == "Merge_Request" ||
+			label == "Note") {
```
MartinBasti, line 237: Note is not enough, there is more

https://github.com/openshift-pipelines/pipelines-as-code/blob/75bc53f3a67db10312611ecd92e0e231e2307600/pkg/opscomments/comments.go#L88

yftacherzog: I see this is not released yet. The latest release includes logic similar to the one I implemented here.
I suggest that we wait until this is released and then use the function they provide directly instead of duplicating the code.

dirgim (approved):
+1

PR #602, opened 2024-03-04
fix(integrationteststatus): Deprecate SEB states

MartinBasti:
We cannot remove SEB related states due compatibility, but we can mark them as deprecated, so developers won't use them anymore.


pkg/integrationteststatus/integration_test_status.go
```diff
@@ -35,10 +35,10 @@ const (
...
 	// Integration PLR deleted for this ITS and snapshot
 	IntegrationTestStatusDeleted // Deleted
-	// The environment provision experienced error for this ITS and snapshot
-	IntegrationTestStatusEnvironmentProvisionError // EnvironmentProvisionError
-	// The SEB deployment experienced error for this ITS and snapshot
-	IntegrationTestStatusDeploymentError // DeploymentError
+	// (Deprecated) The environment provision experienced error for this ITS and snapshot
+	IntegrationTestStatusEnvironmentProvisionError_Deprecated // EnvironmentProvisionError
+	// (Deprecated) The SEB deployment experienced error for this ITS and snapshot
+	IntegrationTestStatusDeploymentError_Deprecated // DeploymentError
```
dheerajodha, line 41: just to confirm, this changes nothing for the user, right? They'd continue to see the same statuses on the older snapshots, which would never be visible on the newer Snapshots since we'd (or we've) removed the code that sets these deprecated statuses.

This change is purely for us, integration service engineers, to be able to differentiate between deprecated statuses and active ones. Right?

jsztuka (approved):
lgtm

dirgim (approved):
LGTM

pkg/integrationteststatus/integration_test_status.go
MartinBasti: exactly

dheerajodha (approved):

PR #606, opened 2024-03-05
fix(STONEINTG-769): do not block reconciliation when report fail

MartinBasti:
This is regression from https://github.com/MartinBasti/integration-service/commit/bc758969f9ec196d57002c3abe2ea130e108bcd3

Before this status has been ignored at lower level, hotfixing issue.

if PLR doesn't exist anymore, at least provide users some results withou
details, not just failing silently.

ContinueProcessing had to be removed, it would prevent unlocking
finalizers.

This is just workaround, to prevent catastrophic issues on deployments.
Future fix should be about preventing to even get to state where we are
missing PLR data.

### DIFF
--- a/controllers/statusreport/statusreport_adapter.go
+++ b/controllers/statusreport/statusreport_adapter.go
@@ -83,8 +83,6 @@ func (a *Adapter) EnsureSnapshotTestStatusReportedToGitProvider() (controller.Op
 			"snapshot.Namespace", a.snapshot.Namespace, "snapshot.Name", a.snapshot.Name)
 		if helpers.IsObjectYoungerThanThreshold(a.snapshot, SnapshotRetryTimeout) {
 			return controller.RequeueWithError(err)
-		} else {
-			return controller.ContinueProcessing()
 		}
 	}
 
--- a/status/status.go
+++ b/status/status.go
@@ -29,36 +29,13 @@ import (
 	"github.com/redhat-appstudio/integration-service/helpers"
 	intgteststat "github.com/redhat-appstudio/integration-service/pkg/integrationteststatus"
 	tektonv1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 // NamePrefix is a common name prefix for this service.
 const NamePrefix = "Red Hat Trusted App Test"
-
-// generate TestReport to be used by all reporters
-func generateTestReport(ctx context.Context, client client.Client, detail intgteststat.IntegrationTestStatusDetail, snapshot *applicationapiv1alpha1.Snapshot) (*TestReport, error) {
-	text, err := generateText(client, ctx, detail, snapshot.Namespace)
-	if err != nil {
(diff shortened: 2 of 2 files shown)


jsztuka:
Didnt catch the error in unit tests.

controllers/statusreport/statusreport_adapter.go
```diff
@@ -81,11 +81,6 @@ func (a *Adapter) EnsureSnapshotTestStatusReportedToGitProvider() (controller.Op
 	if err != nil {
 		a.logger.Error(err, "failed to report test status to git provider for snapshot",
 			"snapshot.Namespace", a.snapshot.Namespace, "snapshot.Name", a.snapshot.Name)
-		if helpers.IsObjectYoungerThanThreshold(a.snapshot, SnapshotRetryTimeout) {
-			return controller.RequeueWithError(err)
-		} else {
-			return controller.ContinueProcessing()
```
dirgim, line 87: Should we just skip this line and leave the `helpers.IsObjectYoungerThanThreshold()` call?

MartinBasti: I don't follow what you mean

dirgim: I mean that we could keep the logic to [requeue if we encounter an error within the threshold](https://github.com/redhat-appstudio/integration-service/pull/606/files#diff-d00409fb87121cfd0dfb79643900ddc48645e414a6029ded78c2a0f50b3d1c41L84-L85), but remove the call to `ContinueProcessing()` here and instead proceed to remove the finalizers in the lower lines.

hongweiliu17: I guess Kruno means keeping
```
if helpers.IsObjectYoungerThanThreshold(a.snapshot, SnapshotRetryTimeout) {
}
```
Right? I just guess.

Then if it still fails after several tries, then we remove the finalizer from PLR directly.

MartinBasti: it would block finalizers for 3 hours (for nothing)

MartinBasti: before bc75896 it was not retried at all, just well hidden in CRA creation

dirgim: My vote then is to also reduce the [SnapshotRetryTimeout](https://github.com/redhat-appstudio/integration-service/blob/1ab32ad73a03c81a0395ad67c317bf5e53ee512e/controllers/statusreport/statusreport_adapter.go#L39) to ~15mins.

MartinBasti: this creation time is not the greatest, for long running tests it may not retry even once

MartinBasti: I added new commit how to address this issue to avoid retries on something that retries won't help, please check, if you like it I can rework it

dirgim: I like the new solution, does that mean that the changes in this file don't need to happen?

dirgim: Right, fully agree on that.

MartinBasti:
PR updated with the new solution, ready for review

dirgim (approved):
LGTM

kasemAlem (approved):

PR #616, opened 2024-03-11
chore(RHTAPWATCH-766): cronjob to release allocated namespace

Omeramsc:
A saftey-mechanism in case the namespace allocated by the `allocate-namespace` task wasn't removed when the pipeline finished running. Instead, a cronjob will run periodicly to make sure all allocated namespaces which are owned by a finished pipelineRun are removed.

### DIFF
--- a/config/environment_provisioning/environment_provisioning.yaml
+++ b/config/environment_provisioning/environment_provisioning.yaml
@@ -1,0 +1,44 @@
+---
+kind: CronJob
+apiVersion: batch/v1
+metadata:
+  name: spacerequest-garbage-collector
+spec:
+  schedule: "0 5 * * *" # every day at 5AM UTC
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          restartPolicy: Never
+          serviceAccountName: appstudio-pipeline
+          containers:
+            - name: cleanup-container
+              image: quay.io/redhat-user-workloads/rhtap-o11y-tenant/tools/tools:3e4c9a30e734aa34f57a084633ccfbb73d466fbd
+              command:
+                - /bin/bash
+                - -c
+                - |
+                #!/bin/bash
+
+                ALL_PIPELINERUNS=$(oc get pipelineruns -o=jsonpath='{.items[*].metadata.name}')
+                FINISHED_PIPELINERUNS=$(oc get pipelineruns -o=jsonpath='{.items[?(@.status.conditions[*].type=="Succeeded")].metadata.name}')
+                IFS=' ' read -ra PIPELINERUNS <<< "$FINISHED_PIPELINERUNS"
+
+                for SPACEREQUEST_NAME in $(oc get spacerequests -o=jsonpath='{.items[*].metadata.name}'); do
+                  if [ -n "$PIPELINERUN_NAME" ]; then
+                      if { echo "$FINISHED_PIPELINERUNS" | grep -q "$PIPELINERUN_NAME" || ! echo "$ALL_PIPELINERUNS" | grep -q "$PIPELINERUN_NAME"; }; then
+                          oc delete spacerequest $SPACEREQUEST_NAME
+                      fi
+                  fi
+                done
+              imagePullPolicy: Always
+              resources:
+                requests:
+                  cpu: 250m
+                  memory: 125Mi
+                limits:
+                  cpu: 250m
+                  memory: 125Mi
+              securityContext:
+                readOnlyRootFilesystem: true
+                runAsNonRoot: true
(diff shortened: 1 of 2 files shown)


config/environment_provisioning/environment_provisioning.yaml
```diff
@@ -0,0 +1,42 @@
...
+          containers:
+            - name: cleanup-container
+              image: quay.io/redhat-user-workloads/rhtap-o11y-tenant/tools/tools:3e4c9a30e734aa34f57a084633ccfbb73d466fbd
+              command:
+                - /bin/bash
+                - -c
+                - |
+                #!/bin/bash
+
+                FINISHED_PIPELINERUNS=$(oc get pipelineruns -o=jsonpath='{.items[?(@.status.conditions[*].type=="Succeeded")].metadata.name}')
```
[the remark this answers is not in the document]
Omeramsc: In that case, I may remove the spacerequest not only when the associated pipelineRun finished, but also if the pipelineRun does not exist. how does that sound?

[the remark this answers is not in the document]
Omeramsc: fixed

config/environment_provisioning/environment_provisioning.yaml
```diff
@@ -0,0 +1,44 @@
...
+                  fi
+                done
+              imagePullPolicy: Always
+              resources:
+                requests:
+                  cpu: 1000m
+                  memory: 500Mi
+                limits:
+                  cpu: 1000m
+                  memory: 500Mi
```
MartinBasti, line 41: Isn't it quite much resources?

Omeramsc: I used the resources of the `snapshotgc` [cronjob](https://github.com/redhat-appstudio/integration-service/blob/main/config/snapshotgc/snapshotgc.yaml) as a rule of thumb. I'm open for changes, would you recommend a different amount?

MartinBasti: it has to process much more data than this.

like quarter of resources only?

Omeramsc: @MartinBasti  ack, changed accordingly

config/environment_provisioning/environment_provisioning.yaml
```diff
@@ -0,0 +1,44 @@
...
+                - -c
+                - |
+                #!/bin/bash
+
+                ALL_PIPELINERUNS=$(oc get pipelineruns -o=jsonpath='{.items[*].metadata.name}')
+                FINISHED_PIPELINERUNS=$(oc get pipelineruns -o=jsonpath='{.items[?(@.status.conditions[*].type=="Succeeded")].metadata.name}')
+                IFS=' ' read -ra PIPELINERUNS <<< "$FINISHED_PIPELINERUNS"
+
+                for SPACEREQUEST_NAME in $(oc get spacerequests -o=jsonpath='{.items[*].metadata.name}'); do
+                  if [ -n "$PIPELINERUN_NAME" ]; then
```
MartinBasti, line 28: Where is this variable defined?

config/environment_provisioning/environment_provisioning.yaml
```diff
@@ -0,0 +1,44 @@
...
+  schedule: "0 5 * * *" # every day at 5AM UTC
+  jobTemplate:
+    spec:
+      template:
+        spec:
+          restartPolicy: Never
+          serviceAccountName: appstudio-pipeline
+          containers:
+            - name: cleanup-container
+              image: quay.io/redhat-user-workloads/rhtap-o11y-tenant/tools/tools:3e4c9a30e734aa34f57a084633ccfbb73d466fbd
```
MartinBasti, line 16: shouldn't be this an released image, not image taken directly from user-workloads?

yftacherzog (commented):
@Omeramsc will be on leave for a few weeks. I'll try taking over this task. If I'm not able to push to his branch, I'll create a new PR and address the comments there.

One thing I see is that, as far as I understand, this cronjob should be deployed once per cluster, but at the moment, it's only interacting with resource at the current namespace, which probably won't do.

config/environment_provisioning/environment_provisioning.yaml
yftacherzog: I don't think we need this particular image. We can do with something more general-purpose with `oc` installed.

dirgim:
@Omeramsc @yftacherzog since this functionality will be implemented in a separate component, are we ok to close this PR?

yftacherzog:
> @Omeramsc @yftacherzog since this functionality will be implemented in a separate component, are we ok to close this PR?

@dirgim Yes. Thank you!

PR #623, opened 2024-03-13
WIP: feat(STONEINTG-770): create/update MR comments in gitLab

dirgim:
* Implement gitlab notes API package to create/update MR comment to show test status/details
* Add the snapshot name to test reports


status/reporter_gitlab_test.go
```diff
@@ -43,9 +43,10 @@ import (
 var _ = Describe("GitLabReporter", func() {
 
 	const (
-		repoUrl   = "https://gitlab.com/example/example"
-		digest    = "12a4a35ccd08194595179815e4646c3a6c08bb77"
-		projectID = "123"
+		repoUrl      = "https://gitlab.com/example/example"
+		digest       = "12a4a35ccd08194595179815e4646c3a6c08bb77"
+		projectID    = "123"
+		mergeRequest = "123"
```
MartinBasti, line 49: for easier debugging in case of future need, this should be different random number

status/reporter_gitlab.go
```diff
@@ -132,6 +143,47 @@ func (r *GitLabReporter) setCommitStatus(report TestReport) error {
...
+		}
+	}
+
+	return nil
+}
+
+// GetExistingNoteID returns existing GitLab note for the scenario of ref.
+func (r *GitLabReporter) GetExistingNoteID(notes []*gitlab.Note, scenarioName, snapshotName string) *int {
+	for _, note := range notes {
+		if strings.Contains(note.Body, snapshotName) && strings.Contains(note.Body, scenarioName) {
```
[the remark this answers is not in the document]
MartinBasti: it's also error prone, in case we change content of summary :(

MartinBasti (approved):
Just thoughts in comments, LGTM

status/reporter_gitlab.go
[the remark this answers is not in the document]
MartinBasti: https://github.com/openshift-pipelines/pipelines-as-code/blob/aaf2da92b5603353ba516f1155eaa0111bcb729c/pkg/provider/gitlab/gitlab.go#L215  PAC is always creating a new comment

hongweiliu17, line 178: This makes me think if we retest one merge request, it will create a new snapshot, then there will be a new comment for this scenario and this new snapshot, should we check the scenario name `strings.Contains(note.Body, scenarioName)` only to make sure the old comment for scenario and old snapshot will be overwritten to have a better user experience?
Github reporter has the same usage in [GetExistingCommentID()](https://github.com/redhat-appstudio/integration-service/blob/main/git/github/github.go#L385) but nobody uses github webhook yet so that it has not shown to users but we might need to fix as well in another ticket.
Others look good to me.

MartinBasti: that's true, we would create duplicated comments, we were discussing to use component name in checkRuns, should we apply the same here?

hongweiliu17: Yes, that is reasonable, I feel.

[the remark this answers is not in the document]
dirgim: I actually had a initial solution for this which would use templating when creating the summary ([here](https://github.com/redhat-appstudio/integration-service/blob/93f0ff0a0ef11d491d249180f8380eba1d9abcce/status/status.go#L215)), but I felt it a bit too fiddly at the time.

dirgim: At this point I'm leaning towards replicating the PaC behavior of simply always adding a new comment tbh.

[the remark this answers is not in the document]
MartinBasti: let's keep consistency with GH, we will see in future how it works

MartinBasti: ok, disregard my comment in previous thread :D

hongweiliu17 (approved):
LGTM

status/reporter_gitlab_test.go
dirgim: Ack, updated.

status/reporter_gitlab.go
dirgim: We decided to keep the logic as-is for now and we can revisit it in a follow-up ticket after we start differentiating tests by components.

kasemAlem (approved):

status/reporter_gitlab.go
dirgim: Created [STONEINTG-858](https://issues.redhat.com/browse/STONEINTG-858), let's resolve this discussion for now.

status/reporter_gitlab_test.go
dirgim: Resolving the discussion after update.

PR #627, opened 2024-03-14
fix(STONEINTG-848): drop logic for detecting latest PLR

MartinBasti:
This logic were supposed to prevent accidentally create snapshot for older build, however there is a bug that not only push build PLR are considered but it also compares pull-request build PLR and false positively prevents creation of push snapshots.

Given the radical prunning of tekton PLR, this feature has small benefit but great potential for future to break things again. Thus the best is to remove it altogether.

### DIFF
--- a/controllers/buildpipeline/buildpipeline_adapter.go
+++ b/controllers/buildpipeline/buildpipeline_adapter.go
@@ -99,19 +99,6 @@ func (a *Adapter) EnsureSnapshotExists() (result controller.OperationResult, err
 		return controller.ContinueProcessing()
 	}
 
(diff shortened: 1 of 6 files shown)


hongweiliu17 (approved):
LGTM

dirgim (approved):
LGTM to resolve the current issue, but I think this functionality still has merit when quickly merging multiple PRs in succession. 
Maybe we can create a ticket for future consideration?

MartinBasti:
> LGTM to resolve the current issue, but I think this functionality still has merit when quickly merging multiple PRs in succession. Maybe we can create a ticket for future consideration?

I can but even in that case users are expecting to see parity between build and test pipelines, so it must be done differently

PR #636, opened 2024-03-20
fix(KFLUXBUGS-1147): add component to gitHub externalID

dirgim:
* The gitHub checkRun externalID is now mapped to ID consisting of component and integration scenario name if the Snapshot contains the component label
* This prevents the reports overwriting themselves when multiple components are built from the same PR

Example commitStatus in gitLab:
image

Example checkRuns in gitHub:
image

### DIFF
--- a/status/status.go
+++ b/status/status.go
@@ -286,9 +286,14 @@ func (s *Status) generateTestReport(ctx context.Context, detail intgteststat.Int
 		return nil, fmt.Errorf("failed to generate summary message: %w", err)
 	}
 
+	fullName := fmt.Sprintf("%s / %s", NamePrefix, detail.ScenarioName)
+	if snapshot.Labels[gitops.SnapshotComponentLabel] != "" {
+		fullName = fmt.Sprintf("%s / %s", fullName, snapshot.Labels[gitops.SnapshotComponentLabel])
+	}
+
 	report := TestReport{
 		Text:           text,
-		FullName:       NamePrefix + " / " + snapshot.Name + " / " + detail.ScenarioName,
+		FullName:       fullName,
 		ScenarioName:   detail.ScenarioName,
 		SnapshotName:   snapshot.Name,
 		ComponentName:  snapshot.Labels[gitops.SnapshotComponentLabel],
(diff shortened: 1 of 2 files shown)


hongweiliu17 (approved):
LGTM

kasemAlem (approved):

status/status.go
```diff
@@ -286,11 +286,17 @@ func (s *Status) generateTestReport(ctx context.Context, detail intgteststat.Int
 		return nil, fmt.Errorf("failed to generate summary message: %w", err)
 	}
 
+	fullName := fmt.Sprintf("%s / %s", NamePrefix, detail.ScenarioName)
+	if snapshot.Labels[gitops.SnapshotComponentLabel] != "" {
+		fullName = fmt.Sprintf("%s / %s", fullName, snapshot.Labels[gitops.SnapshotComponentLabel])
```
MartinBasti, line 291: Why are we swapping order?

`app > component > test` from the most generic to the most specific looks more natural

dirgim: When testing it out, it felt more natural and useful to have the integrationTestScenario as the first item. This was especially important in gitLab since the commitStatus sometimes gets truncated, and in most cases the biggest differentiator between test runs is the scenario.
Also since we can expect some Snapshot to not have the component, having it as the last item makes it more consistent.

MartinBasti: Snapshot without component won't be reported into github, as it won't be a PR processed by build-service. It could be max an accident when user copy paste wrong data or uses another integration.

But if it's for better gitlab readability, then ok

MartinBasti (approved):

hongweiliu17 (approved):
LGTM

status/status.go
```diff
@@ -286,11 +286,17 @@ func (s *Status) generateTestReport(ctx context.Context, detail intgteststat.Int
...
+	}
+
 	report := TestReport{
 		Text:           text,
-		FullName:       NamePrefix + " / " + snapshot.Name + " / " + detail.ScenarioName,
+		FullName:       fullName,
```
hongweiliu17, line 296: Thanks!

kasemAlem (approved):

PR #653, opened 2024-03-29
feat(STONEINTG-578): don't block other ITS test if one of them fails

hongweiliu17:
* Continue running other ITS PLR if one ITS is misconfigured or ITS PLR fails

Signed-off-by: Hongwei Liu <hongliu@gmail.com>

### DIFF
--- a/controllers/snapshot/snapshot_adapter.go
+++ b/controllers/snapshot/snapshot_adapter.go
@@ -221,10 +221,10 @@ func (a *Adapter) EnsureStaticIntegrationPipelineRunsExist() (controller.Operati
 		var errsForPLRCreation error
 		for _, integrationTestScenario := range *integrationTestScenarios {
 			integrationTestScenario := integrationTestScenario //G601
-			if !gitops.IsScenarioValid(&integrationTestScenario) {
+			if !h.IsScenarioValid(&integrationTestScenario) {
 				a.logger.Info("IntegrationTestScenario is invalid, will not create pipelineRun for it",
 					"integrationTestScenario.Name", integrationTestScenario.Name)
-				scenarioStatusCondition := meta.FindStatusCondition(integrationTestScenario.Status.Conditions, gitops.IntegrationTestScenarioValid)
+				scenarioStatusCondition := meta.FindStatusCondition(integrationTestScenario.Status.Conditions, h.IntegrationTestScenarioValid)
 				testStatuses.UpdateTestStatusIfChanged(
 					integrationTestScenario.Name, intgteststat.IntegrationTestStatusTestInvalid,
 					fmt.Sprintf("IntegrationTestScenario '%s' is invalid: %s", integrationTestScenario.Name, scenarioStatusCondition.Message))
--- a/controllers/snapshot/snapshot_adapter.go
+++ b/controllers/snapshot/snapshot_adapter.go
@@ -25,6 +25,7 @@ import (
 	"time"
 
 	clienterrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/meta"
 	ctrl "sigs.k8s.io/controller-runtime"
 
 	applicationapiv1alpha1 "github.com/redhat-appstudio/application-api/api/v1alpha1"
@@ -217,8 +218,18 @@ func (a *Adapter) EnsureStaticIntegrationPipelineRunsExist() (controller.Operati
 			}
 		}()
 
+		var errsForPLRCreation error
 		for _, integrationTestScenario := range *integrationTestScenarios {
 			integrationTestScenario := integrationTestScenario //G601
+			if !gitops.IsScenarioValid(&integrationTestScenario) {
+				a.logger.Info("IntegrationTestScenario is invalid, will not create pipelineRun for it",
+					"integrationTestScenario.Name", integrationTestScenario.Name)
+				scenarioStatusCondition := meta.FindStatusCondition(integrationTestScenario.Status.Conditions, gitops.IntegrationTestScenarioValid)
+				testStatuses.UpdateTestStatusIfChanged(
+					integrationTestScenario.Name, intgteststat.IntegrationTestStatusTestInvalid,
+					fmt.Sprintf("IntegrationTestScenario '%s' is invalid: %s", integrationTestScenario.Name, scenarioStatusCondition.Message))
+				continue
+			}
 			if shouldScenarioRunInEphemeralEnv(&integrationTestScenario) {
 				// integration pipeline runs for scenarios which need an ephemeral environment are handled in EnsureCreationOfEphemeralEnvironments()
 				a.logger.Info("IntegrationTestScenario has environment defined, skipping creation of pipelinerun.", "IntegrationTestScenario", integrationTestScenario)
@@ -234,7 +245,18 @@ func (a *Adapter) EnsureStaticIntegrationPipelineRunsExist() (controller.Operati
 			} else {
 				pipelineRun, err := a.createIntegrationPipelineRun(a.application, &integrationTestScenario, a.snapshot)
 				if err != nil {
-					return a.HandlePipelineCreationError(err, &integrationTestScenario, testStatuses)
+					if clienterrors.IsInvalid(err) {
+						a.logger.Error(err, "pipelineRun failed during creation due to invalid resource",
+							"integrationScenario.Name", integrationTestScenario.Name)
+						testStatuses.UpdateTestStatusIfChanged(
+							integrationTestScenario.Name, intgteststat.IntegrationTestStatusTestInvalid,
+							fmt.Sprintf("Creation of pipelineRun failed during creation due to invalid resource: %s.", err))
+					} else {
+						a.logger.Error(err, "Failed to create pipelineRun for snapshot and scenario",
+							"integrationScenario.Name", integrationTestScenario.Name)
+						errsForPLRCreation = errors.Join(errsForPLRCreation, err)
+					}
+					continue
 				}
 				gitops.PrepareToRegisterIntegrationPipelineRunStarted(a.snapshot) // don't count re-runs
 				testStatuses.UpdateTestStatusIfChanged(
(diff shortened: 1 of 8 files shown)


controllers/snapshot/snapshot_adapter.go
```diff
@@ -217,8 +219,18 @@ func (a *Adapter) EnsureStaticIntegrationPipelineRunsExist() (controller.Operati
 			}
 		}()
 
+		var errsForPLRCreation error
 		for _, integrationTestScenario := range *integrationTestScenarios {
 			integrationTestScenario := integrationTestScenario //G601
+			scenarioStatusCondition := meta.FindStatusCondition(integrationTestScenario.Status.Conditions, gitops.IntegrationTestScenarioValid)
+			if scenarioStatusCondition == nil || scenarioStatusCondition.Status == metav1.ConditionFalse {
```
dirgim, line 226: I'm thinking we may want to be less strict (at least for starters) and not have the `scenarioStatusCondition == nil` condition here.

controllers/snapshot/snapshot_adapter.go
```diff
@@ -217,8 +219,18 @@ func (a *Adapter) EnsureStaticIntegrationPipelineRunsExist() (controller.Operati
 			}
 		}()
 
+		var errsForPLRCreation error
 		for _, integrationTestScenario := range *integrationTestScenarios {
 			integrationTestScenario := integrationTestScenario //G601
+			scenarioStatusCondition := meta.FindStatusCondition(integrationTestScenario.Status.Conditions, gitops.IntegrationTestScenarioValid)
+			if scenarioStatusCondition.Status == metav1.ConditionFalse {
```
dirgim, line 226: Maybe we could have a new function that checks if the integrationTestScenario is valid and returns a boolean? Something like `IsScenarioValid(scenario)`?

hongweiliu17: I thought to add that function but because I need to call `scenarioStatusCondition := meta.FindStatusCondition(integrationTestScenario.Status.Conditions, gitops.IntegrationTestScenarioValid)` to write `statusCondition.Message` to snapshot annotation.
That function is added now, thanks.

kasemAlem (approved):

MartinBasti:
Why to mix refactoring and code changes in the same commit? It's not nice for review, I'd prefer in future, to do refactor in separate PR or commit.

controllers/snapshot/snapshot_adapter.go
```diff
@@ -249,7 +271,15 @@ func (a *Adapter) EnsureStaticIntegrationPipelineRunsExist() (controller.Operati
 
 		err = gitops.WriteIntegrationTestStatusesIntoSnapshot(a.snapshot, testStatuses, a.client, a.context)
 		if err != nil {
-			return controller.RequeueWithError(err)
+			a.logger.Error(err, "Failed to update test status in snapshot annotation")
+			errsForPLRCreation = errors.Join(errsForPLRCreation, err)
+		}
+
+		if errsForPLRCreation != nil {
+			if clienterrors.IsInvalid(errsForPLRCreation) {
```
MartinBasti, line 279: I'd say, we want to stop only if all errors are invalid. Because they might we a test we failed to run due to an intermittent error

hongweiliu17: Here `controller.StopProcessing()` is called when where is invalid resource error in the errors, otherwise `controller.RequeueWithError(errsForPLRCreation)` is called below. Because as long as there is an invalid resource error, manual action is needed.
Then let me exchange their order.

MartinBasti: I don't know what this fixes, it's IMO just invariant
image


What I meant, that if we cannot create a test due an intermittent issue, we won't retry.

So basically, it's still that one invalid test may block creation of others tests because we won't retry on intermittent issue.

hongweiliu17: Do you mean this PR should figure out the intermittent issue and retry for them? If yes, I think so far we can figure out all intermittent issue. The current code will retry if non-invalid error is not included in it but stop if the invalid error is included.
We may need to pay some attention to the integration log to see if there is any more intermittent issue or unrecoverable error, then add code for them.
Do you have any good idea to figure them out now?

MartinBasti: We have N tests, if 1 tests fail on invalid error, why we should stop reconciliation? There can be another failed test that could just failed to be created on intermittent error.

Also, if test is invalid, and it's only optional tests, do we want to stop reconciliation in this case?

I don't like handling of tests in general here, we should try to create all tests, record their status/failures and continue with reconciliation, user can restart tests if needed.
This retry/stop reconciliation logic seems broken to me and just add more work and uncertainty.

Also this story is spike, it was supposed to be investigation how we can handle test parallelization better, why am I reviewing PR?

hongweiliu17: > We have N tests, if 1 tests fail on invalid error, why we should stop reconciliation? There can be another failed test that could just failed to be created on intermittent error.
> 
In the updated code, if the errors doesn't include invalid error, we will retry with error.
> Also, if test is invalid, and it's only optional tests, do we want to stop reconciliation in this case?
> 
We never mentioned the optional tests in this ticket's AC. I think it is another story/spike including more code change, even about how we should handle the failed optional integration test in pr and push event.
> I don't like handling of tests in general here, we should try to create all tests, record their status/failures and continue with reconciliation, user can restart tests if needed. This retry/stop reconciliation logic seems broken to me and just add more work and uncertainty.
> 
It is trying to create all tests, record the status/failure in snapshot annotation, users can restart the test by rerun label or trigger on UI.
> Also this story is spike, it was supposed to be investigation how we can handle test parallelization better, why am I reviewing PR?
From my point of view, even for a spike, we can still gather review/comment/idea/feedback to find a good solution.

MartinBasti: > > We have N tests, if 1 tests fail on invalid error, why we should stop reconciliation? There can be another failed test that could just failed to be created on intermittent error.
> 
> In the updated code, if the errors doesn't include invalid error, we will retry with error.

That's what I mean, still if there is one test that's invalid, it prevents other tests from being retried

controllers/snapshot/snapshot_adapter.go
```diff
@@ -234,7 +245,18 @@ func (a *Adapter) EnsureStaticIntegrationPipelineRunsExist() (controller.Operati
...
 				pipelineRun, err := a.createIntegrationPipelineRun(a.application, &integrationTestScenario, a.snapshot)
 				if err != nil {
-					return a.HandlePipelineCreationError(err, &integrationTestScenario, testStatuses)
+					if clienterrors.IsInvalid(err) {
+						a.logger.Error(err, "pipelineRun failed during creation due to invalid resource",
+							"integrationScenario.Name", integrationTestScenario.Name)
+						testStatuses.UpdateTestStatusIfChanged(
+							integrationTestScenario.Name, intgteststat.IntegrationTestStatusTestInvalid,
+							fmt.Sprintf("Creation of pipelineRun failed during creation due to invalid resource: %s.", err))
+					} else {
```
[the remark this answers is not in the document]
hongweiliu17: Yes，the code should be updated. But I can't think out a good idea. Let me think
```
if !clienterrors.IsInvalid(errsForPLRCreation) {
		return controller.RequeueWithError(errsForPLRCreation)
	}
```

controllers/snapshot/snapshot_adapter.go
hongweiliu17: > > > We have N tests, if 1 tests fail on invalid error, why we should stop reconciliation? There can be another failed test that could just failed to be created on intermittent error.
> > 
> > 
> > In the updated code, if the errors doesn't include invalid error, we will retry with error.
> 
> That's what I mean, still if there is one test that's invalid, it prevents other tests from being retried

Ah, yes, `!clienterrors.IsInvalid(errsForPLRCreation)` should be updated to check none of the error is invalid error. Let me think. Thanks.

controllers/snapshot/snapshot_adapter.go
```diff
@@ -234,7 +245,18 @@ func (a *Adapter) EnsureStaticIntegrationPipelineRunsExist() (controller.Operati
...
+						a.logger.Error(err, "Failed to create pipelineRun for snapshot and scenario",
+							"integrationScenario.Name", integrationTestScenario.Name)
+					}
+					errsForPLRCreation = errors.Join(errsForPLRCreation, err)
```
dirgim, line 258: What if we do the `Join` only for errors that are not `Invalid`? 
So at the end ([here currently](https://github.com/redhat-appstudio/integration-service/pull/653/files#diff-8e718c6bf06f754db7e9e94c6f74b5c6704941dff95d637289e3c4cc194c95a0R278)) we'd only need check if the `errsForPLRCreation` is not nil and then immediately `RequeueWithError` because we'll know there are no invalid errors caught there.

hongweiliu17: Thanks. Updated.

gitops/snapshot.go
```diff
@@ -841,3 +842,29 @@ func CopySnapshotLabelsAndAnnotation(application *applicationapiv1alpha1.Applica
 	_ = metadata.CopyAnnotationsByPrefix(source, &snapshot.ObjectMeta, prefix)
 
 }
+
+// SetScenarioIntegrationStatusAsInvalid sets the IntegrationTestScenarioValid status condition for the Scenario to invalid.
+func SetScenarioIntegrationStatusAsInvalid(scenario *v1beta1.IntegrationTestScenario, message string) {
```
dirgim, line 847: While all of the functions in this file look good to me code-wise, should we place them somewhere else since they do not relate to Snapshots?
Maybe [helpers/integration.go](https://github.com/redhat-appstudio/integration-service/blob/main/helpers/integration.go) or similar?
Alternatively we could even create a new `helpers/scenario.go` for them?

controllers/snapshot/snapshot_adapter.go
dirgim: Thanks, this approach looks good to me.

gitops/snapshot.go
hongweiliu17: Yes. I thought about this, but noticed [IntegrationTestScenarioValid](https://github.com/redhat-appstudio/integration-service/blob/main/gitops/snapshot.go#L144) is in this file, then let me move out of them together.

MartinBasti (approved):

dirgim (approved):
LGTM!

PR #658, opened 2024-04-08
fix(STONEINTG-845): Snapshot's IntegrationStatus stuck on "Unknown" for optional ITS

dheerajodha:


controllers/statusreport/statusreport_adapter.go
```diff
@@ -144,16 +144,30 @@ func (a *Adapter) EnsureSnapshotFinishedAllTests() (controller.OperationResult,
...
 	if len(*integrationTestScenarios) > 0 {
-		// Since all integration tests finished, set the Snapshot Integration status as finished, but don't update the resource yet
-		gitops.SetSnapshotIntegrationStatusAsFinished(a.snapshot,
-			"Snapshot integration status condition is finished since all testing pipelines completed")
-		a.logger.LogAuditEvent("Snapshot integration status condition marked as finished, all testing pipelines completed",
-			a.snapshot, helpers.LogActionUpdate)
+		// Since all integration tests finished, set the Snapshot Integration status as finished
+		if !gitops.IsSnapshotIntegrationStatusMarkedAsFinished(a.snapshot) {
+			err = gitops.MarkSnapshotIntegrationStatusAsFinished(a.client, a.context, a.snapshot,
+				"Snapshot integration status condition is finished since all testing pipelines completed")
```
dirgim, line 150: I just realized that we have uneccessary complexity here where we do the same action when `len(*integrationTestScenarios) > 0` and in the else block, with only difference being the message. 
Can we instead set a `finishedStatusMessage` (or similar) variable based on the above condition (if there are more than 0 ITS) and then execute `gitops.MarkSnapshotIntegrationStatusAsFinished()` and `LogAuditEvent()` calls  outside of the condition?

Something like this (untested):
```
finishedStatusMessage := "Snapshot integration status condition is finished since all testing pipelines completed"
if len(*integrationTestScenarios) == 0 {
    finishedStatusMessage = "Snapshot integration status condition is finished since there are no required testing pipelines defined for its application"
}

err = gitops.MarkSnapshotIntegrationStatusAsFinished(a.client, a.context, a.snapshot, finishedStatusMessage)
if err != nil {
	a.logger.Error(err, "Failed to Update Snapshot AppStudioIntegrationStatus status")
}
a.logger.LogAuditEvent(finishedStatusMessage, a.snapshot, helpers.LogActionUpdate)
```

dirgim (approved):
LGTM

MartinBasti (approved):

PR #667, opened 2024-04-11
WIP: fix(STONEINTG-856): don't update deleted build pipelineRuns

dirgim:
* Add retryOnConflict for final updates to build pipelineRuns which covers adding create Snapshot annotation as well as removing finalizers
* Don't attempt to add the create Snapshot annotation to pipelineRuns that are being deleted


tekton/build_pipeline.go
```diff
@@ -45,6 +45,11 @@ func AnnotateBuildPipelineRun(ctx context.Context, pipelineRun *tektonv1.Pipelin
 func AnnotateBuildPipelineRunWithCreateSnapshotAnnotation(ctx context.Context, pipelineRun *tektonv1.PipelineRun, cl client.Client, ensureSnapshotExistsErr error) error {
 	message := ""
 	status := ""
+
+	if pipelineRun.GetDeletionTimestamp() != nil {
```
MartinBasti, line 49: Shouldn't be this logic at different place? When pipeline is being deleted we shouldn't even call functions that edit data.

Or is it here because of the RetryOnConflict usage?

controllers/buildpipeline/buildpipeline_adapter.go
```diff
@@ -258,10 +254,45 @@ func (a *Adapter) prepareSnapshotForPipelineRun(pipelineRun *tektonv1.PipelineRu
...
-	}
-	return err
+	return retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		var err error
+		a.pipelineRun, err = a.loader.GetPipelineRun(a.client, a.context, a.pipelineRun.Name, a.pipelineRun.Namespace)
+		if err != nil {
+			return err
+		}
+
+		err = tekton.AnnotateBuildPipelineRun(a.context, pipelineRun, tekton.SnapshotNameLabel, snapshot.Name, a.client)
```
hongweiliu17, line 264: Should we use `a.pipelineRun` here?

tekton/build_pipeline.go
dirgim: Yeah, I agree with you, was thinking to streamline the retryOnConflict, but it makes more sense to be checked before trying to annotate the build PLR.

controllers/buildpipeline/buildpipeline_adapter.go
dirgim: Ack, updated

hongweiliu17 (approved):
LGTM

Josh-Everett (approved):

PR #676, opened 2024-04-18
feat(STONEINTG-862): accidental snapshot creation when chains incomplete

Josh-Everett:
Checks that pipelinerun has an annotation verifying it has been processed by chains.


dirgim:
I think that the unit tests in [buildpipeline_adapter_test.go](https://github.com/redhat-appstudio/integration-service/blob/main/controllers/buildpipeline/buildpipeline_adapter_test.go) will need to be updated accordingly to make them work with this change.

controllers/buildpipeline/buildpipeline_adapter.go
```diff
@@ -95,6 +95,11 @@ func (a *Adapter) EnsureSnapshotExists() (result controller.OperationResult, err
 		return controller.ContinueProcessing()
 	}
 
+	if _, found := a.pipelineRun.ObjectMeta.Annotations[tekton.PipelineRunChainsSignedAnnotation]; !found {
+		a.logger.Error(err, "Unable to create Snapshot due to missing Chains annotation")
+		return controller.RequeueWithError(err)
```
dirgim, line 100: I don't think we want to call `RequeueWithError()` here because it's not an outright error if a pipelineRun is not signed - we just don't want to process such a pipelineRun, so we can log something like `Not processing the pipelineRun because it's not yet signed with Chains` and then call `controller.ContinueProcessing()` instead.

Josh-Everett: ack, updated in most recent change

MartinBasti (approved):

dirgim (approved):
LGTM

jsztuka (approved):
lgtm

PR #684, opened 2024-04-25
fix(STONEINTG-828): dont' add finalizer to deleted component

jsztuka:

We have been running into the issue that finalizers are being add to objects that are being deleted - components.
This changes should prevent that from happening by getting the object before such action.

### DIFF
--- a/controllers/component/component_adapter.go
+++ b/controllers/component/component_adapter.go
@@ -27,6 +27,7 @@ import (
 	"github.com/redhat-appstudio/integration-service/loader"
 	"github.com/redhat-appstudio/integration-service/metrics"
 	"github.com/redhat-appstudio/operator-toolkit/controller"
+	"k8s.io/apimachinery/pkg/api/errors"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
@@ -106,12 +107,25 @@ func (a *Adapter) EnsureComponentIsCleanedUp() (controller.OperationResult, erro
 			return controller.RequeueWithError(err)
 		}
 	}
+	// STONEINTG-828: We are refreshing state of component to minimize race condition with updating finalizers
+	a.component, err = a.loader.GetComponent(a.context, a.client, a.component.Name, a.component.Namespace)
+
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return controller.ContinueProcessing()
+		} else {
+			a.logger.Error(err, "Failed to get component.")
+			return controller.RequeueWithError(err)
+		}
+	}
+
 	err = h.RemoveFinalizerFromComponent(a.context, a.client, a.logger, a.component, h.ComponentFinalizer)
 	if err != nil {
 		return controller.RequeueWithError(fmt.Errorf("failed to remove the finalizer: %w", err))
 	}
 
 	return controller.ContinueProcessing()
+
 }
 
 // createUpdatedSnapshot prepares a Snapshot for a given application and component(s).
--- a/controllers/component/component_controller_test.go
+++ b/controllers/component/component_controller_test.go
@@ -1,8 +1,9 @@
 package component
 
 import (
-	"k8s.io/apimachinery/pkg/api/errors"
 	"reflect"
+
+	"k8s.io/apimachinery/pkg/api/errors"
 	"sigs.k8s.io/controller-runtime/pkg/metrics/server"
 	crwebhook "sigs.k8s.io/controller-runtime/pkg/webhook"
 
@@ -22,7 +23,7 @@ import (
 	klog "k8s.io/klog/v2"
 )
 
-var _ = Describe("ComponentController", func() {
+var _ = Describe("ComponentController", Ordered, func() {
 	var (
 		manager             ctrl.Manager
 		componentReconciler *Reconciler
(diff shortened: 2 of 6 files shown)


loader/loader_mock.go
```diff
@@ -262,3 +263,11 @@ func (l *mockLoader) GetPipelineRun(c client.Client, ctx context.Context, name,
 	}
 	return toolkit.GetMockedResourceAndErrorFromContext(ctx, GetPipelineRunContextKey, &tektonv1.PipelineRun{})
 }
+
+// GetComponent returns the resource and error passed as values of the context.
+func (l *mockLoader) GetComponent(c client.Client, ctx context.Context, name, namespace string) (*applicationapiv1alpha1.Component, error) {
```
dirgim, line 268: Note: this function will need unit test coverage in [loader_mock_test.go](https://github.com/redhat-appstudio/integration-service/blob/main/loader/loader_mock_test.go)

controllers/component/component_adapter.go
```diff
@@ -105,6 +105,7 @@ func (a *Adapter) EnsureComponentIsCleanedUp() (controller.OperationResult, erro
 			return controller.RequeueWithError(err)
 		}
 	}
+	a.component, err = a.loader.GetComponent(a.client, a.context, a.component.Name, a.component.Namespace)
```
dirgim, line 108: I think we'll need to check the `err` after this call to see if we managed to get the component successfully.

controllers/component/component_adapter.go
```diff
@@ -105,6 +105,11 @@ func (a *Adapter) EnsureComponentIsCleanedUp() (controller.OperationResult, erro
...
+	if err != nil {
+		a.logger.Error(err, "Failed to get component.")
+		return controller.RequeueWithError(err)
```
dirgim, line 111: Might be a good idea to continue processing if the error is `NotFound` - see [here](https://github.com/redhat-appstudio/integration-service/blob/a28863928205fed45ddf7d952466dd9a70387df3/controllers/buildpipeline/buildpipeline_adapter.go#L74-L80) for example.

jsztuka: Good idea, updated.

jsztuka: Hmm I cutoff the remove finalizer with this logic, I need to move that removal probably in the if condition.

controllers/component/component_adapter.go
```diff
@@ -105,12 +106,17 @@ func (a *Adapter) EnsureComponentIsCleanedUp() (controller.OperationResult, erro
 			return controller.RequeueWithError(err)
 		}
 	}
-	err = h.RemoveFinalizerFromComponent(a.client, a.logger, a.context, a.component, h.ComponentFinalizer)
-	if err != nil {
-		return controller.RequeueWithError(fmt.Errorf("failed to remove the finalizer: %w", err))
+	a.component, err = a.loader.GetComponent(a.client, a.context, a.component.Name, a.component.Namespace)
+	if errors.IsNotFound(err) {
```
MartinBasti, line 110: I don't understand this. if component doesn't exist, how can we remove finalizer?

controllers/component/component_adapter.go
```diff
@@ -105,12 +106,17 @@ func (a *Adapter) EnsureComponentIsCleanedUp() (controller.OperationResult, erro
...
+		err = h.RemoveFinalizerFromComponent(a.client, a.logger, a.context, a.component, h.ComponentFinalizer)
```
MartinBasti, line 111: what's the value of `a.component` when there is not found error? Nil or empty?

controllers/component/component_adapter.go
jsztuka: Should not this be other way around, if error Is**Not**Found, meaning we could get component and we can remove finalizer in that case.

jsztuka: I have probably misundersoot using errors.IsNotFound(err)
The way I thaught of it:
1. we try to get component
2. if there is no error > remove finalizer
3. in case of error getting component > inform about such fact and requeue with error
4. In case of error removing the finalizer do the same as in 3.

jsztuka: I changed the logic, let me know if its correct this itme.

controllers/component/component_adapter.go
jsztuka: Updated the logic, let me know if its correct now.

controllers/component/component_controller_test.go
```diff
@@ -22,7 +23,7 @@ import (
 	klog "k8s.io/klog/v2"
 )
 
-var _ = Describe("ComponentController", func() {
+var _ = Describe("ComponentController", Ordered, func() {
```
MartinBasti, line 26: Why do we need this change to ordered tests?

controllers/component/component_adapter.go
```diff
@@ -105,12 +106,24 @@ func (a *Adapter) EnsureComponentIsCleanedUp() (controller.OperationResult, erro
 			return controller.RequeueWithError(err)
 		}
 	}
+	a.component, err = a.loader.GetComponent(a.client, a.context, a.component.Name, a.component.Namespace)
```
MartinBasti, line 109: Maybe add a comment before, that we are doing it to refresh state of component to minimize race condition with updating finalizers

controllers/component/component_controller_test.go
jsztuka: Unit test won't pass if its not ordered, same issue as we had in 841 with removing resoures.

controllers/component/component_adapter.go
jsztuka: Agreed.

MartinBasti (approved):

dirgim (approved):
LGTM

PR #715, opened 2024-05-09
fix(STONE-828): get component before add finalizer

jsztuka:

Since the issue is still happening (less but still), I think it could help to get the component even before we try to add finalizer to the component itself.


dirgim (approved):
LGTM code-wise

MartinBasti:
What is this actually fixing? All the errors we see are from removal, not from adding. This is first adapter call, I don't think that there can be something to refresh yet.
```
error occurred while patching the updated Component after finalizer removal
```

jsztuka:
I agree with Martins last comment, however how about we make sure we are getting component in correct state even in **component_controller.go** the similar way we are doing it in adapter?

jsztuka:
Closing this PR with following:
Controller is the first one to work with component object, hence the refresh of the object won't help us solve this issue.

PR #736, opened 2024-05-20
fix(STONE-876): catch admission validation error

jsztuka:
Goal is to prevent reconcialtion being in infinite look retrying to create pipelinerun. We will catch this specific action when the webhook validation failed and report test status as invalid.

### DIFF
--- a/internal/controller/snapshot/snapshot_adapter.go
+++ b/internal/controller/snapshot/snapshot_adapter.go
@@ -522,9 +522,14 @@ func (a *Adapter) HandlePipelineCreationError(err error, integrationTestScenario
 		return controller.RequeueWithError(itsErr)
 	}
 
+	if strings.Contains(err.Error(), "admission webhook") && strings.Contains(err.Error(), "denied the request") {
+		//Stop processing in case the error runs in admission webhook validation error:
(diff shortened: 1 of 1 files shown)


jsztuka:
Part of this story was including unit tests covering this behavior. Turned out that admission validation webhook error is generic error withou status or code. That means following: error won't show up in unit tests. Since there is currently no obvious simple way how to mock such situation, it could be potential for upcomming: https://issues.redhat.com/browse/STONEINTG-690

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -522,9 +522,14 @@ func (a *Adapter) HandlePipelineCreationError(err error, integrationTestScenario
 		return controller.RequeueWithError(itsErr)
 	}
 
+	if strings.Contains(err.Error(), "admission webhook") {
```
dirgim, line 525: Maybe we could expand this condition to also require `denied the request` to be present in the error message?

jsztuka: Searched for elegant way how to check for multiple strings. None of them were elegant.

dirgim (approved):
LGTM

PR #760, opened 2024-06-03
feat(STONEINTG-892): support new override snapshot

hongweiliu17:
* introduce a new type of snapshot: override
* update global candidate list if a override snapshot contains the component with valid digest

- [x] [Controllers diagrams](https://github.com/konflux-ci/integration-service/tree/main/docs) are updated when PR changes controllers code  (if applicable)

### DIFF
--- a/gitops/snapshot.go
+++ b/gitops/snapshot.go
@@ -86,6 +86,9 @@ const (
 	// SnapshotCompositeType is the type of Snapshot which was created for multiple components.
 	SnapshotCompositeType = "composite"
 
+	// SnapshotOverrideType is the type of Snapshot which was created for override Global Candidate List.
+	SnapshotOverrideType = "override"
+
 	// PipelineAsCodeEventTypeLabel is the type of event which triggered the pipelinerun in build service
 	PipelineAsCodeEventTypeLabel = PipelinesAsCodePrefix + "/event-type"
 
@@ -842,4 +845,17 @@ func CopySnapshotLabelsAndAnnotation(application *applicationapiv1alpha1.Applica
 	_ = metadata.CopyLabelsByPrefix(source, &snapshot.ObjectMeta, prefix)
 	_ = metadata.CopyAnnotationsByPrefix(source, &snapshot.ObjectMeta, prefix)
 
+}
+
+// IsOverrideSnapshot returns true if snapshot label 'test.appstudio.openshift.io/type' is 'override'
+func IsOverrideSnapshot(snapshot *applicationapiv1alpha1.Snapshot) bool {
+	return metadata.HasLabelWithValue(snapshot, SnapshotTypeLabel, SnapshotOverrideType)
+}
+
+func IsComponentSnapshot(snapshot *applicationapiv1alpha1.Snapshot) bool {
+	return metadata.HasLabelWithValue(snapshot, SnapshotTypeLabel, SnapshotComponentType)
+}
+
+func IsComponentSnapshotCreatedByPACPushEvent(snapshot *applicationapiv1alpha1.Snapshot) bool {
+	return IsComponentSnapshot(snapshot) && IsSnapshotCreatedByPACPushEvent(snapshot)
 }
--- a/internal/controller/snapshot/snapshot_adapter.go
+++ b/internal/controller/snapshot/snapshot_adapter.go
@@ -279,10 +279,11 @@ func (a *Adapter) EnsureIntegrationPipelineRunsExist() (controller.OperationResu
 // EnsureGlobalCandidateImageUpdated is an operation that ensure the ContainerImage in the Global Candidate List
 // being updated when the Snapshot is created
 func (a *Adapter) EnsureGlobalCandidateImageUpdated() (controller.OperationResult, error) {
-	if a.snapshot.GetLabels()[gitops.SnapshotTypeLabel] != gitops.SnapshotComponentType || !gitops.IsSnapshotCreatedByPACPushEvent(a.snapshot) {
-		a.logger.Info("The Snapshot wasn't created for a single component push event, not updating the global candidate list.")
+	if !gitops.IsComponentSnapshotCreatedByPACPushEvent(a.snapshot) && !gitops.IsOverrideSnapshot(a.snapshot) {
+		a.logger.Info("The Snapshot was neither created for a single component push event nor override type, not updating the global candidate list.")
 		return controller.ContinueProcessing()
 	}
+
 	if gitops.IsSnapshotMarkedAsAddedToGlobalCandidateList(a.snapshot) {
 		a.logger.Info("The Snapshot's component was previously added to the global candidate list, skipping adding it.")
 		return controller.ContinueProcessing()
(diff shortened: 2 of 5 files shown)


internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -280,48 +280,61 @@ func (a *Adapter) EnsureIntegrationPipelineRunsExist() (controller.OperationResu
...
+		if err != nil {
+			_, loaderError := h.HandleLoaderError(a.logger, err, fmt.Sprintf("Failed to get component %s from applicaion, won't update global candidate list for this component", snapshotComponent.Name), a.application.Name)
+			if loaderError != nil {
+				return controller.RequeueWithError(loaderError)
+			}
+			continue
+		}
+
+		patch := client.MergeFrom(component.DeepCopy())
+		component.Spec.ContainerImage = snapshotComponent.ContainerImage
```
dirgim, line 310: It seems like this change updates all components for both the `component` and `override` Snapshots.

I think we should update the `spec.containerImage` only for the Snapshot's component in case of the `component` Snapshot.

hongweiliu17: Thanks. Updated. I thought `override` snapshot will not have `.Status.LastBuiltCommit`. I add code `if !gitops.IsOverrideSnapshot(a.snapshot) {` to only update `containerImage`  for the component snapshot. Thanks.

gitops/snapshot.go
```diff
@@ -843,3 +846,9 @@ func CopySnapshotLabelsAndAnnotation(application *applicationapiv1alpha1.Applica
 	_ = metadata.CopyAnnotationsByPrefix(source, &snapshot.ObjectMeta, prefix)
 
 }
+
+// IsOverrideSnapshot return true if this snapshot's is override 'test.appstudio.openshift.io/type' is override
```
jsztuka, line 850:
```suggestion
// IsOverrideSnapshot returns true if snapshot label 'test.appstudio.openshift.io/type' is override
```

hongweiliu17: Updated. Thanks.

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -279,62 +278,68 @@ func (a *Adapter) EnsureIntegrationPipelineRunsExist() (controller.OperationResu
...
-		a.logger.Info("The Snapshot wasn't created for a single component push event, not updating the global candidate list.")
+	if (a.snapshot.GetLabels()[gitops.SnapshotTypeLabel] != gitops.SnapshotComponentType ||
+		!gitops.IsSnapshotCreatedByPACPushEvent(a.snapshot)) && !gitops.IsOverrideSnapshot(a.snapshot) {
+		a.logger.Info("The Snapshot was neither created for a single component push event nor override type, not updating the global candidate list.")
 		return controller.ContinueProcessing()
 	}
-	if gitops.IsSnapshotMarkedAsAddedToGlobalCandidateList(a.snapshot) {
-		a.logger.Info("The Snapshot's component was previously added to the global candidate list, skipping adding it.")
+
+	if gitops.IsOverrideSnapshot(a.snapshot) && !gitops.HaveAppStudioTestsSucceeded(a.snapshot) {
```
dirgim, line 287: After giving this some thought, I think we should promote the override Snapshot component images to GCL immediately and not wait to check if it passes tests. This should keep commonality with the way that the component Snapshots are handled.

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -279,62 +278,68 @@ func (a *Adapter) EnsureIntegrationPipelineRunsExist() (controller.OperationResu
...
+		a.logger.Info("This is a override snapshot but hasn't yet passed all required integration tests, not updating the global candidate list.")
 		return controller.ContinueProcessing()
 	}
 
-	var componentToUpdate *applicationapiv1alpha1.Component
-	var err error
-	err = retry.OnError(retry.DefaultRetry, func(_ error) bool { return true }, func() error {
-		componentToUpdate, err = a.loader.GetComponentFromSnapshot(a.context, a.client, a.snapshot)
```
dirgim, line 294: I think we'll still need to get the component from Snapshot in case of `component` type Snapshots because for those kinds of Snapshots we only want to update that single component and leave the others as-is.

internal/controller/snapshot/snapshot_adapter.go
dirgim: If I'm reading this right, the current code still updates `spec.containerImage` fields for all components, which I still think should happen only in case of `override` Snapshots.

hongweiliu17: But in our old `component` snapshot GCL, we need to update the component's `spec.containerImage` [here](https://github.com/konflux-ci/integration-service/blob/main/internal/controller/snapshot/snapshot_adapter.go#L294-L309), I feel we should cover it, right?

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -279,62 +278,68 @@ func (a *Adapter) EnsureIntegrationPipelineRunsExist() (controller.OperationResu
...
-		_, loaderError := h.HandleLoaderError(a.logger, err, fmt.Sprintf("Component or '%s' label", tekton.ComponentNameLabel), "Snapshot")
-		if loaderError != nil {
-			return controller.RequeueWithError(loaderError)
-		}
+	if gitops.IsSnapshotMarkedAsAddedToGlobalCandidateList(a.snapshot) {
+		a.logger.Info("The Snapshot's component was previously added to the global candidate list, skipping adding it.")
 		return controller.ContinueProcessing()
 	}
 
 	for _, snapshotComponent := range a.snapshot.Spec.Components {
```
dirgim, line 297: To clarify, a `component` Snapshot will still have all of the application's components listed in the `snapshot.Spec.Components` field. 
Because of that, we will still need to read the Snapshot's component label ([appstudio.openshift.io/component](https://github.com/konflux-ci/integration-service/blob/80bf5eb5703d2d62e9b945db3c470c3fe506e5eb/tekton/utils.go#L40)) to figure out which specific component we wish to promote to GCL.

internal/controller/snapshot/snapshot_adapter.go
hongweiliu17: Removed the integration test check, but [STONEINTG-83](https://issues.redhat.com/browse/STONEINTG-83) mentions:
```
Allow users to create manual ‘override’ Snapshots
    This would provide a mitigation mechanism for cases when the entire application is blocked from releasing because of a single bad component build
```
I felt it is to make the entire application pass integration test.

internal/controller/snapshot/snapshot_adapter.go
hongweiliu17: Updated. Thanks.

internal/controller/snapshot/snapshot_adapter.go
hongweiliu17: Updated. Thanks for the clarification.

internal/controller/snapshot/snapshot_adapter.go
dirgim: Yup I see your point but I think this will prevent confusion on the user's side because now both types of Snapshot will be treated the same.
We can  change this in the future based on the user's feedback, but I think this is the way to go for now.
Thanks for making the update!

internal/controller/snapshot/snapshot_adapter.go
dirgim: I think the latest changes work well for both cases, thanks!

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -279,62 +279,80 @@ func (a *Adapter) EnsureIntegrationPipelineRunsExist() (controller.OperationResu
...
-					componentToUpdate, h.LogActionUpdate,
-					"lastBuildCommit", componentToUpdate.Status.LastBuiltCommit)
+				// Updated the component for component snapshot, break
+				break
+			} else {
+				// expected component not found, continue
+				continue
+			}
+		}
+	} else {
```
dirgim, line 331: Maybe we can add a check to ensure that we only update all components for override Snapshots just in case?
Something like:
```
} else if gitops.IsOverrideSnapshot(a.snapshot) {
```

internal/controller/snapshot/snapshot_adapter.go
hongweiliu17: Agree. Thanks.

internal/controller/snapshot/snapshot_adapter.go
hongweiliu17: Updated. Thanks.

dirgim (approved):
LGTM

Josh-Everett (approved):
nice patch

PR #770, opened 2024-06-12
fix(STONEINTG-939): update checkruns correctly

dheerajodha:
Before this commit:
* When we were updating a pre-existing checkrun,
   the correct status wouldn't get reflected on the
   GH until the test is finished.
* This is expected behaviour as per GH: An already
   "completed" checkrun should NOT be updated to
   a non-completed status.
* Instead, we should create another checkrun with
   the same external ID and that would reset the
   already completed the checkrun correctly.
* Source: "https://github.com/isaacs/github/issues/1530#issuecomment-599958904"

After this commit:
* The users can see the correct status AND total
   duration of Checkrun, while retriggering their tests.

### DIFF
--- a/status/reporter_github.go
+++ b/status/reporter_github.go
@@ -238,19 +238,32 @@ func (cru *CheckRunStatusUpdater) UpdateStatus(ctx context.Context, report TestR
 		if err != nil {
 			cru.logger.Error(err, "failed to create checkrun",
 				"checkRunAdapter", checkRunAdapter)
-			return err
 		}
-	} else {
-		cru.logger.Info("found existing checkrun", "existingCheckRun", existingCheckrun)
-		err = cru.ghClient.UpdateCheckRun(ctx, *existingCheckrun.ID, checkRunAdapter)
-		if err != nil {
-			cru.logger.Error(err, "failed to update checkrun",
-				"checkRunAdapter", checkRunAdapter)
-			return err
-		}
-
+		return err
 	}
-	return nil
+
+	cru.logger.Info("found existing checkrun", "existingCheckRun", existingCheckrun)
+
+	// If pre-existing checkrun is already completed, then create a
+	// new checkrun with same external ID, rather than updating it
+	if existingCheckrun.GetStatus() == "completed" {
+		cru.logger.Info("The existing checkrun is already in completed state, re-creating a new checkrun for scenario test status of snapshot",
+			"snapshot.NameSpace", cru.snapshot.Namespace, "snapshot.Name", cru.snapshot.Name, "scenarioName", report.ScenarioName)
+		_, err = cru.ghClient.CreateCheckRun(ctx, checkRunAdapter)
+		if err != nil {
+			cru.logger.Error(err, "failed to create checkrun",
+				"checkRunAdapter", checkRunAdapter)
+		}
+		return err
+	}
+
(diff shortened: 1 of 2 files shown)


dheerajodha:
I think the e2e-test failed while deleting build plr that failed with:
```
"message": "Pipeline integration2-izyb-tenant/ can't be Run; it contains Tasks that don't exist: Couldn't retrieve Task \"\": retryable error validating referenced object buildah: Timeout: request did not complete within requested timeout - context deadline exceeded",
```
looks flaky
/test integration-service-e2e

dheerajodha:
<details>
<summary>CLICK HERE TO VIEW THE BEFORE & AFTER SCREENSHOTS</summary>

**BEFORE**:
image

(Notice the wrong status and current duration while the test is running)

image
(Notice the wrong total duration that should've been ~5 mins)

---------------------------------------------
**AFTER**:
image
(Notice the correct status and current duration)
image
(Notice the correct duration of ~5 mins)

</details>

hongweiliu17 (commented):
Thanks for the patch, it looks good to me! Some more test for the code will be better.

kasemAlem (approved):

dheerajodha:
> Thanks for the patch, it looks good to me! Some more test for the code will be better.

@hongweiliu17 Ack! I added the test coverage now, PTAL.
Thanks!

hongweiliu17 (approved):
LGTM

status/reporter_github.go
```diff
@@ -238,19 +238,32 @@ func (cru *CheckRunStatusUpdater) UpdateStatus(ctx context.Context, report TestR
...
+			cru.logger.Error(err, "failed to create checkrun",
 				"checkRunAdapter", checkRunAdapter)
-			return err
 		}
+		return err
+	}
 
+	err = cru.ghClient.UpdateCheckRun(ctx, *existingCheckrun.ID, checkRunAdapter)
+	if err != nil {
+		cru.logger.Error(err, "failed to update checkrun",
```
jsztuka, line 262: Would it be more convenient to add existingCheckrun.ID also into the error message to better traceability? Feel free to prove me wrong, I don't know what checkrunAdapter information consist of.

dheerajodha: Hmm good point, I think that since the checkrun's ID field is just a bunch of numbers (example: `ID:26087296145`) it's not very helpful while debugging. We can instead use ExternalID field as that is a combination of scenario-name + component-name and it is unique for each checkrun. It's already a part of `checkrunAdapter` ([here](https://github.com/konflux-ci/integration-service/blob/be17a66db2d4ac7fc7ed676b35158a8bd8ccc4e8/git/github/github.go#L38)), that gets printed within the above error message.

Also, we're printing the ExternalID again [here](https://github.com/konflux-ci/integration-service/blob/be17a66db2d4ac7fc7ed676b35158a8bd8ccc4e8/git/github/github.go#L381), and then the entire existing checkrun object [here](https://github.com/konflux-ci/integration-service/pull/770/files#diff-83bd1d9146518ccfb3036c984464501bca2c88acc23ffcd801c9057e7c0df48eR245). So, we'd be able to compare both existing checkrun and the checkrun we're trying to create/update, in any case.

dirgim: I think that the `externalID` is the main info that we use for debugging so I think that's enough for our needs.

dirgim (approved):
LGTM

PR #774, opened 2024-06-17
fix(KFLUXBUGS-1355): skip components with invalid images

dirgim:
* Don't return an error when constructing a Snapshot in case one of the existing (non-built) components has an invalid image reference in their GCL (spec.containerImage)
* Do return an error in case the newly built image is invalid


Josh-Everett (approved):

dirgim:
We've gotten confirmation that Snapshots can be partially released and users can decide if the Snapshot is releasable based on their Release Plan.

PR #781, opened 2024-06-21
fix(STONEINTG-960): show conditions in webUI

jsztuka:
This PR prevents reporting empty string to webUI when PLR finishes. In case that TEST_OUTPUT is nil, we check status conditions for given taskrun and report it.

### DIFF
--- a/status/format.go
+++ b/status/format.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/go-logr/logr"
 	"github.com/konflux-ci/integration-service/helpers"
+	"knative.dev/pkg/apis"
 )
 
 const commentTemplate = `### {{ .Title }}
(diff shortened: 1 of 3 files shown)


status/format.go
```diff
@@ -170,19 +178,23 @@ func FormatDetails(taskRun *helpers.TaskRun) (string, error) {
 	}
 
 	if result == nil {
-		return "", nil
+		var emoji string
+		taskSucceededReason := taskRun.GetStatusCondition("Succeeded").GetReason()
```
dirgim, line 182: Suggestion: maybe we can use the `apis.ConditionSucceeded` constant instead of the hardcoded `"Succeeded"` string?

status/format.go
```diff
@@ -249,3 +261,17 @@ func FormatTaskLogURL(taskRun *helpers.TaskRun, pipelinerun string, namespace st
 	}
 	return buf.String()
 }
+
+func EvaluateReason(taskRun *helpers.TaskRun) (string, error) {
```
dirgim, line 265: I may be missing something, but I'm not seeing where this function is used?

jsztuka: You are not missing anything, actually I missed this one. Interesting that CI didn't catch that.

status/format.go
jsztuka: Added the recommended conditionSucceeded, let me know if that fulfills your requirements.

dirgim (approved):
LGTM

status/format.go
```diff
@@ -170,19 +179,23 @@ func FormatDetails(taskRun *helpers.TaskRun) (string, error) {
...
+			emoji = ":question:"
+		}
+		return fmt.Sprintf(emoji+" Reason: %s", taskSucceededReason), nil
 	}
 
 	if result.ValidationError != nil {
 		return fmt.Sprintf("Invalid result: %s", result.ValidationError), nil
 	}
-
-	if result.TestOutput == nil {
```
hongweiliu17, line 180: I am just curious about this removal. Will this condition not occurs?

jsztuka: So I need to add clarification here, this condition is true also when `result.ValidationError != nil`, meaning it will never reach it, since its covered by previous condition.

kasemAlem (approved):

PR #799, opened 2024-07-04
feat(STONEINTG-966): ensure override snapshot is valid to be released

hongweiliu17:
* Validate Override snapshot to release it and mark it as invalid if it doesn follow:
      the snapshotComponent should exist as component
      containerImage in snapshotComponent has valid digest
      snapshotComponent has git source defined
    

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -433,6 +435,73 @@ func (a *Adapter) EnsureAllReleasesExist() (controller.OperationResult, error) {
...
+	}
+
+	var err error
+	if !controllerutil.HasControllerReference(a.snapshot) {
+		a.snapshot, err = gitops.SetOwnerReference(a.context, a.client, a.snapshot, a.application)
+		if err != nil {
+			a.logger.Error(err, fmt.Sprintf("Failed to set owner reference for snapshot %s/%s", a.snapshot.Namespace, a.snapshot.Name))
+			return controller.RequeueWithError(err)
+		}
+		a.logger.Info("Owner reference has been set to snapshot")
```
jsztuka, line 458: Just a nit, if we can also log a.snapshot.Name here for example. Let me know if you think that this information (name) would be helpful.

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -433,6 +435,73 @@ func (a *Adapter) EnsureAllReleasesExist() (controller.OperationResult, error) {
...
+		if !gitops.HaveGitSource(snapshotComponent) {
+			a.logger.Error(err, "snapshotComponent has no git url/revision", "snapshotComponent.Name", snapshotComponent.Name, "snapshotComponent.ContainerImage", snapshotComponent.ContainerImage)
+			errsForSnapshot = errors.Join(errsForSnapshot, err)
+			continue
+		}
+	}
+
+	if errsForSnapshot != nil {
+		a.logger.Error(errsForSnapshot, "mark the override snapshot as invalid due to invalid containerImage or git source")
+		err = gitops.MarkSnapshotAsInvalid(a.context, a.client, a.snapshot, errsForSnapshot.Error())
```
jsztuka, line 493: I suppose override snapshot is set as valid by default? I could not find section where we set it to valid.
Or I presume valid is not set at all, just setting invalid in cases we found errors?

internal/controller/snapshot/snapshot_adapter.go
hongweiliu17: I feel this [logger](https://github.com/konflux-ci/integration-service/blob/main/internal/controller/snapshot/snapshot_controller.go#L74) will log with snapshot name, right?

internal/controller/snapshot/snapshot_adapter.go
hongweiliu17: In this [IsSnapshotValid()](https://github.com/konflux-ci/integration-service/blob/main/gitops/snapshot.go#L389), I think if it is not been set to invalid, it will be considered as `valid`, right?

jsztuka (approved):
lgtm

Josh-Everett (approved):

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -433,6 +435,73 @@ func (a *Adapter) EnsureAllReleasesExist() (controller.OperationResult, error) {
...
+				return controller.RequeueWithError(loaderError)
+			} else {
+				errsForSnapshot = errors.Join(errsForSnapshot, fmt.Errorf("snapshotComponent %s defined in snapshot %s doesn't exist under application %s/%s", snapshotComponent.Name, a.snapshot.Name, a.application.Namespace, a.application.Name))
+				continue
+			}
+		}
+
+		err = gitops.ValidateImageDigest(snapshotComponent.ContainerImage)
+		if err != nil {
+			a.logger.Error(err, "containerImage in snapshotComponent has no valid digest", "snapshotComponent.Name", snapshotComponent.Name, "snapshotComponent.ContainerImage", snapshotComponent.ContainerImage)
```
dheerajodha, line 479: nit
```suggestion
			a.logger.Error(err, "containerImage in snapshotComponent has invalid digest", "snapshotComponent.Name", snapshotComponent.Name, "snapshotComponent.ContainerImage", snapshotComponent.ContainerImage)
```

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -433,6 +435,73 @@ func (a *Adapter) EnsureAllReleasesExist() (controller.OperationResult, error) {
...
+			errsForSnapshot = errors.Join(errsForSnapshot, err)
+			continue
+		}
+
+		if !gitops.HaveGitSource(snapshotComponent) {
+			a.logger.Error(err, "snapshotComponent has no git url/revision", "snapshotComponent.Name", snapshotComponent.Name, "snapshotComponent.ContainerImage", snapshotComponent.ContainerImage)
```
dheerajodha, line 485: nit:
```suggestion
			a.logger.Error(err, "snapshotComponent has no git url/revision fields defined", "snapshotComponent.Name", snapshotComponent.Name, "snapshotComponent.ContainerImage", snapshotComponent.ContainerImage)
```

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -433,6 +435,73 @@ func (a *Adapter) EnsureAllReleasesExist() (controller.OperationResult, error) {
...
+			errsForSnapshot = errors.Join(errsForSnapshot, err)
+			continue
+		}
+	}
+
+	if errsForSnapshot != nil {
+		a.logger.Error(errsForSnapshot, "mark the override snapshot as invalid due to invalid containerImage or git source")
```
dheerajodha, line 492: This could be misleading as the snapshot could be invalid due to absence of snapshotComponent, so we could either mention all the possible reasons like this:
```suggestion
		a.logger.Error(errsForSnapshot, "mark the override snapshot as invalid due to non-existent components or git source or due invalid containerImage defined within it")
```
or mention none of them since they will end up printed by the `errsForSnapshot` field:
```suggestion
		a.logger.Error(errsForSnapshot, "mark the override snapshot as invalid")
```
wdyt?

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -433,6 +435,73 @@ func (a *Adapter) EnsureAllReleasesExist() (controller.OperationResult, error) {
...
+		err = gitops.MarkSnapshotAsInvalid(a.context, a.client, a.snapshot, errsForSnapshot.Error())
+		if err != nil {
+			a.logger.Error(err, "Failed to update snapshot to Invalid",
+				"snapshot.Namespace", a.snapshot.Namespace, "snapshot.Name", a.snapshot.Name)
+			return controller.RequeueWithError(err)
+		}
+		a.logger.Info("Snapshot has been marked as invalid due to invalid containerImage or git source",
```
dheerajodha, line 499: same as above: either mention all the possible reasons we could've marked the snapshot as invalid, or mention none at all. Wdyt?

internal/controller/snapshot/snapshot_controller.go
```diff
@@ -121,6 +121,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		adapter.EnsureGlobalCandidateImageUpdated,
 		adapter.EnsureRerunPipelineRunsExist,
 		adapter.EnsureIntegrationPipelineRunsExist,
+		adapter.EnsureRerunPipelineRunsExist,
```
dheerajodha, line 124: a typo
```suggestion
		adapter.EnsureOverrideSnapshotValid,
```

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -433,6 +435,73 @@ func (a *Adapter) EnsureAllReleasesExist() (controller.OperationResult, error) {
...
+		snapshotComponent := snapshotComponent //G601
+		_, err := a.loader.GetComponent(a.context, a.client, snapshotComponent.Name, a.snapshot.Namespace)
+		if err != nil {
+			a.logger.Error(err, "Failed to get component from applicaion", "application.Name", a.application.Name, "component.Name", snapshotComponent.Name)
+			_, loaderError := h.HandleLoaderError(a.logger, err, snapshotComponent.Name, a.application.Name)
+			if loaderError != nil {
+				return controller.RequeueWithError(loaderError)
+			} else {
+				errsForSnapshot = errors.Join(errsForSnapshot, fmt.Errorf("snapshotComponent %s defined in snapshot %s doesn't exist under application %s/%s", snapshotComponent.Name, a.snapshot.Name, a.application.Namespace, a.application.Name))
+				continue
```
dheerajodha, line 473: Why are we skipping the rest of the for loop after we find the first place of error?

Due to this, while debugging, we would not know all the reasons why a snapshot is invalid. I think this is how it would go:
* We see that Component A is missing and this is why we have marked the override snapshot as invalid
* We would notify the user that this is the reason, and they would fix it.
* Then we will go back in this for loop, but now we are failing in the lines below this, while trying to validate the image. So we'll repeat the previous step again.

We can skip this back and forth by just capturing all the errors at once and recording & logging them. Wdyt?

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -433,6 +435,73 @@ func (a *Adapter) EnsureAllReleasesExist() (controller.OperationResult, error) {
...
+func (a *Adapter) EnsureOverrideSnapshotValid() (controller.OperationResult, error) {
+	if !gitops.IsOverrideSnapshot(a.snapshot) {
+		a.logger.Info("The snapshot was not override snapshot, skipping")
+		return controller.ContinueProcessing()
+	}
+
+	if gitops.IsSnapshotMarkedAsInvalid(a.snapshot) {
+		a.logger.Info("The override snapshot has been marked as invalid, skipping")
+		return controller.ContinueProcessing()
+	}
```
dheerajodha, line 449: Let's say:
* A user created an override snapshot with an invalid digest, so we mark it as invalid (as per the logic defined in the for loop somewhere below this LOC)
* Now they realized the reason behind why it is invalid, so they fixed the digest.
* But now this Snapshot is stuck at Invalid state because we call `return controller.ContinueProcessing()` here.

Should we reconsider this `if` condition?

docs/snapshot-controller.md
```diff
@@ -98,6 +98,22 @@ flowchart TD
   rerun_static_env                ---->    remove_rerun_label
 
 
+  %%%%%%%%%%%%%%%%%%%%%%% Drawing EnsureOverrideSnapshotValid() function
+
+  %% Node definitions
+  ensure4(Process further if: Snapshot has override type label)
+  validate_override_valid{Check if override snapshot has <br>valid image digest and git source}
```
dheerajodha, line 105: this should be phrased as a question I think
```suggestion
  validate_override_valid{Is the override snapshot <br>defined with valid components, <br>image digest, and git source?}
```

docs/snapshot-controller.md
```diff
@@ -98,6 +98,22 @@ flowchart TD
...
+  mark_snapshot_invalid(<b>Mark</b> overide snapshot as invalid)
```
dheerajodha, line 106: nit:
```suggestion
  mark_snapshot_invalid(<b>Mark</b> override snapshot as invalid)
```

internal/controller/snapshot/snapshot_adapter.go
hongweiliu17: This is just `continue` the rest of loop, not `break`.

dheerajodha: Yes, it's `continue`, that's my point because let's say that a snapshotComponent is missing the component, valid digest, and git source too. Then based on the current logic, `errsForSnapshot` will ONLY contain the **first** error message, and **skip** through the rest of the error messages.

internal/controller/snapshot/snapshot_adapter.go
hongweiliu17: Thanks. Let's change to `Snapshot has been marked as invalid due to invalid snapshotComponent`

internal/controller/snapshot/snapshot_adapter.go
hongweiliu17: Good question. The snapshotComponents in other type of snapshot is immutable. For this type of `override` snapshot, let me think.

internal/controller/snapshot/snapshot_adapter.go
dheerajodha: Nice, that makes sense to me.

internal/controller/snapshot/snapshot_adapter.go
hongweiliu17: I think if a component doesn't exist, then checking its source doesn't make sense...

internal/controller/snapshot/snapshot_adapter.go
hongweiliu17: For snapshot, we reconcile it when it is created or its annotation is updated, if users wants to reconcile it after updating it, they have to create a new snapshot.

internal/controller/snapshot/snapshot_adapter.go
dheerajodha: Hmm, I still think it makes sense to show all the errors to the users in one go.
But I need your help to figure out if I'm thinking correctly here: Let's say I'm a user who doesn't know how to create override snapshots and that's why its definition is all wrong (wrong component, invalid digest, missing git source), so I, as a user, would prefer to know everything that I've written wrong rather than just knowing that I've mentioned the wrong component name. It will save me some time. Wdyt?

hongweiliu17: Ok, let me remove the `continue` to prompt all erros, thanks.

dheerajodha: Thank you!

internal/controller/snapshot/snapshot_adapter.go
hongweiliu17: Thanks for the review, it is not urgent.

dheerajodha: > if users wants to reconcile it after updating it, they have to create a new snapshot.

that's not a very user-friendly UX, but I'm okay with it.

If we have 100s of 'dumb' users creating override snapshots then we'd have at least 100s of invalid override snapshots that are useless in the cluster. Should we then need to update the garbage collector to delete straightaway the **invalid override** snapshots?

dheerajodha (approved):

internal/controller/snapshot/snapshot_adapter.go
hongweiliu17: Snapshot GC has covered it.

PR #837, opened 2024-08-12
fix(stoneintg-971): reconciliation invalid image digest err

jencull:
stop reconciliation when creating snapshot meets invalid image digest error

### DIFF
--- a/internal/controller/buildpipeline/buildpipeline_adapter.go
+++ b/internal/controller/buildpipeline/buildpipeline_adapter.go
@@ -111,8 +111,7 @@ func (a *Adapter) EnsureSnapshotExists() (result controller.OperationResult, err
 
 	expectedSnapshot, err := a.prepareSnapshotForPipelineRun(a.pipelineRun, a.component, a.application)
 	if err != nil {
-		result, err = a.updatePipelineRunWithCustomizedError(&canRemoveFinalizer, err, a.context, a.pipelineRun, a.client, a.logger)
-		return controller.RequeueWithError(err)
+		return a.updatePipelineRunWithCustomizedError(&canRemoveFinalizer, err, a.context, a.pipelineRun, a.client, a.logger)
 	}
 
 	err = a.client.Create(a.context, expectedSnapshot)
@@ -361,7 +360,7 @@ func (a *Adapter) updatePipelineRunWithCustomizedError(canRemoveFinalizer *bool,
 		}
 		logger.Error(cerr, "Build PipelineRun failed with error, should be fixed and re-run manually", "pipelineRun.Name", pipelineRun.Name)
 		*canRemoveFinalizer = true
-		return controller.ContinueProcessing()
+		return controller.StopProcessing()
 	}
 	return controller.RequeueOnErrorOrContinue(cerr)
 
--- a/internal/controller/buildpipeline/buildpipeline_adapter_test.go
+++ b/internal/controller/buildpipeline/buildpipeline_adapter_test.go
@@ -662,7 +662,7 @@ var _ = Describe("Pipeline Adapter", Ordered, func() {
 			Expect(helpers.IsInvalidImageDigestError(err)).To(BeTrue())
 			Eventually(func() bool {
 				result, err := adapter.EnsureSnapshotExists()
-				return !result.CancelRequest && err == nil
+				return result.CancelRequest && err == nil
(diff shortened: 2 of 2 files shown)


internal/controller/buildpipeline/buildpipeline_adapter.go
```diff
@@ -111,6 +111,16 @@ func (a *Adapter) EnsureSnapshotExists() (result controller.OperationResult, err
 
 	expectedSnapshot, err := a.prepareSnapshotForPipelineRun(a.pipelineRun, a.component, a.application)
 	if err != nil {
```
dirgim, line 113: Thinking out loud, the `updatePipelineRunWithCustomizedError()` function seems to do much of the same things as the added lines.
To prevent code duplication, maybe something like this would work too here?
```
	if err != nil {
		result, err = a.updatePipelineRunWithCustomizedError(&canRemoveFinalizer, err, a.context, a.pipelineRun, a.client, a.logger)
		if h.IsInvalidImageDigestError(err) {
		}
	}
```

hongweiliu17: Read [updatePipelineRunWithCustomizedError()](https://github.com/konflux-ci/integration-service/blob/main/internal/controller/buildpipeline/buildpipeline_adapter.go#L357-L364) and I realized the current code has included `h.IsInvalidImageDigestError(cerr)` and return `controller.ContinueProcessing()` if `if h.IsMissingInfoInPipelineRunError(cerr) || h.IsInvalidImageDigestError(cerr) || h.IsMissingValidComponentError(cerr)` , so maybe we can just update this return in [controller.RequeueWithError(err)](https://github.com/konflux-ci/integration-service/blob/main/internal/controller/buildpipeline/buildpipeline_adapter.go#L115)? . Can you @jencull @dirgim help. to confirm it? Thanks!

dirgim: That's a good point, we currently ignore what's returned in [updatePipelineRunWithCustomizedError()](https://github.com/konflux-ci/integration-service/blob/main/internal/controller/buildpipeline/buildpipeline_adapter.go#L357-L364) and just check if an error was returned, which doesn't look right to me.

Since the function in question already returns the result that's expected by the controller, maybe simply doing this would work (untested, would need some testing to confirm):
```
return a.updatePipelineRunWithCustomizedError(&canRemoveFinalizer, err, a.context, a.pipelineRun, a.client, a.logger)
```

dirgim: @jencull are you fine with trying out the proposed change? Let me know if you need more info.

dheerajodha:
@jencull these test suites are consistently timing out while waiting for the build PLR to get created.
I don't think this is flaky behaviour. Could it be due to the changes in your PR?

If not, then my next guess is that this could be because the size of the cluster assigned to our CI is small and unable to run these suites in parallel anymore.

jencull:
> @jencull these test suites are consistently timing out while waiting for the build PLR to get created. I don't think this is flaky behaviour. Could it be due to the changes in your PR?
> 
> If not, then my next guess is that this could be because the size of the cluster assigned to our CI is small and unable to run these suites in parallel anymore.

thanks @dheerajodha  fiddling with annotations and that could be an issue. Will investigate, thanks.

internal/controller/buildpipeline/buildpipeline_adapter.go
hongweiliu17: One question: Does https://github.com/konflux-ci/integration-service/blob/main/internal/controller/buildpipeline/buildpipeline_adapter.go#L364 mean we will continue to call the other `Ensure**` functions in this controller thread? Because I notice https://github.com/konflux-ci/integration-service/blob/main/internal/controller/buildpipeline/buildpipeline_adapter.go#L348 will return  `controller.StopProcessing()` when meeting unrecoverable error `errors.IsForbidden(cerr)`

dirgim (approved):
LGTM code-wise

kasemAlem (approved):

Josh-Everett (approved):

kasemAlem:
it failed on deployment stage 
/test integration-service-e2e

dheerajodha:
Now that we have [1 more worker node added to our CI cluster](https://github.com/openshift/release/pull/55879), let's retry.

/test integration-service-e2e

PR #841, opened 2024-08-15
fix(STONEINTG-980): consider target branch when overwriting ITS

Josh-Everett:
Integration Service was overwriting the revision sha of the integration test scenarios targeting a specific sha with the sha of the source branch. In situations where the source branch sha differs from the ITS within the PR, the sha from the target branch will be used to create the PLR.

### DIFF
--- a/internal/controller/snapshot/snapshot_adapter.go
+++ b/internal/controller/snapshot/snapshot_adapter.go
@@ -600,6 +600,13 @@ func shouldUpdateIntegrationTestGitResolver(integrationTestScenario *v1beta2.Int
 	annotations := snapshot.GetAnnotations()
 	params := resolverParamsToMap(testResolverRef.Params)
 
+	// if target revision sha differs from source revision sha we do not want to overwrite it
+	targetRevision := annotations[gitops.PipelineAsCodeTargetBranchAnnotation]
+	sourceRevision := params[tekton.TektonResolverGitParamRevision]
+	if targetRevision != sourceRevision {
+		return false
+	}
+
 	if urlVal, ok := params[tekton.TektonResolverGitParamURL]; ok {
 		// urlVal may or may not have git suffix specified :')
 		return urlToGitUrl(urlVal) == urlToGitUrl(annotations[gitops.PipelineAsCodeRepoURLAnnotation])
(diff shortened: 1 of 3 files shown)


chipspeak:
LGTM apart from those blasted git linting errors

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -600,6 +600,13 @@ func shouldUpdateIntegrationTestGitResolver(integrationTestScenario *v1beta2.Int
 	annotations := snapshot.GetAnnotations()
 	params := resolverParamsToMap(testResolverRef.Params)
 
+	// if target revision sha differs from source revision sha we do not want to overwrite it
+	sourceRevisionSha := annotations[gitops.PipelineAsCodeSHAAnnotation]
```
dirgim, line 604: I think we should use the target branch (`pipelinesascode.tekton.dev/branch`) instead of the SHA here.

Josh-Everett: I have finding myself mixed up re: the target/source @dirgim, are you referring to [pipelinesascode.tekton.dev/source-branch](https://github.com/konflux-ci/integration-service/blob/a0e1e6c5feef521f210c5ce613e2b03c38f3132e/tekton/build_pipeline.go#L39C2-L40C83)? How does this value get loaded?

hongweiliu17: @dirgim I guess you mean `pac.test.appstudio.openshift.io/branch` in snapshot_adapter.go because the prefix in `pipelinesascode.tekton.dev/branch` is updated when copying [label/annotation](https://github.com/konflux-ci/integration-service/blob/main/gitops/snapshot.go#L866-L867) , right?
@Josh-Everett I think `/branch` mean the branch that PR will be merged to, and `/source-branch` means the PR is opened from, or you may find some pipelineRun/snapshot and compare them.

dirgim: @hongweiliu17 you are correct, I forgot the prefix change when it's applied to Snapshots. 
You are also right about the `/branch` meaning target branch, which is what we should be considering when deciding if we want to overwrite the IntegrationTestScenarios git reference details.

Josh-Everett: Ok I think I was able to untangle my confusion with your help. I updated the existing unit test to make it reflect the new behavior but I am wondering if it is enough to illustrate the desired effect.

kasemAlem (approved):

dirgim (approved):
LGTM aside the tiny nit

dirgim (approved):
LGTM

jsztuka (approved):
lgtm

PR #853, opened 2024-09-02
fix(Stoneintg 1018): handle invalid Gitlab api commit status update

kasemAlem:
Gitlab api will returns 400 error code when trying to update commit status with invalid one: 

- transition  from `running` to  `enqueue`
-  transition from `pending`  to  `enqueue`

for more details it can be tracked in [STONEINTG-1018](https://issues.redhat.com/browse/STONEINTG-1018)

### DIFF
--- a/status/reporter_gitlab.go
+++ b/status/reporter_gitlab.go
@@ -53,6 +53,7 @@ func NewGitLabReporter(logger logr.Logger, k8sClient client.Client) *GitLabRepor
 
 // check if interface has been correctly implemented
 var _ ReporterInterface = (*GitLabReporter)(nil)
+var existingCommitStatus *gitlab.CommitStatus
 
 // Detect if snapshot has been created from gitlab provider
 func (r *GitLabReporter) Detect(snapshot *applicationapiv1alpha1.Snapshot) bool {
@@ -151,16 +152,22 @@ func (r *GitLabReporter) setCommitStatus(report TestReport) error {
 		opt.TargetURL = gitlab.Ptr(url)
 	}
 
-	// Special case for gitLab `running` state because of a bug where it can't be updated to the same state again
-	if glState == gitlab.Running {
+	// Fetch commit statuses only if necessary
+	if glState == gitlab.Running || glState == gitlab.Pending {
 		allCommitStatuses, _, err := r.client.Commits.GetCommitStatuses(r.sourceProjectID, r.sha, nil)
 		if err != nil {
 			return fmt.Errorf("error while getting all commitStatuses for sha %s: %w", r.sha, err)
 		}
-		existingCommitStatus := r.GetExistingCommitStatus(allCommitStatuses, report.FullName)
-		if existingCommitStatus != nil && existingCommitStatus.Status == string(gitlab.Running) {
-			r.logger.Info("Will not update the existing commit status from `running` to `running`",
-				"scenario.name", report.ScenarioName, "commitStatus.ID", existingCommitStatus.ID)
+		existingCommitStatus = r.GetExistingCommitStatus(allCommitStatuses, report.FullName)
+
+		// special case, we want to skip updating commit status if the status from running to running, from enque to pending
+		if existingCommitStatus != nil && (existingCommitStatus.Status == "enqueue" ||
+			existingCommitStatus.Status == string(gitlab.Running)) {
+			r.logger.Info("Skipping commit status update",
+				"scenario.name", report.ScenarioName,
+				"commitStatus.ID", existingCommitStatus.ID,
+				"current_status", existingCommitStatus.Status,
+				"new_status", glState)
 			return nil
 		}
 	}


status/reporter_gitlab.go
```diff
@@ -158,9 +158,9 @@ func (r *GitLabReporter) setCommitStatus(report TestReport) error {
 			return fmt.Errorf("error while getting all commitStatuses for sha %s: %w", r.sha, err)
 		}
 		existingCommitStatus := r.GetExistingCommitStatus(allCommitStatuses, report.FullName)
-		if existingCommitStatus != nil && existingCommitStatus.Status == string(gitlab.Running) {
-			r.logger.Info("Will not update the existing commit status from `running` to `running`",
-				"scenario.name", report.ScenarioName, "commitStatus.ID", existingCommitStatus.ID)
+		if existingCommitStatus != nil && (existingCommitStatus.Status == string(gitlab.Running) || existingCommitStatus.Status == "enqueue") {
```
dirgim, line 161: Do note that this part of the code is only reached if glState == gitlab.Running ([here currently](https://github.com/konflux-ci/integration-service/pull/842/files#diff-cbf7e6d58cc12a67b910b5b9108ba68d206e75cc8560ca0ab2b19e2d34415702R858)).

kasemAlem: my bad, 
you are right , went to check outside of `glState == "running "` 
if the required update to be `enqueue` then skip the update

status/reporter_gitlab.go
```diff
@@ -165,6 +167,13 @@ func (r *GitLabReporter) setCommitStatus(report TestReport) error {
 		}
 	}
 
+	// When trying to update the commit status to `enqueue` don't update
+	if existingCommitStatus != nil && glState == "enqueue" {
```
hongweiliu17, line 171: I am afraid this `existingCommitStatus` is always `nil`, because it is declared [here](https://github.com/konflux-ci/integration-service/pull/853/files#diff-9173b9d26ca5486f2c07b5dfe966142cae4ca991f7787168bc520c2c655ac83aR56),  and then is set [here](https://github.com/konflux-ci/integration-service/pull/853/files#diff-9173b9d26ca5486f2c07b5dfe966142cae4ca991f7787168bc520c2c655ac83aR157-R162) only when `glState == gitlab.Running` .
So I feel you need to move `existingCommitStatus := r.GetExistingCommitStatus(allCommitStatuses, report.FullName)` to a line before [glState == gitlab.Running](https://github.com/konflux-ci/integration-service/pull/853/files#diff-9173b9d26ca5486f2c07b5dfe966142cae4ca991f7787168bc520c2c655ac83aR157)
Or  directly update [if glState == gitlab.Running](https://github.com/konflux-ci/integration-service/pull/853/files#diff-9173b9d26ca5486f2c07b5dfe966142cae4ca991f7787168bc520c2c655ac83aR157) to `if glState == gitlab.Running || glState == "enqueue"` and update their logs as well

status/reporter_gitlab.go
```diff
@@ -157,14 +158,29 @@ func (r *GitLabReporter) setCommitStatus(report TestReport) error {
...
 		}
 	}
 
+	// Skip updating the commit status if the current status is "enqueue"
+	if glState == gitlab.Running || glState == gitlab.Pending {
+		allCommitStatuses, _, err := r.client.Commits.GetCommitStatuses(r.sourceProjectID, r.sha, nil)
+		if err != nil {
+			return fmt.Errorf("error while getting all commitStatuses for sha %s: %w", r.sha, err)
+		}
+		existingCommitStatus = r.GetExistingCommitStatus(allCommitStatuses, report.FullName)
```
hongweiliu17, line 176: I think you can just use the returned `existingCommitStatus` from https://github.com/konflux-ci/integration-service/pull/853/files#diff-9173b9d26ca5486f2c07b5dfe966142cae4ca991f7787168bc520c2c655ac83aR161, no need to call it here again. HDYT?

status/reporter_gitlab.go
```diff
@@ -157,14 +158,24 @@ func (r *GitLabReporter) setCommitStatus(report TestReport) error {
 		if err != nil {
 			return fmt.Errorf("error while getting all commitStatuses for sha %s: %w", r.sha, err)
 		}
-		existingCommitStatus := r.GetExistingCommitStatus(allCommitStatuses, report.FullName)
+		existingCommitStatus = r.GetExistingCommitStatus(allCommitStatuses, report.FullName)
```
dirgim, line 161: We might want to get the existingCommitStatus outside of this condition (e.g. [here](https://github.com/konflux-ci/integration-service/pull/853/files#diff-9173b9d26ca5486f2c07b5dfe966142cae4ca991f7787168bc520c2c655ac83aR154)) so we also have it available for the new condition. 
That's because in the current code, we'd only fetch the existingCommitStatus in case the `glState == gitlab.Running`, while the [other (new) condition](https://github.com/konflux-ci/integration-service/pull/853/files#diff-9173b9d26ca5486f2c07b5dfe966142cae4ca991f7787168bc520c2c655ac83aR171) also includes `glState == gitlab.Pending`.
Another option might be to combine these two conditions into one condition, but I think that's optional.

hongweiliu17: +1

status/reporter_gitlab.go
```diff
@@ -151,18 +152,24 @@ func (r *GitLabReporter) setCommitStatus(report TestReport) error {
...
-		if existingCommitStatus != nil && existingCommitStatus.Status == string(gitlab.Running) {
-			r.logger.Info("Will not update the existing commit status from `running` to `running`",
-				"scenario.name", report.ScenarioName, "commitStatus.ID", existingCommitStatus.ID)
-			return nil
-		}
+		existingCommitStatus = r.GetExistingCommitStatus(allCommitStatuses, report.FullName)
+	}
+
+	// special case, we want to skip updating commit status if the status from running to running, from enque to pending
+	if (glState == gitlab.Running && existingCommitStatus != nil && existingCommitStatus.Status == string(gitlab.Running)) ||
```
dirgim, line 165: nit: it seems like this long condition could be simplified if it was moved under [here](https://konflux-ci.dev/docs/how-tos/testing/integration/choosing-contexts/), wdyt?

dirgim (approved):
LGTM

PR #858, opened 2024-09-04
fix(STONEINTG-998): fix pipeline as code event

hongweiliu17:
* fix IsSnapshotCreatedByPACPushEvent and IsPLRCreatedByPACPushEvent since incoming event can trigger pull-request and push event plr

### DIFF
--- a/tekton/build_pipeline_test.go
+++ b/tekton/build_pipeline_test.go
@@ -47,6 +47,7 @@ var _ = Describe("build pipeline", func() {
 					"pipelines.openshift.io/strategy":          "s2i",
 					"appstudio.openshift.io/component":         "component-sample",
 					"build.appstudio.redhat.com/target_branch": "main",
+					"pipelinesascode.tekton.dev/pull-request":  "1",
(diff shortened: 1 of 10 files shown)


tekton/build_pipeline_test.go
```diff
@@ -47,7 +47,7 @@ var _ = Describe("build pipeline", func() {
 					"pipelines.openshift.io/strategy":          "s2i",
 					"appstudio.openshift.io/component":         "component-sample",
 					"build.appstudio.redhat.com/target_branch": "main",
-					"pipelinesascode.tekton.dev/event-type":    "pull_request",
```
dirgim, line 50: I think it would be good to keep this label because in most cases it will be present, wdyt?

dirgim (approved):
LGTM aside the nit

kasemAlem (approved):

tekton/build_pipeline_test.go
hongweiliu17: Yes, you are right, these labels and annotations should all exist, thanks! Updated.

PR #861, opened 2024-09-05
wip(STONEINTG-1007): Migrate to status.LastPromotedImage field

Josh-Everett:
TODO: write unit test


Josh-Everett:
@dirgim I saw that [1006](https://issues.redhat.com/browse/STONEINTG-1006) which is responsible for the changes to the application-api is still in release pending, do I need to wait to be able to update the reference?

dirgim:
[The PR was merged](https://github.com/konflux-ci/application-api/pull/86) and deployed to the [application-api component](https://github.com/redhat-appstudio/infra-deployments/pull/4305), so I think you should be all set to update the reference.

Josh-Everett:
Will require [this pr](https://github.com/konflux-ci/integration-service/pull/868) for updating application-api references to be merged first

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -312,9 +312,14 @@ func (a *Adapter) EnsureGlobalCandidateImageUpdated() (controller.OperationResul
 			return controller.ContinueProcessing()
 		}
 
-		// look for the expected snapshotComponnet and update
+		// update PromotedImage for all Components of GCL
+		// then look for the expected snapshotComponent and update
 		for _, snapshotComponent := range a.snapshot.Spec.Components {
 			snapshotComponent := snapshotComponent //G601
+			err = a.updateComponentLastPromotedImage(a.context, a.client, componentToUpdate, &snapshotComponent)
```
dirgim, line 319: Should this be executed under [this condition](https://github.com/konflux-ci/integration-service/pull/861/files#diff-cbf7e6d58cc12a67b910b5b9108ba68d206e75cc8560ca0ab2b19e2d34415702R323) only?
Meaning, we should only run this function when the `snapshotComponent`'s `name` is the same as the `componentToUpdate`.

dirgim:
@Josh-Everett I think that the new `updateComponentLastPromotedImage` function will also need to be called alongside [this existing call](https://github.com/konflux-ci/integration-service/blob/7173ad88b720ea01479e45d1b36ae49cfe6b8bcd/internal/controller/snapshot/snapshot_adapter.go#L361) which is executed for override Snapshots.

14rcole (approved):

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -853,6 +858,23 @@ func (a *Adapter) updateComponentSource(ctx context.Context, c client.Client, co
 	return nil
 }
 
+func (a *Adapter) updateComponentLastPromotedImage(ctx context.Context, c client.Client, component *applicationapiv1alpha1.Component, snapshotComponent *applicationapiv1alpha1.SnapshotComponent) error {
+	if component.Status.LastPromotedImage != snapshotComponent.ContainerImage {
```
kasemAlem, line 862: would it be optimal if we check first in case of same image, then early exit, wdyt?
```
  if component.Status.LastPromotedImage == snapshotComponent.ContainerImage {
    return nil
  }
```

internal/controller/snapshot/snapshot_adapter.go
Josh-Everett: Yes I think you are correct. I was misinterpreting `Existing components on all clusters whose status.lastPromotedImage has not been updated`  but on further thought I believe you are correct and it should be only those relevant components.

dirgim (approved):
LGTM

kasemAlem (approved):

kasemAlem:
failed when provisioning oras cluster , will retest

PR #877, opened 2024-09-13
feat(STONEINTG-957): validate ITS if git resolver

kasemAlem:
According to the [Tekton Git Resolver](https://tekton.dev/docs/pipelines/git-resolver/#parameters)  we should not enable ITS creation if both parameters exists in git resolver , URL & Repo+Org 

[STONEINTG-957](https://issues.redhat.com/browse/STONEINTG-957)

### DIFF
--- a/api/v1beta2/integrationtestscenario_webhook.go
+++ b/api/v1beta2/integrationtestscenario_webhook.go
@@ -53,6 +53,39 @@ func (r *IntegrationTestScenario) ValidateCreate() (warnings admission.Warnings,
 					"param manually defined because it will be automatically generated"+
 					"by the integration service")
 		}
+		// we won't enable ITS if git resolver with url & repo+org
+		urlResolverExist := false
+		repoResolverExist := false
+		orgResolverExist := false
+
+		for _, gitResolverParam := range r.Spec.ResolverRef.Params {
+			if gitResolverParam.Name == "url" && gitResolverParam.Value != "" {
+				urlResolverExist = true
+			}
+			if gitResolverParam.Name == "repo" && gitResolverParam.Value != "" {
+				repoResolverExist = true
+			}
+			if gitResolverParam.Name == "org" && gitResolverParam.Value != "" {
+				orgResolverExist = true
+			}
+		}
+
+		if urlResolverExist {
+			if repoResolverExist || orgResolverExist {
+				return nil, field.Invalid(field.NewPath("Spec").Child("ResolverRef").Child("Params"), param.Name,
+					"an IntegrationTestScenario resource can only have one of the gitResolver parameters,"+
+						"either url or repo (with org), but not both.")
+
+			}
+		} else {
+			if !repoResolverExist || !orgResolverExist {
+				return nil, field.Invalid(field.NewPath("Spec").Child("ResolverRef").Child("Params"), param.Name,
+					"IntegrationTestScenario is invalid: missing mandatory repo or org parameters."+
+						"If both are absent, a valid url is highly recommended.")
+
+			}
+		}
+
 	}
 
 	return nil, nil
--- a/api/v1beta2/integrationtestscenario_webhook_test.go
+++ b/api/v1beta2/integrationtestscenario_webhook_test.go
@@ -26,7 +26,7 @@ import (
 
 var _ = Describe("IntegrationTestScenario webhook", func() {
 
-	var integrationTestScenario *IntegrationTestScenario
+	var integrationTestScenario, integrationTestScenarioInvalidGitResolver *IntegrationTestScenario
 
 	BeforeEach(func() {
 		integrationTestScenario = &IntegrationTestScenario{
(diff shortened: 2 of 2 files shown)


api/v1beta2/integrationtestscenario_webhook.go
```diff
@@ -53,6 +53,28 @@ func (r *IntegrationTestScenario) ValidateCreate() (warnings admission.Warnings,
...
+			if gitResolverParam.Name == "URL" && gitResolverParam.Value != "" {
+				urlResolverExist = true
+			}
+			if gitResolverParam.Name == "REPO" && gitResolverParam.Value != "" {
+				repoResolverExist = true
+			}
+			if gitResolverParam.Name == "ORG" && gitResolverParam.Value != "" {
+				repoResolverExist = true
+			}
+			if urlResolverExist && repoResolverExist && orgResolverExist {
```
14rcole, line 71: This will only fail if all three parameters exist.  What if we have URL, Repo, and no org?  Do we continue silently and just use the URL?  Do we fail because repo was provided with no org?   Something else?

dirgim: That's a good point, I think that Tekton would fail even if just one of the repo and org was defined alongside the URL.
Maybe something like this would work better:
```
if urlResolverExist && (repoResolverExist || orgResolverExist) {
```

kasemAlem: yep , 
thanks @14rcole  for this important comment, 
agree with @dirgim suggestion, looks good enough

dirgim (approved):
LGTM

api/v1beta2/integrationtestscenario_webhook.go
```diff
@@ -53,6 +53,28 @@ func (r *IntegrationTestScenario) ValidateCreate() (warnings admission.Warnings,
...
+			if gitResolverParam.Name == "URL" && gitResolverParam.Value != "" {
+				urlResolverExist = true
+			}
+			if gitResolverParam.Name == "REPO" && gitResolverParam.Value != "" {
+				repoResolverExist = true
+			}
+			if gitResolverParam.Name == "ORG" && gitResolverParam.Value != "" {
+				repoResolverExist = true
+			}
+			if urlResolverExist && (repoResolverExist || orgResolverExist) {
```
MartinBasti, line 71: This should be probably outside of loop for better readability.

MartinBasti, line 71: This check doesn't check for missing repo or org, if repo+org config is used. It will allow to specify only repo without org.

api/v1beta2/integrationtestscenario_webhook.go
```diff
@@ -53,6 +53,28 @@ func (r *IntegrationTestScenario) ValidateCreate() (warnings admission.Warnings,
...
+
+		for _, gitResolverParam := range r.Spec.ResolverRef.Params {
+			if gitResolverParam.Name == "URL" && gitResolverParam.Value != "" {
+				urlResolverExist = true
+			}
+			if gitResolverParam.Name == "REPO" && gitResolverParam.Value != "" {
+				repoResolverExist = true
+			}
+			if gitResolverParam.Name == "ORG" && gitResolverParam.Value != "" {
+				repoResolverExist = true
```
MartinBasti, line 69: this is wrong variable, should be `orgResolverExist`

api/v1beta2/integrationtestscenario_webhook.go
kasemAlem: taking this condition out of loop would work , but the way it's implemented here will apply fast exit on case condition is met then we return a proper error . @MartinBasti, isn't ?

MartinBasti: yes, but no if you address the case that org must be defined together with requirements to have both repo+org.

I'd say it's micro-optimization that is not worth

kasemAlem: right, 
I've fixed so the condition is moved out loop

kasemAlem: good point @MartinBasti 
I've fixed it to cover the missing use cases
also I ran the unit tests with all scenarios , it works fine for me

api/v1beta2/integrationtestscenario_webhook.go
```diff
@@ -53,6 +53,38 @@ func (r *IntegrationTestScenario) ValidateCreate() (warnings admission.Warnings,
...
+			}
+			if gitResolverParam.Name == "org" && gitResolverParam.Value != "" {
+				orgResolverExist = true
+			}
+		}
+
+		if urlResolverExist {
+			if repoResolverExist || orgResolverExist {
+				return nil, field.Invalid(field.NewPath("Spec").Child("ResolverRef").Child("Params"), param.Name,
+					"an IntegrationTestScenario resource should not have both gitResolver "+
```
MartinBasti, line 76: this sentence looks unfinished

api/v1beta2/integrationtestscenario_webhook.go
```diff
@@ -53,6 +53,38 @@ func (r *IntegrationTestScenario) ValidateCreate() (warnings admission.Warnings,
...
+						"Either url, or repo (with org) must be specified, but not both")
+			}
+		} else {
+			if !repoResolverExist || !orgResolverExist {
+				return nil, field.Invalid(field.NewPath("Spec").Child("ResolverRef").Child("Params"), param.Name,
+					"an IntegrationTestScenario Invalid for missing at least on of the mandtory"+
```
MartinBasti, line 82:
```suggestion
					"an IntegrationTestScenario is Invalid for missing at least one of the mandatory"+
```

kasemAlem: fixed

api/v1beta2/integrationtestscenario_webhook.go
kasemAlem: do you mean the sentence is not correct ? 

```
"an IntegrationTestScenario resource should not have both gitResolver "+
						"Either url, or repo (with org) must be specified, but not both.")
```

MartinBasti: `Either` starts with capital letter, it means new sentence. While there is no full stop `.` in previous line.

`both gitResolver` both what? Thee should be 2 things named.

api/v1beta2/integrationtestscenario_webhook.go
```diff
@@ -53,6 +53,39 @@ func (r *IntegrationTestScenario) ValidateCreate() (warnings admission.Warnings,
...
+				return nil, field.Invalid(field.NewPath("Spec").Child("ResolverRef").Child("Params"), param.Name,
+					"an IntegrationTestScenario resource can only have one of the gitResolver parameters,"+
+						"either url or repo (with org), but not both.")
+
+			}
+		} else {
+			if !repoResolverExist || !orgResolverExist {
+				return nil, field.Invalid(field.NewPath("Spec").Child("ResolverRef").Child("Params"), param.Name,
+					"an IntegrationTestScenario is Invalid for missing at least one of the mandatory"+
+						" repo & org.")
```
MartinBasti, line 84:
```suggestion
						" repo & org parameters.")
```

Also, if none of the parameters is defined, we should probably recommend to set url parameter as that is the one supported by konflux UI at this point.

Josh-Everett (approved):
LGTM changes to wording looks good

dirgim (approved):
LGTM after updates

PR #886, opened 2024-09-26
fix(STONEINTG-1048): stop reconciliation when failing to get pr status

Josh-Everett:
When gitlab/github repo is deleted group snapshot creation fails while attempting to get the pr status from the repo.

### DIFF
--- a/internal/controller/snapshot/snapshot_adapter.go
+++ b/internal/controller/snapshot/snapshot_adapter.go
@@ -603,6 +603,10 @@ func (a *Adapter) EnsureGroupSnapshotExist() (controller.OperationResult, error)
 
 	groupSnapshot, componentSnapshotInfos, err := a.prepareGroupSnapshot(a.application, prGroupHash)
 	if err != nil {
+		if strings.Contains(err.Error(), "failed to get pull request for") {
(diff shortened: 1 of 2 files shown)


internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -602,6 +602,10 @@ func (a *Adapter) EnsureGroupSnapshotExist() (controller.OperationResult, error)
 
 	groupSnapshot, componentSnapshotInfos, err := a.prepareGroupSnapshot(a.application, prGroupHash)
 	if err != nil {
+		if strings.Contains(err.Error(), "failed to get pull request for GitHub") {
```
kasemAlem, line 605: I wonder is this would work for both Gitlab and Github , then we better make the error message  more informative as put what is the related git provider , wdyt?

Josh-Everett: Yes, I think you are correct, though I believe there is a few other sections of code from[ group snapshots](https://issues.redhat.com/browse/STONEINTG-998) that refer to github that apply to gitlab as well.

kasemAlem (approved):

chipspeak (approved):
LGTM

PR #893, opened 2024-10-07
fix(KFLUXBUGS-1340):  add branch param from fork to snapshot

kasemAlem:
fixing when the source of git repo of fork PR should take the branch so which was lately exposed in PaC v0.27.1

[KFLUXBUGS-1340](https://issues.redhat.com/browse/KFLUXBUGS-1340)

### DIFF
--- a/gitops/snapshot.go
+++ b/gitops/snapshot.go
@@ -77,6 +77,9 @@ const (
 	// SnapshotGitSourceRepoURLAnnotation contains URL of the git source repository (usually needed for forks)
 	SnapshotGitSourceRepoURLAnnotation = "test.appstudio.openshift.io/source-repo-url"
 
+	// PipelineAsCodeGitProviderAnnotation contains the git provider attached via PaC (e.g., github, gitlab)
+	PipelineAsCodeGitSourceURLAnnotation = PipelinesAsCodePrefix + "/source-repo-url"
+
 	// SnapshotStatusReportAnnotation contains metadata of tests related to status reporting to git provider
 	SnapshotStatusReportAnnotation = "test.appstudio.openshift.io/git-reporter-status"
 
--- a/internal/controller/snapshot/snapshot_adapter.go
+++ b/internal/controller/snapshot/snapshot_adapter.go
@@ -756,7 +756,7 @@ func shouldUpdateIntegrationTestGitResolver(integrationTestScenario *v1beta2.Int
 func getGitResolverUpdateMap(snapshot *applicationapiv1alpha1.Snapshot) map[string]string {
 	annotations := snapshot.GetAnnotations()
 	return map[string]string{
-		tekton.TektonResolverGitParamURL:      urlToGitUrl(annotations[gitops.SnapshotGitSourceRepoURLAnnotation]), // should have .git in url for consistency and compatibility
+		tekton.TektonResolverGitParamURL:      urlToGitUrl(annotations[gitops.PipelineAsCodeGitSourceURLAnnotation]), // should have .git in url for consistency and compatibility
 		tekton.TektonResolverGitParamRevision: annotations[gitops.PipelineAsCodeSHAAnnotation],
 	}
 }
--- a/internal/controller/snapshot/snapshot_adapter_test.go
+++ b/internal/controller/snapshot/snapshot_adapter_test.go
@@ -983,9 +983,11 @@ var _ = Describe("Snapshot Adapter", Ordered, func() {
 						foundRevision = true
 						Expect(param.Value.StringVal).To(Equal(sourceRepoRef))
 					}
+
 				}
 				Expect(foundUrl).To(BeTrue())
 				Expect(foundRevision).To(BeTrue())
+
 			})
 
 		})


kasemAlem:
failed on rosa provision and dashboard upload 
 	/retest

gitops/snapshot.go
```diff
@@ -77,6 +77,9 @@ const (
 	// SnapshotGitSourceRepoURLAnnotation contains URL of the git source repository (usually needed for forks)
 	SnapshotGitSourceRepoURLAnnotation = "test.appstudio.openshift.io/source-repo-url"
 
+	// SnapshotGitSourceRepoBranchAnnotation contains branch of the git source repository (usually needed for forks)
+	SnapshotGitSourceRepoBranchAnnotation = "test.appstudio.openshift.io/source-repo-branch"
```
hongweiliu17, line 81: Do we have `/source-repo-branch` exported by PAC? I know we have `/source-branch` exported by PAC

gitops/snapshot.go
```diff
@@ -155,6 +158,9 @@ const (
 	// PipelineAsCodeSHAAnnotation is the commit which triggered the pipelinerun in build service.
 	PipelineAsCodeSHAAnnotation = PipelinesAsCodePrefix + "/sha"
 
+	// PipelineAsCodBranchAnnotation is the commit which triggered the pipelinerun in build service.
+	PipelineAsCodBranchAnnotation = PipelinesAsCodePrefix + "/branch"
```
hongweiliu17, line 162: We have had this annotation `/branch` as target branch, see line 144 above

gitops/snapshot.go
kasemAlem: yes, 
double check the[ PaC release v0.27.1](https://github.com/openshift-pipelines/pipelines-as-code/commit/789342d31e705c1c179442968ee8b3b38af7776f#diff-5c2a268808defe023737ef3597791fad8ede63bb8adc22721ae25660ad7a361fR35) , it's  `/source-branch` , 
fixed

gitops/snapshot.go
kasemAlem: nice catch ,
removed this duplication

jencull (approved):
lgtm

gitops/snapshot.go
```diff
@@ -77,6 +77,9 @@ const (
 	// SnapshotGitSourceRepoURLAnnotation contains URL of the git source repository (usually needed for forks)
 	SnapshotGitSourceRepoURLAnnotation = "test.appstudio.openshift.io/source-repo-url"
 
+	// SnapshotGitSourceRepoBranchAnnotation contains branch of the git source repository (usually needed for forks)
+	SnapshotGitSourceRepoBranchAnnotation = "test.appstudio.openshift.io/source-branch"
```
jsztuka, line 81: I see this new constant being added, where is it being used exactly?

kasemAlem: no, 
was not used in any place,  I forgot to clean it after debugging , removed
thx

jsztuka (approved):
lgtm, just needs some correction in unit tests.

kasemAlem:
deployment issue, 
/retest

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -746,7 +746,8 @@ func shouldUpdateIntegrationTestGitResolver(integrationTestScenario *v1beta2.Int
 
 	if urlVal, ok := params[tekton.TektonResolverGitParamURL]; ok {
 		// urlVal may or may not have git suffix specified :')
-		return urlToGitUrl(urlVal) == urlToGitUrl(annotations[gitops.PipelineAsCodeRepoURLAnnotation])
+		return urlToGitUrl(urlVal) == urlToGitUrl(annotations[gitops.PipelineAsCodeRepoURLAnnotation]) &&
+			params[tekton.TektonResolverGitParamBranch] == annotations[gitops.PipelineAsCodeTargetBranchAnnotation]
```
dirgim, line 750: I don't think this would solve the issue from the original [KFLUXBUGS-1340](https://issues.redhat.com/browse/KFLUXBUGS-1340) bug.
The underlying issue was that the `CHAINS-GIT_URL` build pipeline result wasn't reliable enough because it depended on users setting it to be equal to `source_url` PaC variable. 
If the users don't set it that way in their build pipelineRun definition, the above logic fails and causes the behavior in the [KFLUXBUGS-1340](https://issues.redhat.com/browse/KFLUXBUGS-1340) bug.

dirgim: I think that a possible solution would be to update the [getGitResolverUpdateMap](https://github.com/konflux-ci/integration-service/blob/3009a326ac53631063a730ae7b37c5d711b33bc5/internal/controller/snapshot/snapshot_adapter.go#L756C6-L762) function to use the newly introduced PaC `source-repo-url` annotation  [here](https://github.com/konflux-ci/integration-service/blob/3009a326ac53631063a730ae7b37c5d711b33bc5/internal/controller/snapshot/snapshot_adapter.go#L759) instead of the `test.appstudio.openshift.io/source-repo-url` (which is equal to the `CHAINS-GIT_URL` result).

kasemAlem: thanks for the important correction @dirgim ,   based on your two former comments,
would it be enough to use  PaC annotation `pac.test.appstudio.openshift.io/source-repo-url` without taking in consideration the branch one  `pac.test.appstudio.openshift.io/source-branch`?

dirgim: As far as I'm aware, that should be correct. I don't think that the source branch needs to be considered.

gitops/snapshot.go
```diff
@@ -158,6 +161,8 @@ const (
 	// PipelineAsCodeSHAAnnotation is the commit which triggered the pipelinerun in build service.
 	PipelineAsCodeSHAAnnotation = PipelinesAsCodePrefix + "/sha"
 
+	//
```
dirgim, line 164: This can probably be removed.

dirgim (approved):
LGTM

PR #897, opened 2024-10-10
feat(stoneintg-1003): add failure annotation to build PLRs

14rcole:
This PR adds an annotation to failed build PLRs.  When the integration service is attempting to create a snapshot for a PR group, it will check for this annotation for all PRs in the group and, if any build pipelines have failed, report the error to all PRs

### DIFF
--- a/internal/controller/buildpipeline/buildpipeline_adapter.go
+++ b/internal/controller/buildpipeline/buildpipeline_adapter.go
@@ -330,6 +330,15 @@ func (a *Adapter) handleUnsuccessfulPipelineRun(canRemoveFinalizer *bool) {
 	if h.HasPipelineRunFinished(a.pipelineRun) || a.pipelineRun.GetDeletionTimestamp() != nil {
 		// The pipeline run has failed
 		// OR pipeline has been deleted but it's still in running state (tekton bug/feature?)
+
+		// Add a failure annotation.  Used to propagate failure information to other PRs in group
+		FailureAnnotation := fmt.Sprintf("%s/%s", tekton.ResourceLabelSuffix, "failure")
+		FailureAnnotationValue := fmt.Sprintf("The pipeline %s failed", a.pipelineRun.Name)
+		err := tekton.AnnotateBuildPipelineRun(a.context, a.pipelineRun, FailureAnnotation, FailureAnnotationValue, a.client)
+		if err != nil {
+			return
+		}
+
 		a.logger.Info("Finished processing of unsuccessful build PLR",
 			"statusCondition", a.pipelineRun.GetStatusCondition(),
 			"deletionTimestamp", a.pipelineRun.GetDeletionTimestamp(),


14rcole:
Note: the drawback to this approach is that if _all_ build pipelines fail for a PR group then the error will not be propagated.  For that the error would have to be propagated by the build pipeline.  However, if all builds fail then every PR will have a failure message from the build pipeline anyway.

internal/controller/buildpipeline/buildpipeline_adapter.go
```diff
@@ -330,6 +330,12 @@ func (a *Adapter) handleUnsuccessfulPipelineRun(canRemoveFinalizer *bool) {
 	if h.HasPipelineRunFinished(a.pipelineRun) || a.pipelineRun.GetDeletionTimestamp() != nil {
 		// The pipeline run has failed
 		// OR pipeline has been deleted but it's still in running state (tekton bug/feature?)
+
+		// Add a failure annotation.  Used to propagate failure information to other PRs in group
+		FailureAnnotation = fmt.Sprintf("%s/%s", tekton.ResourceLabelSuffix, "failure")
```
dirgim, line 335: This should do the trick of resolving the check failures:
```suggestion
		FailureAnnotation := fmt.Sprintf("%s/%s", tekton.ResourceLabelSuffix, "failure")
```

internal/controller/buildpipeline/buildpipeline_adapter.go
```diff
@@ -330,6 +330,12 @@ func (a *Adapter) handleUnsuccessfulPipelineRun(canRemoveFinalizer *bool) {
...
+		FailureAnnotationValue = fmt.Sprintf("The pipeline %s failed", a.pipelineRun.Name)
```
dirgim, line 336: Same here, this should fix it:
```suggestion
		FailureAnnotationValue := fmt.Sprintf("The pipeline %s failed", a.pipelineRun.Name)
```

internal/controller/buildpipeline/buildpipeline_adapter.go
```diff
@@ -330,6 +330,15 @@ func (a *Adapter) handleUnsuccessfulPipelineRun(canRemoveFinalizer *bool) {
 	if h.HasPipelineRunFinished(a.pipelineRun) || a.pipelineRun.GetDeletionTimestamp() != nil {
 		// The pipeline run has failed
 		// OR pipeline has been deleted but it's still in running state (tekton bug/feature?)
+
+		// Add a failure annotation.  Used to propagate failure information to other PRs in group
+		FailureAnnotation := fmt.Sprintf("%s/%s", tekton.ResourceLabelSuffix, "failure")
+		FailureAnnotationValue := fmt.Sprintf("The pipeline %s failed", a.pipelineRun.Name)
+		err := tekton.AnnotateBuildPipelineRun(a.context, a.pipelineRun, FailureAnnotation, FailureAnnotationValue, a.client)
```
hongweiliu17, line 337: Don't know why the comment I left yesterday disappear. 
I think we need to annotate component snapshot as [comment](https://issues.redhat.com/browse/STONEINTG-1003?focusedId=25757403&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-25757403), not build plr.
Once a build PLR fails, users will know it obviously. But other component snapshot belonging to the same PR group don't know it, this is why this story need to notify them.

14rcole: I see, I misunderstood that.  What happens to component snapshots for components in PR group?  Do we do the full set of tests with them?  I would expect us to terminate if only some of the components in a PR group are in the snapshot

hongweiliu17: Yes, if there is only one component snapshot for that pr group, we don't need to annotate it.

14rcole: On second thought, I don't think this is a good approach.  We can only annotate component snapshots for components that have been built.  If a build fails immediately it will have nothing to update.

dirgim: I think that the point of this ticket is to annotate all component Snapshots that did have successful builds in to send a notification on their PRs that some of the other builds for their PR group didn't succeed.

14rcole: So if a build pipeline fails should we wait until all pipelines in the PR group have finished?  That seems like a lot of overhead just to notify other PRs

hongweiliu17: Refer to [this code](https://github.com/konflux-ci/integration-service/blob/main/internal/controller/snapshot/snapshot_adapter.go#L552-L592), it has been handling the pipelineruns.

dirgim:
Closing this as it was superseded by #916

PR #912, opened 2024-10-29
feat(STONEINTG-1065): let users rerun ALL tests

dheerajodha:
* Before this commit, a user can only rerun one test at a time by adding the "run" label on snapshot with the name of ITS they want to rerun.
* After this commit, a user can add the value of "all" in the "run" label of snapshot to automatically rerun ALL the tests easily.

### DIFF
--- a/internal/controller/snapshot/snapshot_adapter.go
+++ b/internal/controller/snapshot/snapshot_adapter.go
@@ -91,73 +91,125 @@ func scenariosNamesToList(integrationTestScenarios *[]v1beta2.IntegrationTestSce
 	return &result
 }
 
-// EnsureRerunPipelineRunsExist is responsible for recreating integration test pipelines triggered by users
+// EnsureRerunPipelineRunsExist is responsible for recreating integration test pipelineruns triggered by users
 func (a *Adapter) EnsureRerunPipelineRunsExist() (controller.OperationResult, error) {
-
-	scenarioName, ok := gitops.GetIntegrationTestRunLabelValue(a.snapshot)
+	runLabelValue, ok := gitops.GetIntegrationTestRunLabelValue(a.snapshot)
 	if !ok {
-		// no test rerun triggered
 		return controller.ContinueProcessing()
 	}
-	integrationTestScenario, err := a.loader.GetScenario(a.context, a.client, scenarioName, a.application.Namespace)
 
-	if err != nil {
-		if clienterrors.IsNotFound(err) {
-			a.logger.Error(err, "scenario for integration test re-run not found", "scenario", scenarioName)
-			// scenario doesn't exist just remove label and continue
-			if err = gitops.RemoveIntegrationTestRerunLabel(a.context, a.client, a.snapshot); err != nil {
-				return controller.RequeueWithError(err)
-			}
-			return controller.ContinueProcessing()
-		}
-		return controller.RequeueWithError(fmt.Errorf("failed to fetch requested scenario %s: %w", scenarioName, err))
+	integrationTestScenarios, opResult, err := a.getScenariosToRerun(runLabelValue)
+	if integrationTestScenarios == nil {
+		return opResult, err
 	}
-
-	a.logger.Info("Re-running integration test for scenario", "scenario", scenarioName)
 
 	testStatuses, err := gitops.NewSnapshotIntegrationTestStatusesFromSnapshot(a.snapshot)
 	if err != nil {
 		return controller.RequeueWithError(err)
 	}
 
-	integrationTestScenarioStatus, ok := testStatuses.GetScenarioStatus(integrationTestScenario.Name)
-	if ok && (integrationTestScenarioStatus.Status == intgteststat.IntegrationTestStatusInProgress ||
-		integrationTestScenarioStatus.Status == intgteststat.IntegrationTestStatusPending) {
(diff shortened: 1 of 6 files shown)


internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -90,73 +90,125 @@ func scenariosNamesToList(integrationTestScenarios *[]v1beta2.IntegrationTestSce
...
+		return opResult, err
+	}
+
+	testStatuses, err := gitops.NewSnapshotIntegrationTestStatusesFromSnapshot(a.snapshot)
+	if err != nil {
+		return controller.RequeueWithError(err)
+	}
+
+	skipScenarioRerunCount, opResult, err := a.handleScenarioReruns(integrationTestScenarios, testStatuses)
+	if skipScenarioRerunCount == -1 {
```
jsztuka, line 111: When does this condition appears? How can skip scenario rerun count be negative?

dheerajodha: > How can skip scenario rerun count be negative?

I'm returning skipScenarioRerunCount to "-1" [here](https://github.com/konflux-ci/integration-service/pull/912/files#diff-cbf7e6d58cc12a67b910b5b9108ba68d206e75cc8560ca0ab2b19e2d34415702R170).

I'm doing to pick up the situation when we have called either [StopProcessing()](https://github.com/konflux-ci/operator-toolkit/blob/ef6dcbeca69d/controller/operation.go#L87) or [RequeueWithError()](https://github.com/konflux-ci/operator-toolkit/blob/ef6dcbeca69d/controller/operation.go#L78) in any of the previous function calls. The tricky thing is `StopProcessing()` function [returns nil error](https://github.com/konflux-ci/operator-toolkit/blob/ef6dcbeca69d8d6d47ac5d58cf9183543b2deed9/controller/operation.go#L92), so we can't just use something like `if err != nil { //handle error }`.

This is why I tried to catch both of these type of calls using [if opResult.CancelRequest || err != nil](https://github.com/konflux-ci/integration-service/pull/912/files#diff-cbf7e6d58cc12a67b910b5b9108ba68d206e75cc8560ca0ab2b19e2d34415702R168) condition, and returned `-1` to help us identify this state later on.

But now that I think about it, I can replace the above condition to:
```
if opResult.CancelRequest || err != nil {
```
And it would work the same, let me know if you have any concerns about this. Thanks for making me re-think!

jsztuka: Thank you sir for explanation, I'm fine with either way.

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -90,73 +90,125 @@ func scenariosNamesToList(integrationTestScenarios *[]v1beta2.IntegrationTestSce
...
+		scenarios, err := a.loader.GetAllIntegrationTestScenariosForApplication(a.context, a.client, a.application)
+		if err != nil {
+			a.logger.Error(err, "Failed to get IntegrationTestScenarios", "Application.Namespace", a.application.Namespace)
+			opResult, err := controller.RequeueWithError(err)
+			return nil, opResult, err
+		}
+		filteredScenarios := gitops.FilterIntegrationTestScenariosWithContext(scenarios, a.snapshot)
+		if filteredScenarios == nil {
+			a.logger.Info("None of the Scenarios' context are applicable to the Snapshot, nothing to re-run")
+		}
```
dirgim, line 134: Would the new [GetRequiredIntegrationTestScenariosForSnapshot](https://github.com/konflux-ci/integration-service/blob/7a6ff94f0ce3a7dc074f7382e4d5d6b78b278b3d/loader/loader.go#L227C18-L227C64) function be a good replacement for this code block?

dheerajodha: Hmm, that function only returns required ITS, but when a user puts the `run` label's value to "all" maybe they'd expect to run all the ITS (even optional ones), or would they not?

The answer to the above question would decide if we want to refactor the suggested function to have a boolean variable to deal with that.

dirgim: Oh, good point. Yeah maybe we could make that function a bit more generic by adding an `includeOptional` bool that dictates if it also returns the optional IntegrationTestScenarios. 
Also @hongweiliu17 wdyt?

dirgim: Tbh this is more of a nice-to have suggestion, so I'm also fine if the code remains as-is.

dirgim (approved):
LGTM aside the optional suggestion

kasemAlem (approved):

jsztuka (approved):
lgtm

internal/controller/snapshot/snapshot_adapter.go
hongweiliu17: Sorry, I didn't notice this message previously.

hongweiliu17: I agree to have this nice-to-have, the function can the name like GetAllIntegrationTestScenariosForSnapshot(), then we can reduce the code complexity all the time.

hongweiliu17 (approved):
LGTM aside the nice-to-have.

Josh-Everett (approved):

PR #922, opened 2024-11-04
feat(STONEINTG-1071): skip checking snapshot not affected by pr group

hongweiliu17:
* don't check the snapshots which are not affected by pr group during group snapshot preparation

### DIFF
--- a/internal/controller/snapshot/snapshot_adapter.go
+++ b/internal/controller/snapshot/snapshot_adapter.go
@@ -855,7 +855,7 @@ func (a *Adapter) updateComponentLastPromotedImage(ctx context.Context, c client
 }
 
 func (a *Adapter) prepareGroupSnapshot(application *applicationapiv1alpha1.Application, prGroup, prGroupHash string) (*applicationapiv1alpha1.Snapshot, []gitops.ComponentSnapshotInfo, error) {
-	componentsToCheck, err := a.getComponentsForPRGroup(application, prGroup, prGroupHash)
+	componentsToCheck, err := a.getComponentsForPRGroup(prGroup, prGroupHash)
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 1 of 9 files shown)


internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -1011,3 +976,93 @@ func isLatestBuildPipelineRunInComponent(pipelineRun *tektonv1.PipelineRun, pipe
...
+// haveAllPipelineRunProcessedForPrGroup checks if all build plr has been processed for the given pr group
+func (a *Adapter) haveAllPipelineRunProcessedForPrGroup(prGroup, prGroupHash string) (bool, error) {
+	pipelineRuns, err := a.loader.GetPipelineRunsWithPRGroupHash(a.context, a.client, a.snapshot, prGroupHash)
+	if err != nil {
+		a.logger.Error(err, fmt.Sprintf("Failed to get build pipelineRuns for given pr group hash %s", prGroupHash))
+		return false, err
+	}
+
+	for _, pipelineRun := range *pipelineRuns {
+		pipelineRun := pipelineRun
```
kasemAlem, line 989: why do we need this assignment ?

hongweiliu17: It is the same usage in other loop using `&` to avoid G601, let me comment it.

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -896,7 +865,16 @@ func (a *Adapter) updateComponentLastPromotedImage(ctx context.Context, c client
...
 }
 
-func (a *Adapter) prepareGroupSnapshot(application *applicationapiv1alpha1.Application, prGroupHash string) (*applicationapiv1alpha1.Snapshot, []gitops.ComponentSnapshotInfo, error) {
+func (a *Adapter) prepareGroupSnapshot(application *applicationapiv1alpha1.Application, prGroup, prGroupHash string) (*applicationapiv1alpha1.Snapshot, []gitops.ComponentSnapshotInfo, error) {
+	componentsToCheck, err := a.getComponentsForPRGroup(application, prGroup, prGroupHash)
+	if err != nil {
+		return nil, nil, err
+	}
+	if len(componentsToCheck) < 2 {
+		a.logger.Info(fmt.Sprintf("The number %d of component affected by this pr group %s is less than 2, skipping group snapshot creation", len(componentsToCheck), prGroup))
```
dirgim, line 874: nit:
```suggestion
		a.logger.Info(fmt.Sprintf("The number %d of components affected by this PR group %s is less than 2, skipping group snapshot creation", len(componentsToCheck), prGroup))
```

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -907,35 +885,38 @@ func (a *Adapter) prepareGroupSnapshot(application *applicationapiv1alpha1.Appli
...
-			return nil, nil, err
-		}
 
-		sortedSnapshots := gitops.SortSnapshots(*snapshots)
-		// find the latest component snapshot created for open PR/MR
-		for _, snapshot := range sortedSnapshots {
-			snapshot := snapshot
-			// find the built image for pull/merge request build PLR from the latest opened pull request component snapshot
-			isPRMROpened, err = a.status.IsPRMRInSnapshotOpened(a.context, &snapshot)
+		if slices.Contains(componentsToCheck, applicationComponent.Name) {
```
dirgim, line 889: Maybe we could, instead of nesting the logic within this `if` condition, just call `continue` if this condition is not true, e.g:
```
if !slices.Contains(componentsToCheck, applicationComponent.Name) {
    continue
}
snapshots, err := a.loader.GetMatchingComponentSnapshotsForComponentAndPRGroupHash(a.context, a.client, a.snapshot, applicationComponent.Name, prGroupHash)
(...etc...)
```

hongweiliu17: If not containing that component, we still need to get snapshotComponent data from GCL for this component.

dirgim: Ah, ok that's a good point.

internal/controller/snapshot/snapshot_adapter.go
hongweiliu17: Updated. Thank you!

internal/controller/snapshot/snapshot_adapter.go
dirgim: Maybe in that case everything that's contained within this `if ` condition could be extracted to a separate function since its logic is contained to finding out if a PR/MR is opened? 
Something like `findSnapshotWithOpenedPR()` or similar?

hongweiliu17: Yes, I tried it yesterday like [this](https://github.com/konflux-ci/integration-service/commit/39febc38d38f3322267ed70af08c5205b74b7601#diff-cbf7e6d58cc12a67b910b5b9108ba68d206e75cc8560ca0ab2b19e2d34415702R1045-R1068), then I checked the complexity but I found the complexity doesn't change because I checked EnsureGroupSnapshotExist() wrongly not this prepareGroupSnapshot.

I am also thinking to use pointer for `groupSnapshot, componentSnapshotInfos` to have code like ` if ok {func_a} else {func_b}` without returning groupSnapshot, componentSnapshotInfos

hongweiliu17: `findSnapshotWithOpenedPR()` is a good name, let me update. Thank you.

dirgim: That looks great, thanks!

jsztuka (approved):
lgtm

dirgim (approved):
LGTM

kasemAlem (approved):

PR #943, opened 2024-11-25
fix(STONEINTG-1091): don't set commitStatus for PR/MR from forked repo

hongweiliu17:
* Don't set commitStatus if the PR/MR is from forked repo since the limitation access

### DIFF
--- a/status/reporter_github.go
+++ b/status/reporter_github.go
@@ -417,49 +417,56 @@ func (csu *CommitStatusUpdater) updateStatusInComment(ctx context.Context, repor
 // UpdateStatus updates commit status in PR
 func (csu *CommitStatusUpdater) UpdateStatus(ctx context.Context, report TestReport) error {
 
-	allCommitStatuses, err := csu.getAllCommitStatuses(ctx)
-	if err != nil {
-		csu.logger.Error(err, "failed to get all CommitStatuses for scenario",
-			"snapshot.NameSpace", csu.snapshot.Namespace, "snapshot.Name", csu.snapshot.Name,
-			"scenario.Name", report.ScenarioName)
-		return err
-	}
-
-	commitStatusAdapter, err := csu.createCommitStatusAdapterForSnapshot(report)
-	if err != nil {
-		csu.logger.Error(err, "failed to create CommitStatusAdapter for scenario, skipping update",
-			"snapshot.NameSpace", csu.snapshot.Namespace, "snapshot.Name", csu.snapshot.Name,
-			"scenario.Name", report.ScenarioName,
-		)
-		return nil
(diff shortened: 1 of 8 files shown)


chipspeak (approved):
LGTM

status/reporter_github.go
```diff
@@ -417,49 +417,56 @@ func (csu *CommitStatusUpdater) updateStatusInComment(ctx context.Context, repor
...
-	if err != nil {
-		csu.logger.Error(err, "failed to create CommitStatusAdapter for scenario, skipping update",
-			"snapshot.NameSpace", csu.snapshot.Namespace, "snapshot.Name", csu.snapshot.Name,
-			"scenario.Name", report.ScenarioName,
-		)
-		return nil
-	}
+	sourceRepoOwner := gitops.GetSourceRepoOwnerFromSnapshot(csu.snapshot)
+	// we create/update commitStatus only when the source and target repo owner are the same
+	if csu.owner == sourceRepoOwner {
```
dirgim, line 422: In order to reduce the cognitive complexity, could we do something like this instead and exit the function early if source repo owner doesn't match the target owner (untested):
```
if csu.owner != sourceRepoOwner {
    // Create a comment when integration test is neither pending nor inprogress since comment for pending/inprogress is less meaningful and there is commitStatus for all statuses
	_, isPullRequest := csu.snapshot.GetAnnotations()[gitops.PipelineAsCodePullRequestAnnotation]
	if report.Status != intgteststat.IntegrationTestStatusPending && report.Status != intgteststat.IntegrationTestStatusInProgress && isPullRequest {
		err := csu.updateStatusInComment(ctx, report)
		if err != nil {
			csu.logger.Error(err, "failed to update comment", "snapshot.NameSpace", csu.snapshot.Namespace, "snapshot.Name", csu.snapshot.Name, "scenarioName", report.ScenarioName)
			return err
		}
    return nil
}
```
This way we'd also skip adding a comment if the owners match since then we can just update the commit status, wdyt?

hongweiliu17: We need to add comment for the test details test result no matter what the source repo owner is. Because commitStatus contains the status only like [this](https://gitlab.cee.redhat.com/service/ocm-backend-tests/-/merge_requests/955/pipelines), without details like [this comment](https://gitlab.cee.redhat.com/service/ocm-backend-tests/-/merge_requests/955#note_13612713) .
CommitStatus is not like Checkrun, it doesn't include the details for tasks.

dirgim: Ah, you are right, then this way makes the most sense, thanks.

dirgim (approved):
LGTM

PR #975, opened 2025-01-03
fix(STONEINTG-1110): fix for IsSnapshotCreatedByPACPushEvent

hongweiliu17:
fix func IsSnapshotCreatedByPACPushEvent for group snapshot

### DIFF
--- a/gitops/snapshot_test.go
+++ b/gitops/snapshot_test.go
@@ -844,6 +844,7 @@ var _ = Describe("Gitops functions for managing Snapshots", Ordered, func() {
 			})
 
 			It("Testing annotating snapshot", func() {
+				hasSnapshot.Labels[gitops.PipelineAsCodeEventTypeLabel] = gitops.PipelineAsCodePullRequestType
 				componentSnapshotInfos := []gitops.ComponentSnapshotInfo{
 					{
 						Component:        "com1",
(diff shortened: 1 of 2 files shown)


jencull (approved):

jencull:
lgtm

Josh-Everett (approved):

gitops/snapshot_test.go
```diff
@@ -862,6 +863,8 @@ var _ = Describe("Gitops functions for managing Snapshots", Ordered, func() {
 				Expect(err).ToNot(HaveOccurred())
 				Expect(componentSnapshotInfos).To(HaveLen(2))
 				Expect(snapshot.Labels[gitops.SnapshotTypeLabel]).To(Equal("group"))
+				Expect(gitops.IsSnapshotCreatedByPACPushEvent(snapshot)).To(BeFalse())
+				Expect(!gitops.IsSnapshotCreatedByPACPushEvent(snapshot)).To(BeTrue())
```
dheerajodha, line 867: Aren't these two lines testing the same thing?

Should we create a new `It()` block and test the condition when the group snapshot is missing the `PipelineAsCodeEventTypeLabel` and the `IsSnapshotCreatedByPACPushEvent()` returns `false`

hongweiliu17: Thanks for your question. Group snapshot has event-type: pull-request which is the same with the component snapshot which triggers the group snapshot creation. It is set before created. So I feel we don't need it. Do you agree?

dheerajodha: Thanks, that makes sense.
I do feel like the 2 assertion we're making above is redundant. Can we remove one of them?

hongweiliu17: You are right. Let me remove one. Actually I wanted to test the multiple conditions in this functions.

dheerajodha: Thanks!

dheerajodha (approved):
/lgtm

kasemAlem (approved):

PR #993, opened 2025-01-20
fix(STONEINTG-1087): retry longer to fetch App from ITS, and stop processing if not found

dheerajodha:
* There are some cases where the App and ITS are created almost at the same time.
* This causes a problem when the App is not found while the controllers reconcile ITS.
* To fix this, we increased the retry time from 10 ms to 10 seconds. This should give enough time to fetch the App.
* And if we're still unable to find App, we stop reconciling, since there's no point in reconciling an ITS without any parent.

### DIFF
--- a/internal/controller/scenario/scenario_controller.go
+++ b/internal/controller/scenario/scenario_controller.go
@@ -103,6 +103,8 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		return ctrl.Result{}, nil
 	}
 
+	logger = logger.WithApp(*application)
+
 	adapter := NewAdapter(ctx, application, scenario, logger, loader, r.Client)
 
 	return controller.ReconcileHandler([]controller.Operation{
--- a/internal/controller/scenario/scenario_controller.go
+++ b/internal/controller/scenario/scenario_controller.go
@@ -18,6 +18,7 @@ package scenario
 
 import (
 	"context"
+	"time"
 
 	"github.com/konflux-ci/integration-service/loader"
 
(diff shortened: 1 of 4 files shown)


internal/controller/scenario/scenario_controller.go
```diff
@@ -84,6 +92,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 	})
 	if err != nil {
 		logger.Error(err, "Failed to get Application from the IntegrationTestScenario after retry", "application", scenario.Spec.Application)
+		return ctrl.Result{}, nil
```
dheerajodha, line 95: Once we have retried long enough for the Application to exist in the cluster, and still couldn't find it, the operator will stop processing this ITS. But in that case, this ITS would be an orphan and dangle in the cluster. Do you think we should delete this ITS here?

dirgim: Hm, not sure, maybe we can at least mark it as invalid?

internal/controller/scenario/scenario_controller.go
```diff
@@ -75,15 +77,30 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
...
+		errString := "Failed to get Application from the IntegrationTestScenario even after retrying"
+		logger.Error(err, errString, "application", scenario.Spec.Application)
+		patch := client.MergeFrom(scenario.DeepCopy())
+		helpers.SetScenarioIntegrationStatusAsInvalid(scenario, errString)
+		err := r.Client.Status().Patch(ctx, scenario, patch)
+		if err != nil {
+			logger.Error(err, "Failed to update Scenario as Invalid")
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{}, nil
```
dheerajodha, line 103: Because we continue processing in case the Application is not found, I think we can remove [this block of code](https://github.com/konflux-ci/integration-service/blob/main/internal/controller/scenario/scenario_adapter.go#L72-L84) from the scenario_adapter.go file, wdyt?

dirgim: Yup, that makes sense to me.

dheerajodha: Thanks! Removed

dirgim (approved):
LGTM

kasemAlem (approved):

jsztuka (approved):
lgtm

PR #1020, opened 2025-02-13
feat(STONEINTG-1050): cancel old intg pipelineruns

jsztuka:

This is currently WIP, need to add unit tests for coverity. Reviews/feedback in a meantime are most welcome.

### DIFF
--- a/gitops/snapshot.go
+++ b/gitops/snapshot.go
@@ -204,6 +204,9 @@ const (
 	// AppStudioTestSucceededConditionFailed is the reason that's set when the AppStudio tests fail.
 	AppStudioTestSucceededConditionFailed = "Failed"
 
+	// AppStudioIntegrationStatusCanceled is the reason that's set when the AppStudio tests cancel.
+	AppStudioIntegrationStatusCanceled = "Canceled"
+
 	// AppStudioIntegrationStatusInvalid is the reason that's set when the AppStudio integration gets into an invalid state.
 	AppStudioIntegrationStatusInvalid = "Invalid"
 
@@ -215,6 +218,9 @@ const (
 
 	//AppStudioIntegrationStatusFinished is the reason that's set when the AppStudio tests finish.
 	AppStudioIntegrationStatusFinished = "Finished"
+
+	//AppStudioIntegrationStatusCancelled is the reason that's set when the AppStudio tests pipelinerun gets cancelled.
+	AppStudioIntegrationStatusCancelled = "CancelledRunFinally"
 
 	// the statuses needed to report to GiHub when creating check run or commit status, see doc
 	// https://docs.github.com/en/rest/guides/using-the-rest-api-to-interact-with-checks?apiVersion=2022-11-28
@@ -328,6 +334,28 @@ func MarkSnapshotAsFailed(ctx context.Context, adapterClient client.Client, snap
 		Type:    AppStudioTestSucceededCondition,
 		Status:  metav1.ConditionFalse,
 		Reason:  AppStudioTestSucceededConditionFailed,
+		Message: message,
+	}
+	meta.SetStatusCondition(&snapshot.Status.Conditions, condition)
+
+	err := adapterClient.Status().Patch(ctx, snapshot, patch)
+	if err != nil {
+		return err
+	}
+
+	snapshotCompletionTime := &metav1.Time{Time: time.Now()}
+	go metrics.RegisterCompletedSnapshot(condition.Type, condition.Reason, snapshot.GetCreationTimestamp(), snapshotCompletionTime)
+	return nil
+}
+
+// MarkSnapshotAsCanceled updates the AppStudio Test canceled condition for the Snapshot to 'Canceled'.
+// If the patch command fails, an error will be returned.
+func MarkSnapshotAsCanceled(ctx context.Context, adapterClient client.Client, snapshot *applicationapiv1alpha1.Snapshot, message string) error {
+	patch := client.MergeFrom(snapshot.DeepCopy())
+	condition := metav1.Condition{
+		Type:    AppStudioIntegrationStatusCondition,
+		Status:  metav1.ConditionTrue,
+		Reason:  AppStudioIntegrationStatusCanceled,
 		Message: message,
 	}
 	meta.SetStatusCondition(&snapshot.Status.Conditions, condition)
--- a/helpers/integration.go
+++ b/helpers/integration.go
@@ -31,6 +31,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 
 	"github.com/go-logr/logr"
+	intgteststat "github.com/konflux-ci/integration-service/pkg/integrationteststatus"
 	"github.com/santhosh-tekuri/jsonschema/v5"
 	tektonv1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"
 	"knative.dev/pkg/apis"
@@ -504,4 +505,19 @@ func IsObjectYoungerThanThreshold(obj metav1.Object, threshold time.Duration) bo
 	durationSinceObjectCreation := time.Since(objectCreationTime)
 
 	return durationSinceObjectCreation < threshold
+}
+
+// GetRunningIntegrationPipelineRunStatus checks the Tekton results for a given PipelineRun and returns status of test.
+// since we care only about in-progress state we return testpassed in case the test is not running at the moment
+func GetRunningIntegrationPipelineRunStatus(pipelineRun *tektonv1.PipelineRun) (intgteststat.IntegrationTestStatus, string, error) {
+	// Check if the pipelineRun finished from the condition of status
+	if !HasPipelineRunFinished(pipelineRun) {
+		// Mark the pipelineRun's status as "Deleted" if its not finished yet and is marked for deletion (with a non-nil deletionTimestamp)
+		if pipelineRun.GetDeletionTimestamp() != nil {
+			return intgteststat.IntegrationTestStatusDeleted, fmt.Sprintf("Integration test which is running as pipeline run '%s', has been deleted", pipelineRun.Name), nil
+		} else {
+			return intgteststat.IntegrationTestStatusInProgress, fmt.Sprintf("Integration test is running as pipeline run '%s'", pipelineRun.Name), nil
+		}
+	}
+	return intgteststat.IntegrationTestStatusTestPassed, "", nil
 }
--- a/internal/controller/snapshot/snapshot_adapter.go
+++ b/internal/controller/snapshot/snapshot_adapter.go
@@ -179,6 +179,13 @@ func (a *Adapter) rerunIntegrationPipelinerunForScenario(scenario *v1beta2.Integ
 	if err != nil {
 		return a.HandlePipelineCreationError(err, scenario, testStatuses)
 	}
+	if metadata.HasLabelWithValue(a.snapshot, gitops.PipelineAsCodeEventTypeLabel, gitops.PipelineAsCodePullRequestType) {
+		err = a.checkAndCancelOldSnapshotsPipelineRun(a.application, a.snapshot)
+		if err != nil {
+			a.logger.Error(err, "Failed to check and cancel old snapshot's pipelineruns",
+				"snapshot.Name:", a.snapshot.Name)
+		}
+	}
 
 	testStatuses.UpdateTestStatusIfChanged(scenario.Name, intgteststat.IntegrationTestStatusInProgress, fmt.Sprintf("PipelineRun '%s' created", pipelineRun.Name))
 	if err := testStatuses.UpdateTestPipelineRunName(scenario.Name, pipelineRun.Name); err != nil {
@@ -297,6 +304,13 @@ func (a *Adapter) EnsureIntegrationPipelineRunsExist() (controller.OperationResu
 					// it doesn't make sense to restart reconciliation here, it will be eventually updated by integrationpipeline adapter
 					a.logger.Error(err, "Failed to update pipelinerun name in test status")
 				}
+			}
+		}
+		if metadata.HasLabelWithValue(a.snapshot, gitops.PipelineAsCodeEventTypeLabel, gitops.PipelineAsCodePullRequestType) {
+			err = a.checkAndCancelOldSnapshotsPipelineRun(a.application, a.snapshot)
+			if err != nil {
+				a.logger.Error(err, "Failed to check and cancel old snapshot's pipelineruns",
+					"snapshot.Name:", a.snapshot.Name)
 			}
 		}
 
@@ -1057,4 +1071,70 @@ func (a *Adapter) findSnapshotWithOpenedPR(snapshot *applicationapiv1alpha1.Snap
 		}
 	}
 	return nil, nil
+}
+
+// checkAndCancelOldSnapshotsPipelineRun sorts all snapshots for application and cancels all running integrationTest pipelineruns within application
+// removes finalizer before the pipelinerun is set as CancelledRunFinally to be gracefully cancelled
+func (a *Adapter) checkAndCancelOldSnapshotsPipelineRun(application *applicationapiv1alpha1.Application, snapshot *applicationapiv1alpha1.Snapshot) error {
+
+	snapshots, err := a.loader.GetAllSnapshotsForPR(a.context, a.client, application, snapshot.GetLabels()[gitops.PipelineAsCodePullRequestAnnotation], snapshot.GetLabels()[gitops.PipelineAsCodeRepoURLAnnotation])
+	if err != nil {
+		a.logger.Error(err, "Failed to fetch Snapshots for the application",
+			"application.Name:", application.Name)
+		return err
+	}
+	sortedSnapshots := gitops.SortSnapshots(*snapshots)
+	if len(sortedSnapshots) < 2 {
+		return nil
+	}
+
+	for i := 1; i < len(sortedSnapshots); i++ {
+		if gitops.IsComponentSnapshot(&sortedSnapshots[i]) {
+			err = gitops.MarkSnapshotAsCanceled(a.context, a.client, &sortedSnapshots[i], "Snapshot canceled/superseded")
+			if err != nil {
+				a.logger.Error(err, "Failed to mark snapshot as canceled", "snapshot.Name", &sortedSnapshots[i].Name)
+				return err
+			}
+			err = a.cancelAllPipelineRunsForSnapshot(&sortedSnapshots[i])
+		}
+	}
+	return err
+}
+
+// cancelAllPipelineRunsForSnapshot gets all integration test pipelieruns for a given snapshot
+func (a *Adapter) cancelAllPipelineRunsForSnapshot(snapshot *applicationapiv1alpha1.Snapshot) error {
+	// get all integration pipelineruns for a snapshot
+	integrationTestPipelineruns, err := a.loader.GetAllIntegrationPipelineRunsForSnapshot(a.context, a.client, snapshot)
+	if err != nil {
+		a.logger.Error(err, "Failed to get all integration pipelineruns for snapshot", "snapshot.Name", snapshot.Name)
+		return err
+	}
+	if len(integrationTestPipelineruns) < 1 {
+		a.logger.Info("No integrationTest pipelineruns were found for snapshot", "snapshot.Name", snapshot.Name)
+		return nil
+	}
+	for _, plr := range integrationTestPipelineruns {
+		plr := plr
+		pipelinerunStatus, _, err := h.GetRunningIntegrationPipelineRunStatus(&plr)
+		if err != nil {
+			a.logger.Error(err, "Failed to get integration test pipelinerun status", "snapshot.Name", plr.Name)
+			return err
+		}
+		if pipelinerunStatus == intgteststat.IntegrationTestStatusInProgress {
+			//remove finalizer and cancel pipelinerun
+			err = h.RemoveFinalizerFromPipelineRun(a.context, a.client, a.logger, &plr, h.IntegrationPipelineRunFinalizer)
+			if err != nil {
+				return err
+			}
+			// set "CancelledRunFinally" to PLR status, should gracefully cancel pipelinerun, this is so raw I hate this
+			patch := client.MergeFrom(plr.DeepCopy())
+			plr.Spec.Status = tektonv1.PipelineRunSpecStatusCancelledRunFinally
+			err := a.client.Patch(a.context, &plr, patch)
+			if err != nil {
+				return err
+			}
+			a.logger.LogAuditEvent("IntegrationTestscenario pipelineRun has been cancelled.", &plr, h.LogActionUpdate)
+		}
+	}
+	return nil
 }
--- a/internal/controller/snapshot/snapshot_adapter_test.go
+++ b/internal/controller/snapshot/snapshot_adapter_test.go
@@ -18,7 +18,6 @@ package snapshot
 
 import (
 	"bytes"
-	"context"
 	"fmt"
 	"reflect"
 	"strconv"
@@ -54,7 +53,6 @@ import (
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/types"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
 var _ = Describe("Snapshot Adapter", Ordered, func() {
@@ -71,6 +69,8 @@ var _ = Describe("Snapshot Adapter", Ordered, func() {
 		hasCom3                                   *applicationapiv1alpha1.Component
 		hasSnapshot                               *applicationapiv1alpha1.Snapshot
 		hasSnapshotPR                             *applicationapiv1alpha1.Snapshot
+		hasOldCompSnapshotPR                      *applicationapiv1alpha1.Snapshot
+		hasNewCompSnapshotPR                      *applicationapiv1alpha1.Snapshot
 		hasOverRideSnapshot                       *applicationapiv1alpha1.Snapshot
 		hasInvalidSnapshot                        *applicationapiv1alpha1.Snapshot
 		hasInvalidOverrideSnapshot                *applicationapiv1alpha1.Snapshot
@@ -81,6 +81,7 @@ var _ = Describe("Snapshot Adapter", Ordered, func() {
 		integrationTestScenario1                  *v1beta2.IntegrationTestScenario
 		integrationTestScenarioForInvalidSnapshot *v1beta2.IntegrationTestScenario
 		buildPipelineRun1                         *tektonv1.PipelineRun
+		integrationPipelineRunOld                 *tektonv1.PipelineRun
 	)
 	const (
 		SampleRepoLink      = "https://github.com/devfile-samples/devfile-sample-java-springboot-basic"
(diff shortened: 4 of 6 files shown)


gitops/snapshot.go
```diff
@@ -1044,6 +1075,22 @@ func GetPRGroupFromSnapshot(snapshot *applicationapiv1alpha1.Snapshot) (string,
 	return "", ""
 }
 
+// GetPipelineAsCodeRepoURLAnnotation gets the value of annotaiton "pac.test.appstudio.openshift.io/repo-url"
```
hongweiliu17, line 1078: I think you may combine these two functions by passing the annotation name as a parameter

jsztuka: That was also my idea, good to see you have the same opinion on that. Will change it

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -1058,3 +1068,122 @@ func (a *Adapter) findSnapshotWithOpenedPR(snapshot *applicationapiv1alpha1.Snap
...
+				for _, plr := range integrationTestPipelineruns {
+					plr := plr
+					pipelinerunStatus, _, err := h.GetRunningIntegrationPipelineRunStatus(a.context, a.client, &plr)
+					if err != nil {
+						a.logger.Error(err, "Failed to get integration test pipelinerun status", "snapshot.Name", plr.Name)
+						return err
+					}
+					if pipelinerunStatus == intgteststat.IntegrationTestStatusInProgress {
+						//remove finalizer and cancel pipelinerun
+						err = h.RemoveFinalizerFromPipelineRun(a.context, a.client, a.logger, &plr, h.IntegrationPipelineRunFinalizer)
```
jsztuka, line 1115: I think we should check here whether we are removing finalizers from GH or GL, wdyt?

gitops/snapshot.go
```diff
@@ -342,6 +351,28 @@ func MarkSnapshotAsFailed(ctx context.Context, adapterClient client.Client, snap
 	return nil
 }
 
+// MarkSnapshotAsCanceled updates the AppStudio Test canceled condition for the Snapshot to canceled.
+// If the patch command fails, an error will be returned.
+func MarkSnapshotAsCanceled(ctx context.Context, adapterClient client.Client, snapshot *applicationapiv1alpha1.Snapshot, message string) error {
+	patch := client.MergeFrom(snapshot.DeepCopy())
+	condition := metav1.Condition{
+		Type:    AppStudioTestCanceledCondition,
```
dirgim, line 359: I'm thinking it might be better to use the existing `AppStudioIntegrationStatusCondition` here instead of creating a new condition, wdyt?

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -1058,3 +1068,122 @@ func (a *Adapter) findSnapshotWithOpenedPR(snapshot *applicationapiv1alpha1.Snap
 	}
 	return nil, nil
 }
+
+// checkOldSnapshotsPipelineRun sorts all snapshots for application and cancells all running interationTest pipelineruns within application
+// removes finalizer before the pipelinerun is set as cancelled run finally to be gracefuly cancelled
+// god please forgive me for this monstrosity of cognitive complexity 40+
+func (a *Adapter) checkAndCancelOldSnapshotsPipelineRun(application *applicationapiv1alpha1.Application, snapshot *applicationapiv1alpha1.Snapshot) error {
+	snapshots, err := a.loader.GetAllSnapshots(a.context, a.client, application)
```
dirgim, line 1076: I think it would be a good optimization here to add a new function to loader which gets all Snapshots for a given PR, similar to how it's done in the [GetAllSnapshotsForBuildPipelineRun](https://github.com/konflux-ci/integration-service/blob/9b2d7160579f1a24bf20dc186368078916af03de/loader/loader.go#L334-L350) function, just with different labels. This way we wouldn't have to loop through all of the users Snapshots.

gitops/snapshot.go
```diff
@@ -1044,6 +1075,14 @@ func GetPRGroupFromSnapshot(snapshot *applicationapiv1alpha1.Snapshot) (string,
 	return "", ""
 }
 
+// GetPipelineAsCodeAnnotation gets the value of input annotaiton
+func GetPipelineAsCodeAnnotation(snapshot *applicationapiv1alpha1.Snapshot, annotation string) string {
```
dirgim, line 1079: Alternatively, we could use something like [this](https://github.com/konflux-ci/integration-service/blob/9b2d7160579f1a24bf20dc186368078916af03de/status/reporter_github.go#L373) to safely fetch an annotation.

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -287,6 +292,11 @@ func (a *Adapter) EnsureIntegrationPipelineRunsExist() (controller.OperationResu
 					if !clienterrors.IsInvalid(err) {
 						errsForPLRCreation = errors.Join(errsForPLRCreation, err)
 					}
+					err = a.checkAndCancelOldSnapshotsPipelineRun(a.application, a.snapshot)
```
dirgim, line 295: If we're cancelling all integration pipelines for a given Snapshot, I think this line should be called after we've ensured that the new pipelineRuns are all running - so not within this `for` loop, but after it.

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -1058,3 +1068,122 @@ func (a *Adapter) findSnapshotWithOpenedPR(snapshot *applicationapiv1alpha1.Snap
 	}
 	return nil, nil
 }
+
+// checkOldSnapshotsPipelineRun sorts all snapshots for application and cancells all running interationTest pipelineruns within application
+// removes finalizer before the pipelinerun is set as cancelled run finally to be gracefuly cancelled
+// god please forgive me for this monstrosity of cognitive complexity 40+
+func (a *Adapter) checkAndCancelOldSnapshotsPipelineRun(application *applicationapiv1alpha1.Application, snapshot *applicationapiv1alpha1.Snapshot) error {
```
dirgim, line 1075: We should also first check if this is a PR Snapshot before doing anything in this function (or maybe before even calling it) because otherwise we'd just be wasting resources by fetching Snapshots and looping through them.

gitops/snapshot.go
jsztuka: Changed to the existing condition.

internal/controller/snapshot/snapshot_adapter.go
jsztuka: Agreed, getting all snapshots is not efficient way here. Created new method just for this purpose.

gitops/snapshot.go
jsztuka: Updated with recommended approach.

internal/controller/snapshot/snapshot_adapter.go
jsztuka: Good point, moved this part after the loop.

internal/controller/snapshot/snapshot_adapter.go
jsztuka: Agreed, set a condition before the function gets executed.

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -179,6 +179,13 @@ func (a *Adapter) rerunIntegrationPipelinerunForScenario(scenario *v1beta2.Integ
 	if err != nil {
 		return a.HandlePipelineCreationError(err, scenario, testStatuses)
 	}
+	if a.snapshot.GetLabels()["pac.test.appstudio.openshift.io/event-type"] == "pull-request" {
```
hongweiliu17, line 182: Using `HasLabelWithValue(a.snapshot, gitops.PipelineAsCodeEventTypeLabel, gitops.PipelineAsCodePullRequestType)`, `pull_request` not `pull-request`?

jsztuka: Like this solution better, thanks!

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -1058,3 +1072,123 @@ func (a *Adapter) findSnapshotWithOpenedPR(snapshot *applicationapiv1alpha1.Snap
...
+				}
+				for _, plr := range integrationTestPipelineruns {
+					plr := plr
+					pipelinerunStatus, _, err := h.GetRunningIntegrationPipelineRunStatus(a.context, a.client, &plr)
+					if err != nil {
+						a.logger.Error(err, "Failed to get integration test pipelinerun status", "snapshot.Name", plr.Name)
+						return err
+					}
+					if pipelinerunStatus == intgteststat.IntegrationTestStatusInProgress {
+						//remove finalizer and cancel pipelinerun
```
hongweiliu17, line 1159: Why delete the finalizer for the in progress PLR only? I think, once new PLR is triggered, then all older PLRs should not be kept. HDYT?

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -1058,3 +1072,123 @@ func (a *Adapter) findSnapshotWithOpenedPR(snapshot *applicationapiv1alpha1.Snap
...
+		a.logger.Error(err, "Failed to fetch Snapshots for the application",
+			"application.Name:", application.Name)
+		return err
+	}
+	sortedSnapshots := gitops.SortSnapshots(*snapshots)
+	// check for older snapshots, their current status and type
+	latestSnapshot := sortedSnapshots[0]
+	for i := 1; i < len(sortedSnapshots); i++ {
+
+		if gitops.IsGroupSnapshot(&sortedSnapshots[i]) {
```
hongweiliu17, line 1092: Currently we don't have pr number in `group` snapshot because `group` snapshot can be created for one or multiple PRs, so the returned snapshot in `GetAllSnapshotsForGivenPR` won't include `group` snapshot

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -1058,3 +1072,123 @@ func (a *Adapter) findSnapshotWithOpenedPR(snapshot *applicationapiv1alpha1.Snap
...
+		if gitops.IsComponentSnapshot(&sortedSnapshots[i]) {
+
+			if snapshot.GetAnnotations()[gitops.PipelineAsCodeRepoURLAnnotation] == sortedSnapshots[i].GetAnnotations()[gitops.PipelineAsCodeRepoURLAnnotation] && snapshot.GetAnnotations()[gitops.PipelineAsCodePullRequestAnnotation] == sortedSnapshots[i].GetAnnotations()[gitops.PipelineAsCodePullRequestAnnotation] {
+				err = gitops.MarkSnapshotAsCanceled(a.context, a.client, &sortedSnapshots[i], "Snapshot canceled/superseded")
+				if err != nil {
+					a.logger.Error(err, "Failed to mark snapshot as canceled", "snapshot.Name", snapshot.Name)
+					return err
+				}
+				// get all integration pipelineruns for a snapshot
+				integrationTestPipelineruns, err := getAllIntegrationPipelineRunsForSnapshot(a.context, a.client, &sortedSnapshots[i])
```
hongweiliu17, line 1146: I feel you may get the integrationPipelinerunsForPR out of the loop, instead of getting integrationPipelineruns for each snapshot. HDYT?

internal/controller/snapshot/snapshot_adapter.go
jsztuka: This should be the older pipelinerun, which is still running and we dont need it anymore.

internal/controller/snapshot/snapshot_adapter.go
jsztuka: Is there any other label for group snapshots, that we could use for determining which we can cancel?

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -1058,3 +1072,123 @@ func (a *Adapter) findSnapshotWithOpenedPR(snapshot *applicationapiv1alpha1.Snap
...
+							return err
+						}
+
+					}
+				}
+			}
+		}
+		if gitops.IsComponentSnapshot(&sortedSnapshots[i]) {
+
+			if snapshot.GetAnnotations()[gitops.PipelineAsCodeRepoURLAnnotation] == sortedSnapshots[i].GetAnnotations()[gitops.PipelineAsCodeRepoURLAnnotation] && snapshot.GetAnnotations()[gitops.PipelineAsCodePullRequestAnnotation] == sortedSnapshots[i].GetAnnotations()[gitops.PipelineAsCodePullRequestAnnotation] {
```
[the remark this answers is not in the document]
jsztuka: Thanks, yes, that will be more efficient.

internal/controller/snapshot/snapshot_adapter.go
jsztuka: I'm not sure if I follow, we get all integration pipelineruns for a single snapshot, then we check for the ones that are running and cancel them.

hongweiliu17: If I understand correctly, we only keep the latest component snapshot and its integration pipelinerun for the PR, is it?

internal/controller/snapshot/snapshot_adapter.go
hongweiliu17: If you want to get the group snapshot created for that, I feel you may query with the label `test.appstudio.openshift.io/pr-group-sha`

internal/controller/snapshot/snapshot_adapter.go
hongweiliu17: I think the completed older integration PLRs are not needed any more. Ignore this comment. Thanks.

hongweiliu17: This makes me think we may ignore this type of cancelled snapshot in statusreport controller or snapshot controller as well in epic [STONEINTG-1049](https://issues.redhat.com/browse/STONEINTG-1049), HDYT? cc @dirgim

internal/controller/snapshot/snapshot_adapter.go
dirgim: Repo url is not enough to differentiate Snapshots, we'd start cancelling tests for unrelated PRs.
I'd prefer not to cancel group Snapshots rather than doing that.

internal/controller/snapshot/snapshot_adapter.go
dirgim: That's an interesting idea, it might be beneficial to skip processing cancelled Snapshots in the statusreport controller at least.

internal/controller/snapshot/snapshot_adapter.go
jsztuka: Yes, correct.

internal/controller/snapshot/snapshot_adapter.go
dirgim: I don't agree with that approach, getting Snapshots for a given repo is not nearly precise enough for us to decide to cancel running integration tests.
Let's say we open a PR with a change in the repo and our integration tests start running for an hour.
Then dependabot opens a new PR to bump the version of an image - this is a separate PR and shouldn't cause our running CI to be cancelled.

dirgim: Maybe it would be a good idea to have a follow-up story to cover group test Snapshots instead of trying to cover it in this change?

jsztuka: Sounds right. Lets drop the group snapshots part for this change and create follow up story that will take care of that.

jsztuka: Removed the group snapshots logic, together with unit tests, lets see what change this brings to e2e tests.

internal/controller/snapshot/snapshot_adapter.go
hongweiliu17: Then we just need the latest PR snapshot and its latest integration test, others ps snapshot and integration test are not needed.

jsztuka: I see the poit, I'm just not sure how else we could cancel the rest of the pipelineruns without getting all of them.

jsztuka: I was under asumption that the latest snapshot also contains the latest PLR, thats why I never check for both of those and I cancel all others snapshots/PLRs. You think that current behavior is not following this logic as it should?

jsztuka: Added the logic, let me know if that fullfills the requirements.

jsztuka:
Parying for the CI and e2e test gods to let it pass this time.

dheerajodha:
> Parying for the CI and e2e test gods to let it pass this time.

/meow

loader/loader.go
```diff
@@ -349,6 +350,25 @@ func (l *loader) GetAllSnapshotsForBuildPipelineRun(ctx context.Context, c clien
 	return &snapshots.Items, nil
 }
 
+// GetAllSnapshotsForGivenRepo returns all Snapshots for the associated Pull Request for the same repo url.
+// In the case the List operation fails, an error will be returned.
+// gitops.PipelineAsCodePullRequestAnnotation is also a label
+func (l *loader) GetAllSnapshotsForGivenRepo(ctx context.Context, c client.Client, application *applicationapiv1alpha1.Application, repoUrl string) (*[]applicationapiv1alpha1.Snapshot, error) {
```
dirgim, line 356: I still think that this function should be more focused and changed into something like `GetAllSnapshotsForPR` or similar. Basically it would also need to filter on the PR number annotation.

The reason for this is that for a lot of users, this query will return all of their Snapshots, meaning that it would be a more heavy API call, and also we'd have to loop through all of them.

jsztuka: Changed, so the function filters only the snapshots with a given PR.

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -1058,3 +1072,86 @@ func (a *Adapter) findSnapshotWithOpenedPR(snapshot *applicationapiv1alpha1.Snap
 	}
 	return nil, nil
 }
+
+// checkOldSnapshotsPipelineRun sorts all snapshots for application and cancells all running interationTest pipelineruns within application
+// removes finalizer before the pipelinerun is set as cancelled run finally to be gracefuly cancelled
+// god please forgive me for this monstrosity of cognitive complexity 40+
```
dirgim, line 1078: Hopefully it should be now way lower than 40 after adjustments :laughing:

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -1058,3 +1072,86 @@ func (a *Adapter) findSnapshotWithOpenedPR(snapshot *applicationapiv1alpha1.Snap
...
+
+				}
+			}
+
+		}
+	}
+	return err
+}
+
+func getAllIntegrationPipelineRunsForSnapshot(ctx context.Context, adapterClient client.Client, snapshot *applicationapiv1alpha1.Snapshot) ([]tektonv1.PipelineRun, error) {
```
dirgim, line 1145: Not sure but I'm thinking if should this function be moved to loader?

internal/controller/snapshot/snapshot_adapter.go
jsztuka: Let me remove this, its 39 now :(.

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -1058,3 +1072,86 @@ func (a *Adapter) findSnapshotWithOpenedPR(snapshot *applicationapiv1alpha1.Snap
...
+		return err
+	}
+	sortedSnapshots := gitops.SortSnapshots(*snapshots)
+	if len(sortedSnapshots) < 1 {
+		a.logger.Info("There are no found snapshots for given repo at the moment.")
+		return nil
+	}
+	// check for older snapshots, their current status and type
+	for i := 1; i < len(sortedSnapshots); i++ {
+		if gitops.IsComponentSnapshot(&sortedSnapshots[i]) {
```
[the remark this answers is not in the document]
jsztuka: I tried to prevent this by starting the for loop from 1 and not 0( which is the latest snapshot)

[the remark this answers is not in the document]
dirgim: After thinking on this some more, I think that the current solution should work well enough,

internal/controller/snapshot/snapshot_adapter.go
jsztuka: Agreed, moved to loader.

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -1058,3 +1072,77 @@ func (a *Adapter) findSnapshotWithOpenedPR(snapshot *applicationapiv1alpha1.Snap
...
+func (a *Adapter) checkAndCancelOldSnapshotsPipelineRun(application *applicationapiv1alpha1.Application, snapshot *applicationapiv1alpha1.Snapshot) error {
+
+	snapshots, err := a.loader.GetAllSnapshotsForPR(a.context, a.client, application, snapshot.GetLabels()[gitops.PipelineAsCodePullRequestAnnotation], snapshot.GetLabels()[gitops.PipelineAsCodeRepoURLAnnotation])
+	if err != nil {
+		a.logger.Error(err, "Failed to fetch Snapshots for the application",
+			"application.Name:", application.Name)
+		return err
+	}
+	sortedSnapshots := gitops.SortSnapshots(*snapshots)
+	if len(sortedSnapshots) < 1 {
```
dirgim, line 1087: Just realized that maybe this should be checking `if len(sortedSnapshots) < 2`.
If we have just one Snapshot (let's say the current one) in the list, and the `for` loop below starts looping from the index 1, that could cause an index out of range error.

dirgim: Never mind, I just tested it out and the initial condition in the for loop has been well set and catches it correctly.

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -1058,3 +1072,77 @@ func (a *Adapter) findSnapshotWithOpenedPR(snapshot *applicationapiv1alpha1.Snap
...
+		a.logger.Info("There are no found snapshots for given repo at the moment.")
```
dheerajodha, line 1088:
```suggestion
		a.logger.Info("There are no snapshots found for given repo at the moment.")
```
nit: optional

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -1058,3 +1072,77 @@ func (a *Adapter) findSnapshotWithOpenedPR(snapshot *applicationapiv1alpha1.Snap
...
+		return nil
+	}
+	// check for older snapshots, their current status and type
+	for i := 1; i < len(sortedSnapshots); i++ {
+		if gitops.IsComponentSnapshot(&sortedSnapshots[i]) {
+			err = gitops.MarkSnapshotAsCanceled(a.context, a.client, &sortedSnapshots[i], "Snapshot canceled/superseded")
+			if err != nil {
+				a.logger.Error(err, "Failed to mark snapshot as canceled", "snapshot.Name", snapshot.Name)
```
dheerajodha, line 1096:
```suggestion
				a.logger.Error(err, "Failed to mark snapshot as canceled", "snapshot.Name", &sortedSnapshots[i].Name)
```
I think it should be this, so we use the snapshot we're iterating on rather than the one we're reconciling currently.

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -1058,3 +1072,77 @@ func (a *Adapter) findSnapshotWithOpenedPR(snapshot *applicationapiv1alpha1.Snap
...
+			}
+			// get all integration pipelineruns for a snapshot
+			err = a.cancelAllPipelineRunsForSnapshot(&sortedSnapshots[i])
+
+		}
+	}
+	return err
+}
+
+func (a *Adapter) cancelAllPipelineRunsForSnapshot(snapshot *applicationapiv1alpha1.Snapshot) error {
```
dheerajodha, line 1107: Can we name this function in a way that shows we are only canceling all except the latest PLR? Also, can we please have a header comment for this function to briefly describe what it's doing?

internal/controller/snapshot/snapshot_adapter_test.go
```diff
@@ -2259,18 +2346,88 @@ var _ = Describe("Snapshot Adapter", Ordered, func() {
...
+			Expect(result.CancelRequest).To(BeFalse())
+			Expect(result.RequeueRequest).To(BeFalse())
+			foundStatusCondition := meta.FindStatusCondition(hasOldCompSnapshotPR.Status.Conditions, gitops.AppStudioIntegrationStatusCondition)
+			Expect(foundStatusCondition.Reason).To(Equal(gitops.AppStudioIntegrationStatusCanceled))
+			Expect(err).ToNot(HaveOccurred())
+			statuses, err := gitops.NewSnapshotIntegrationTestStatusesFromSnapshot(hasOldCompSnapshotPR)
+			Expect(err).ToNot(HaveOccurred())
+			detail, ok := statuses.GetScenarioStatus(integrationTestScenario.Name)
+			Expect(ok).To(BeTrue())
+			Expect(detail.Status).To(Equal(intgteststat.IntegrationTestStatusInProgress))
```
dheerajodha, line 2417: Curious, why is this integration PLR still in-progress? It belongs to the older cancelled Snapshot, so shouldn't this be in a cancelled state too? I'm sure I'm missing something.

internal/controller/snapshot/snapshot_adapter.go
dheerajodha: Hmm, we're calling this function for older snapshots, so do we really need to not cancel the latest PLR?

internal/controller/snapshot/snapshot_adapter.go
jsztuka: That is a good point, more than one snapshots is required for this function to work correctly.

gitops/snapshot.go
```diff
@@ -342,6 +349,28 @@ func MarkSnapshotAsFailed(ctx context.Context, adapterClient client.Client, snap
 	return nil
 }
 
+// MarkSnapshotAsCanceled updates the AppStudio Test canceled condition for the Snapshot to canceled.
+// If the patch command fails, an error will be returned.
+func MarkSnapshotAsCanceled(ctx context.Context, adapterClient client.Client, snapshot *applicationapiv1alpha1.Snapshot, message string) error {
+	patch := client.MergeFrom(snapshot.DeepCopy())
+	condition := metav1.Condition{
+		Type:    AppStudioIntegrationStatusCondition,
+		Status:  metav1.ConditionTrue,
```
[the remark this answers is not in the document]
jsztuka: by definition:
`These are valid condition statuses. "ConditionTrue" means a resource is in the condition. "ConditionFalse" means a resource is not in the condition. "ConditionUnknown" means kubernetes can't decide if a resource is in the condition or not. In the future, we could add other intermediate conditions, e.g. ConditionDegraded.`
Looks like we want true there.

gitops/snapshot.go
```diff
@@ -1069,6 +1098,20 @@ func SortSnapshots(snapshots []applicationapiv1alpha1.Snapshot) []applicationapi
 	return snapshots
 }
 
+// SortPipelineruns sorts the pipelineruns according to the snapshot annotation CreationTimestamp
```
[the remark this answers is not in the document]
jsztuka: Correct, the definition needs to be updated.

internal/controller/snapshot/snapshot_adapter_test.go
jsztuka: What I tried to achieve there, is that I didn't want to affect other integration tests, that are not connected to the snapshot/ component, so this one should not be affected. Hence the status, I can remove this if this makes little sense.

internal/controller/snapshot/snapshot_adapter.go
dheerajodha: Thanks for updating. I'm still curious about why we are calling this function on an older snapshot because it now cancels all but the latest Intg PLR for that older snapshot. Do we want that? Or do we want to cancel **ALL** the Intg PLRs for the older snapshots?

gitops/snapshot.go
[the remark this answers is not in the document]
dheerajodha: Thanks for the explanation good sir

internal/controller/snapshot/snapshot_adapter_test.go
dheerajodha: that's a good point, I misunderstood between spec.status VS status in snapshot annotation

internal/controller/snapshot/snapshot_adapter.go
dheerajodha: thanks for the update!

gitops/snapshot.go
```diff
@@ -1069,6 +1098,20 @@ func SortSnapshots(snapshots []applicationapiv1alpha1.Snapshot) []applicationapi
...
+		// represents the start time of build PLR
+		// when CreationTimestamp is not set, the value of Atoi is 0
+		time_i, _ := strconv.Atoi(pipelineruns[i].CreationTimestamp.String())
+		time_j, _ := strconv.Atoi(pipelineruns[j].CreationTimestamp.String())
+
+		return time_i > time_j
+	})
+	return pipelineruns
+}
+
```
dirgim, line 1114: This function can probably be removed since it should be unused now.

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -1058,3 +1072,68 @@ func (a *Adapter) findSnapshotWithOpenedPR(snapshot *applicationapiv1alpha1.Snap
...
+				a.logger.Error(err, "Failed to remove finalizer from integration test pipelinerun")
+				return err
+			}
+			// set "CancelledRunFinally" to PLR status, should gracefully cancel pipelinerun, this is so raw I hate this
+			patch := client.MergeFrom(plr.DeepCopy())
+			plr.Spec.Status = tektonv1.PipelineRunSpecStatusCancelledRunFinally
+			err := a.client.Patch(a.context, &plr, patch)
+			if err != nil {
+				return err
+			}
```
dirgim, line 1135: nit: since we're updating(patching) the pipelineRun here, it might be useful to add an audit log call here, similar to how we emit one when [creating the pipelineRun](https://github.com/konflux-ci/integration-service/blob/c23fc034faec01744e823881285f265196f16a4b/internal/controller/snapshot/snapshot_adapter.go#L777-L778).
That said, we already emit an audit log when [removing the finalizer](https://github.com/konflux-ci/integration-service/blob/c23fc034faec01744e823881285f265196f16a4b/helpers/integration.go#L433), so I'd also be fine with not adding another one.

dirgim (approved):
LGTM, nice work on this!

jencull (approved):
lgtm

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -1058,3 +1072,69 @@ func (a *Adapter) findSnapshotWithOpenedPR(snapshot *applicationapiv1alpha1.Snap
...
+		a.logger.Error(err, "Failed to get all integration pipelineruns for snapshot", "snapshot.Name", snapshot.Name)
+		return err
+	}
+	if len(integrationTestPipelineruns) < 1 {
+		a.logger.Info("No integrationTest pipelineruns were found for snapshot", "snapshot.Name", snapshot.Name)
+		return nil
+	}
+	for _, plr := range integrationTestPipelineruns {
+		plr := plr
+		pipelinerunStatus, _, err := h.GetRunningIntegrationPipelineRunStatus(&plr)
```
hongweiliu17, line 1118: The second returned value from `GetRunningIntegrationPipelineRunStatus` are not used here. How about using `!HasPipelineRunFinished(pipelineRun)` directly?

helpers/integration.go
```diff
@@ -505,3 +506,18 @@ func IsObjectYoungerThanThreshold(obj metav1.Object, threshold time.Duration) bo
 
 	return durationSinceObjectCreation < threshold
 }
+
+// GetRunningIntegrationPipelineRunStatus checks the Tekton results for a given PipelineRun and returns status of test.
+// since we care only about in-progress state we return testpassed in case the test is not running at the moment
+func GetRunningIntegrationPipelineRunStatus(pipelineRun *tektonv1.PipelineRun) (intgteststat.IntegrationTestStatus, string, error) {
```
hongweiliu17, line 512: The third returned `err` always is `nil`? so it can be removed.

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -1058,3 +1072,69 @@ func (a *Adapter) findSnapshotWithOpenedPR(snapshot *applicationapiv1alpha1.Snap
...
+		if err != nil {
```
hongweiliu17, line 1119: As I commented above, the err is always `nil`, so this part can be removed.

loader/loader.go
```diff
@@ -486,3 +508,18 @@ func (l *loader) GetMatchingComponentSnapshotsForPRGroupHash(ctx context.Context
...
+func (l *loader) GetAllIntegrationPipelineRunsForSnapshot(ctx context.Context, adapterClient client.Client, snapshot *applicationapiv1alpha1.Snapshot) ([]tektonv1.PipelineRun, error) {
+	integrationPipelineRuns := &tektonv1.PipelineRunList{}
+	opts := []client.ListOption{
+		client.InNamespace(snapshot.Namespace),
+		client.MatchingLabels{
+			"pipelines.appstudio.openshift.io/type": "test",
+			"appstudio.openshift.io/snapshot":       snapshot.Name,
+		},
+	}
+	err := adapterClient.List(ctx, integrationPipelineRuns, opts...)
```
hongweiliu17, line 522: Should we check `if err != nil` before `return integrationPipelineRuns.Items, err`?

hongweiliu17 (commented):
I know the PR has been merged, but left some comment for your information. Thanks for your nice work!

internal/controller/snapshot/snapshot_adapter.go
jsztuka: Would not we get also pending ones? Or other statuses that we are not interested in?

hongweiliu17: Just feel pending status mean it doesn't finish. We can keep the current. Thanks

PR #1049, opened 2025-03-05
feat(STONEINTG-1131): initialize pending intg status for group ITS

hongweiliu17:
* set intg status to pending for group context ITS if group snapshot is expected
* get builds PLR and snapshots for the pr group in build PLR to check if multiple component is affected by the same pr group

### DIFF
--- a/gitops/snapshot.go
+++ b/gitops/snapshot.go
@@ -242,6 +242,8 @@ const (
 
 	//IntegrationTestStatusCancelledGithub is the status reported to github when integration test is cancelled
 	IntegrationTestStatusCancelledGithub = "cancelled"
+
+	ComponentNameForGroupSnapshot = "pr group"
 )
 
 var (
@@ -974,6 +976,31 @@ func CopySnapshotLabelsAndAnnotations(application *applicationapiv1alpha1.Applic
 
 }
 
+// CopyTempGroupSnapshotLabelsAndAnnotations coppies labels and annotations from build pipelineRun or tested snapshot
+// into regular snapshot
+func CopyTempGroupSnapshotLabelsAndAnnotations(application *applicationapiv1alpha1.Application, snapshot *applicationapiv1alpha1.Snapshot, componentName string, source *metav1.ObjectMeta, prefixes []string) {
+	if snapshot.Labels == nil {
+		snapshot.Labels = map[string]string{}
+	}
+
+	if snapshot.Annotations == nil {
+		snapshot.Annotations = map[string]string{}
+	}
+	snapshot.Labels[SnapshotTypeLabel] = SnapshotGroupType
+
+	snapshot.Labels[ApplicationNameLabel] = application.Name
+
+	// Copy PAC annotations/labels from source(tested snapshot or pipelinerun) to snapshot.
+	_ = metadata.CopyLabelsWithPrefixReplacement(source, &snapshot.ObjectMeta, "pipelinesascode.tekton.dev", PipelinesAsCodePrefix)
+	_ = metadata.CopyAnnotationsWithPrefixReplacement(source, &snapshot.ObjectMeta, "pipelinesascode.tekton.dev", PipelinesAsCodePrefix)
+
+	for _, prefix := range prefixes {
+		// Copy labels and annotations prefixed with defined prefix
+		_ = metadata.CopyLabelsByPrefix(source, &snapshot.ObjectMeta, prefix)
+		_ = metadata.CopyAnnotationsByPrefix(source, &snapshot.ObjectMeta, prefix)
+	}
+}
+
 // IsOverrideSnapshot returns true if snapshot label 'test.appstudio.openshift.io/type' is 'override'
 func IsOverrideSnapshot(snapshot *applicationapiv1alpha1.Snapshot) bool {
 	return metadata.HasLabelWithValue(snapshot, SnapshotTypeLabel, SnapshotOverrideType)
@@ -1064,10 +1091,10 @@ func HasPRGroupProcessed(snapshot *applicationapiv1alpha1.Snapshot) bool {
 	return metadata.HasAnnotation(snapshot, PRGroupCreationAnnotation)
 }
 
-// GetPRGroupFromSnapshot gets the value of label test.appstudio.openshift.io/pr-group-sha and annotation from component snapshot
-func GetPRGroupFromSnapshot(snapshot *applicationapiv1alpha1.Snapshot) (string, string) {
-	if metadata.HasLabel(snapshot, PRGroupHashLabel) && metadata.HasAnnotation(snapshot, PRGroupAnnotation) {
-		return snapshot.Labels[PRGroupHashLabel], snapshot.Annotations[PRGroupAnnotation]
+// GetPRGroup gets the value of label test.appstudio.openshift.io/pr-group-sha and annotation from component snapshot or pipelinerun
+func GetPRGroup(object client.Object) (string, string) {
+	if metadata.HasLabel(object, PRGroupHashLabel) && metadata.HasAnnotation(object, PRGroupAnnotation) {
+		return object.GetLabels()[PRGroupHashLabel], object.GetAnnotations()[PRGroupAnnotation]
 	}
 	return "", ""
 }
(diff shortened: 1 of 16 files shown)


hongweiliu17:
the [e2e job](https://console.redhat.com/application-pipeline/workspaces/rhtap-integration/applications/integration-service/pipelineruns/konflux-e2e-nq7sb/logs?task=konflux-e2e-tests) fails with 
` `` 
{
    "name": "wait for the last components build to finish",
    "start": 1741170477222,
    "end": 1741170721623,
    "status": "failed"
  }
```

File: gitops/snapshot.go [hunk 1]
```diff
@@ -1058,16 +1058,31 @@ func FilterIntegrationTestScenariosWithContext(scenarios *[]v1beta2.IntegrationT
...
+		for _, scenarioContext := range scenario.Spec.Contexts {
+			if scenarioContext.Name == contextName {
+				filteredScenarios = append(filteredScenarios, scenario)
+				continue
+			}
+		}
+	}
+	return &filteredScenarios
+}
+
```
kasemAlem, line 1075: For this function I would suggest we use the break approach for better efficiency
also using pointers might cause memory leaks  

```
   func GetRequiredContextIntegrationTestScenario(scenarios []v1beta2.IntegrationTestScenario, contextName string) []v1beta2.IntegrationTestScenario {
	var filteredScenarios []v1beta2.IntegrationTestScenario
	
	for i := range scenarios {
		currentScenario := scenarios[i] // Avoid loop variable capture issue
		
		for _, scenarioContext := range currentScenario.Spec.Contexts {
			if scenarioContext.Name == contextName {
				filteredScenarios = append(filteredScenarios, currentScenario)
				break // Exit inner loop after finding a match
			}
		}
	}
	
	return filteredScenarios
} 
```

hongweiliu17: Thanks! I agree. I wanted to use `break`, but my head doesn't work well at that time...

gitops/snapshot.go
```diff
@@ -1058,16 +1058,30 @@ func FilterIntegrationTestScenariosWithContext(scenarios *[]v1beta2.IntegrationT
 	return &filteredScenarioList
 }
 
+func GetRequiredContextIntegrationTestScenario(scenarios []v1beta2.IntegrationTestScenario, contextName string) *[]v1beta2.IntegrationTestScenario {
```
dirgim, line 1061: There is a potential issue with this function, because this will return only the integrationTestScenarios that have the `group` context, but won't return the ones which would also run for the group Snapshot, like the ones with the `application`/`all` context (which means the ITS runs in all cases, including group testing) as well as the `pull_request` context (which executes all ITS for an open Pull Request, which also includes group testing).

hongweiliu17: I suppose [these code](https://github.com/konflux-ci/integration-service/pull/1049/files#diff-3540d1f95a378c2e6bea89135278f11b57b55c7c4e0a7ac892eb6dfd251b5b55R238-R251) can get other context ITS and also `group` ITS and join them together. But after thinking about again, it can't work well and also you remind me the checkRun's title for `group` snapshot is different from `component` snapshot. Let me fix it. Thanks!

hongweiliu17: Updated by creating a temp group snapshot.

internal/controller/buildpipeline/buildpipeline_adapter.go
```diff
@@ -650,3 +691,112 @@ func generateDetails(buildPLR *tektonv1.PipelineRun, integrationTestStatus intgt
...
+		if slices.Contains(componentsFromPipelineRun, applicationComponent.Name) {
+			a.logger.Info(fmt.Sprintf("There is in progress build pipelineRun for component %s/%s, won't check its component snapshot's pull/merge request", a.pipelineRun.Namespace, applicationComponent.Name))
+			componentsWithOpenPRMR = append(componentsWithOpenPRMR, applicationComponent.Name)
+			continue
+		}
+
+		if slices.Contains(componentsFromSnapshot, applicationComponent.Name) {
+			a.logger.Info(fmt.Sprintf("There is component snapshot for component %s/%s, will check its component snapshot MR/PR status", a.pipelineRun.Namespace, applicationComponent.Name))
+			var foundSnapshotWithOpenedPR *applicationapiv1alpha1.Snapshot
+			foundSnapshotWithOpenedPR, err = a.findSnapshotWithOpenedPR(&applicationComponent, a.pipelineRun.Namespace, prGroupHash)
```
dirgim, line 761: It seems to me that only the `Name` of the applicationComponent is used in the `findSnapshotWithOpenedPR` function. If we modified the function to take the `Name` string, we could greatly simplify [this loop](https://github.com/konflux-ci/integration-service/pull/1049/files#diff-3540d1f95a378c2e6bea89135278f11b57b55c7c4e0a7ac892eb6dfd251b5b55R749) so it doesn't require fetching all applications components and looping through them, and we could simply loop through `componentsFromSnapshot` instead.

dirgim: We also wouldn't need to append the components from in-flight builds [here](https://github.com/konflux-ci/integration-service/pull/1049/files#diff-3540d1f95a378c2e6bea89135278f11b57b55c7c4e0a7ac892eb6dfd251b5b55R752-R756) and could simply check that list directly when making a decision if a group Snapshot will be created.
Basically something like this:

- If `componentsFromPipelineRun` is more than 1, return `true` immediately since we know that a group Snapshot will need to be created for at least those ones, so no need to check further
- Otherwise loop through the `componentsFromSnapshot` and check if they have open PRs.
- At the end, if there are more than 1 `componentsFromSnapshot` that have open PRs, return `true`

internal/controller/buildpipeline/buildpipeline_adapter.go
```diff
@@ -650,3 +691,112 @@ func generateDetails(buildPLR *tektonv1.PipelineRun, integrationTestStatus intgt
...
+	}
+	if len(componentsWithOpenPRMR) < 2 {
+		a.logger.Info(fmt.Sprintf("The number %d of components affected by this PR group %s is less than 2, skipping group snapshot creation", len(joinedComponents), prGroup))
+		return false, nil
+	}
+	return true, nil
+}
+
+// findSnapshotWithOpenedPR is copied from snapshot_adapter.go and it find the snapshot with opened PR/MR for the given component
+func (a *Adapter) findSnapshotWithOpenedPR(applicationComponent *applicationapiv1alpha1.Component, nameSpace, prGroupHash string) (*applicationapiv1alpha1.Snapshot, error) {
```
dirgim, line 779: Not sure how complex it would be to extract both of these copied functions out of their adapters and into the `gitops` package to prevent code duplication, so @hongweiliu17  I leave it up to your judgement if that effort would be worth it.

dirgim (changes requested):

internal/controller/buildpipeline/buildpipeline_adapter.go
hongweiliu17: I thought to combine them together, let me rethink it. thanks for all comments and I will update them.

gitops/snapshot.go
```diff
@@ -974,6 +976,31 @@ func CopySnapshotLabelsAndAnnotations(application *applicationapiv1alpha1.Applic
 
 }
 
+// CopyTempGroupSnapshotLabelsAndAnnotations coppies labels and annotations from build pipelineRun or tested snapshot
+// into regular snapshot
+func CopyTempGroupSnapshotLabelsAndAnnotations(application *applicationapiv1alpha1.Application, snapshot *applicationapiv1alpha1.Snapshot, componentName string, source *metav1.ObjectMeta, prefixes []string) {
```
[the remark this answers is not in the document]
hongweiliu17: Added.Thanks.

internal/controller/buildpipeline/buildpipeline_adapter.go
hongweiliu17: Updated. Thanks!

internal/controller/buildpipeline/buildpipeline_adapter.go
hongweiliu17: Moved to status package. Thanks.

dirgim (approved):
LGTM, great work!

internal/controller/buildpipeline/buildpipeline_adapter.go
dirgim: Thanks, nicely done!

kasemAlem:
Great work you made @hongweiliu17

jsztuka (approved):
lgtm, great work

PR #1058, opened 2025-03-13
fix(STONEINTG-1050): annotation correction

jsztuka:

Made adjustemnt regarding how we filter the snapshots in question.
Since the repo_url PAC annotation is not duplicated in labels, we need to approach it individually.
Also added the case if we have only single snapshot and customer decides to run multiple integration tests. I think that it was not previously covered.


jsztuka:
Will see if the tests pass this time, then I will include more explanation in the commit msg.

jsztuka:
Failing to provision cluster, not sure if this is connected to the changes, but it should not be.

gitops/snapshot.go
```diff
@@ -1097,6 +1098,20 @@ func SortSnapshots(snapshots []applicationapiv1alpha1.Snapshot) []applicationapi
 	return snapshots
 }
 
+// SortPipelineruns sorts the pipelineruns according to the snapshot annotation creationTimeStamp
+func SortPipelineruns(pipelineruns []tektonv1.PipelineRun) []tektonv1.PipelineRun {
```
dirgim, line 1102: I think this function is unused now so best to remove it.

dirgim (approved):
LGTM

kasemAlem (approved):

PR #1072, opened 2025-03-26
feat: add Helm chart

yftacherzog:
Here we introduce a Helm chart that should provide, at least, the manifests required for konflux-ci/konflux-ci. Additional variants will be introduced later.

One of the next steps will be to onboard a new component that will package and push this Helm chart to a registry, so it can later be used in konflux-ci/konflux-ci and in infra-deployments.


avi-biton:
The structure looks good, but I rather wait for other inputs

.github/workflows/test-chart.yml
```diff
@@ -0,0 +1,89 @@
...
+          yq -i '.controllerManager.container.image = env(IMG)' dist/chart/values.yaml
+          make img-build IMG=$IMG
+          kind load docker-image $IMG
+
+      - name: Verify Helm installation
+        run: helm version
+
+      - name: Lint Helm Chart
+        run: |
+          helm lint ./dist/chart
```
amisstea, line 49: We're [using kube-linter](https://github.com/konflux-ci/integration-service/blob/c40254a2fd810b2e2a00d1467a9a989f544d05eb/.github/workflows/pr.yaml#L156-L158) to scan the manifests from `/config` from the PR workflow. I think we also need to run `kube-linter` in this workflow since its using a different set of manifests.

yftacherzog: I wonder if kube-linter will be able to lint Helm templates, as they are not exactly manifests. I'll look into that.

amisstea: Running `helm template` and scanning the result should take care of that.

yftacherzog:
Moving to draft until we have [this one](https://github.com/konflux-ci/integration-service/pull/1074) merged and verified, and then I'll adjust the current PR accordingly. .

.github/workflows/test-chart.yml
yftacherzog: Reading now, I see that it supports Helm natively. I ran it and it actually helped me find a bug in one of the templates. I'll look into running it in CI.

yftacherzog: I just added it (and confirmed it to be able to pass and fail).

yftacherzog:
@dirgim can I have your remove?

.github/workflows/test-chart.yml
```diff
@@ -0,0 +1,102 @@
...
+        run: |
+          curl -Lo ./kind https://kind.sigs.k8s.io/dl/latest/kind-linux-amd64
+          chmod +x ./kind
+          sudo mv ./kind /usr/local/bin/kind
+
+      - name: Verify kind installation
+        run: kind version
+
+      - name: Create kind cluster
+        run: kind create cluster
```
jsztuka, line 34: Are we using default config here? What specifications is the cluster?

.github/workflows/test-chart.yml
```diff
@@ -0,0 +1,102 @@
...
+
+      - name: Install Prometheus Operator CRDs
+        run: |
+          helm repo add prometheus-community https://prometheus-community.github.io/helm-charts
+          helm repo update
+          helm install prometheus-crds prometheus-community/prometheus-operator-crds
+
+      - name: Install Prometheus via Helm
+        run: |
+          helm repo add prometheus-community https://prometheus-community.github.io/helm-charts
```
jsztuka, line 71: Why are we duplicating code from line 65 and 66?

.github/workflows/test-chart.yml
```diff
@@ -0,0 +1,102 @@
...
+        run: |
+          helm upgrade --install my-release ./dist/chart --set chartCreateNamespace=true
+
+      - name: Check Helm release status
+        run: |
+          helm status my-release
+
+      - name: Check Presence of ServiceMonitor
+        run: |
+          kubectl wait --namespace integration-service --for=jsonpath='{.kind}'=ServiceMonitor servicemonitor/integration-service-controller-manager-metrics-monitor
```
jsztuka, line 89: No timeout is needed here?

jsztuka:
Just a few nits, great work on this one! Can't wait to see it in action.

.github/workflows/test-chart.yml
yftacherzog: Yes. It's using the default Kind configs.
I not sure what are the defaults, and I'm not sure it's properly documented. [The documentation](https://kind.sigs.k8s.io/docs/user/configuration/) mentions a few words about defaults, but I could not find a summary of all defaults (The workflow, including the command for creating the Kind cluster was auto-generated by kubebuilder).

.github/workflows/test-chart.yml
yftacherzog: Thanks for pointing this out. I suspect that those two sections were auto-generated pre-commented-out, allowing you to uncomment either independently, so the lines are repeated to allow them to work either way.

As we use both, there's indeed no need to repeat those lines. I'll have it addressed.

.github/workflows/test-chart.yml
yftacherzog: It's using the default 30s. I'll add that for consistency.

.github/workflows/test-chart.yml
jsztuka: Do we maybe want to use predefined config?

dist/chart/templates/rbac/role.yaml
```diff
@@ -0,0 +1,128 @@
+{{- if .Values.rbac.enable }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
dirgim, line 4: Similar to the other comment, we may have updates for these roles in near future, is there a way to regenerate this file?

.github/workflows/test-chart.yml
yftacherzog: I think for now we should be good without it. If later on we see a need, we can have that added.

dist/chart/templates/crd/appstudio.redhat.com_integrationtestscenarios.yaml
```diff
@@ -0,0 +1,646 @@
+{{- if .Values.crd.enable }}
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
```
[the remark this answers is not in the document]
yftacherzog: The manifests, including the CRDs, were generated by kubebuilder by enabling the kubebuilder [Helm plugin](https://book.kubebuilder.io/plugins/available/helm-v1-alpha). They did require customizations afterwards to be aligned with the configs generated by Kustomize.

In this example, when modifying the CRDs, you can run the plugin again (`kubebuilder edit --plugins=helm/v1-alpha`) and it will propagate the changes in the CRDs to the Helm chart.
The caveat here is that it will undo the customizations in the charts mentioned before, so whoever introduces the changes will need to pick what needs to change and what doesn't.

dist/chart/templates/rbac/role.yaml
yftacherzog: Same as for the other comment.

dist/chart/templates/crd/appstudio.redhat.com_integrationtestscenarios.yaml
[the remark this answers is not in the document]
dirgim: Since the soon-to-be-merged #1091 will be modifying the CRD, would you prefer to update this PR or to demonstrate how the Helm chart needs to be updated for that in a follow-up PR?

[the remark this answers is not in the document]
yftacherzog: I think we can demonstrate the process of updating. I'll compare the manifests generated by Helm to those generated by Kustomize both when we get to switch konflux-ci/konflux-ci to use Helm and when we do the same for infra-deployments, so we'll have chances to make sure everything is in-line.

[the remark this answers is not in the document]
dirgim: That sound good to me. I'll approve the PR as-is and we can wait for the follow-up after the CRDs (and potentially RBACs) are updated.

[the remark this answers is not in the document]
dirgim: @14rcole also fyi as #1092 may require updating RBACs to allow creating/updating/deleting `ResolutionRequest` CRs.

dirgim (approved):
LGTM

jsztuka (approved):
lgtm

PR #1092, opened 2025-04-08
feat(stoneintg-1165): resolve pipelineruns from scenario

14rcole:
Support referencing pipelineruns from integrationtestscenario resolver field.  Uses  Tekton's ResolutionRequest CR to grab remote data then creates PipelineRun in the cluster


tekton/integration_pipeline.go
```diff
@@ -125,15 +138,87 @@ func NewIntegrationPipelineRun(prefix, namespace string, integrationTestScenario
...
 				},
 			},
 		},
 	}
+
 	return &IntegrationPipelineRun{pipelineRun}
 }
 
+func getPipelineRunYamlFromPipelineRunResolver(client client.Client, ctx context.Context, prefix, namespace, resolver string, resolverParams []tektonv1.Param) (*IntegrationPipelineRun, error) {
+	request = tektonv1.ResolutionRequest{
```
kasemAlem, line 152:
```suggestion
	request := tektonv1.ResolutionRequest{
```

dirgim:
Note that manipulating ResolutionRequest CRs will probably require RBACs to be updated accordingly, which should be done by expanding the existing kubebuilder directives here: https://github.com/konflux-ci/integration-service/blob/88a813cfef413b4a1304cce54ee753f9935b1c66/internal/controller/snapshot/snapshot_controller.go#L57-L70

config/crd/bases/appstudio.redhat.com_integrationtestscenarios.yaml
```diff
@@ -3,7 +3,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.16.1
+    controller-gen.kubebuilder.io/version: v0.14.0
```
dirgim, line 6: I think it might be better to update to `v0.16.1` version locally by removing(or moving) the `controller-gen` from the `bin` directory in the integration-service and running `make manifests generate` which should automatically install the newer version of the binary based on the [Makefile configuration](https://github.com/konflux-ci/integration-service/blob/88a813cfef413b4a1304cce54ee753f9935b1c66/Makefile#L177).

14rcole: I didn't mean to update this.  Looks like rerunning `make manifests generate` reverted it back to 0.16.1

config/rbac/role.yaml
```diff
@@ -81,6 +81,14 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
```
hongweiliu17, line 84: I think we should also add rbac for `resolutionrequests/status`

tekton/integration_pipeline.go
```diff
@@ -125,15 +152,97 @@ func NewIntegrationPipelineRun(prefix, namespace string, integrationTestScenario
...
+
+	request := resolutionv1alpha1.ResolutionRequest{
+		ObjectMeta: metav1.ObjectMeta{
+			GenerateName: prefix + "-",
+			Namespace:    namespace,
+			Labels: map[string]string{
+				"resolution.tekton.dev/type": resolver,
+			},
+			Annotations: map[string]string{
+				"konflux-ci.dev/created-by": "integration-service", // for backup garbage collection
```
dirgim, line 180: I like this idea, but it might work better as a label since labels are indexed and queryable, so this would make the job of any garbage collectors much easier.

tekton/integration_pipeline.go
```diff
@@ -125,15 +152,97 @@ func NewIntegrationPipelineRun(prefix, namespace string, integrationTestScenario
...
+	}
+	return resolvedRequest.Status.Data, nil
+}
+
+func generateIntegrationPipelineRunFromBase64(base64plr string) (*IntegrationPipelineRun, error) {
+	plrYaml, err := base64.StdEncoding.DecodeString(base64plr)
+	if err != nil {
+		return nil, err
+	}
+	// TODO: Make sure this is sufficient to verify that kind is pipelineRun
```
dirgim, line 237: It might be a good idea to create a follow-up ticket for this.
I'm a bit worried that the current approach may cause issues if there are major differences between the tekton version that we use in the integration service code vs the one that's deployed on the cluster.
Since we're supposed to keep pace with it anyways, it's probably not a major concern, but it may be good to have a tracker for it to improve in the future.

tekton/integration_pipeline.go
14rcole: I used an annotation because they're lighter weight and because KubeArchive works with either. I can switch over to a label though in case we use some other mechanism for GC

tekton/integration_pipeline.go
14rcole: That was actually a note for myself.  I verified and forgot to remove the comment. Unless there's some other CRD  with identical fields to the pipelinerun and a different controller this will be fine.  I think that's unlikely enough that we can just use unmarshalling as verification

dirgim:
@14rcole how feasible would it be to provide unit test coverage for the `pipelineRun` resolution?

On first look, creating a new [loader function](https://github.com/konflux-ci/integration-service/blob/main/loader/loader.go) to handle getting resolutionRequests (which is currently done [here](https://github.com/konflux-ci/integration-service/pull/1092/files#diff-2a3223c5451cf64d18f7bab690054180b4c4881a8592b94ca121ccde19c8f83cR213)) and then using mock loaders to return a `resolutionRequest` with the correct status and a prepared pipelineRun yaml in its `resolvedRequest.Status.Data` field might be able to do most of the job, but I may be missing something.

14rcole:
@dirgim I tried that but we can't import the `loader` package into the `tekton` package because it would create a circular dependency.  I would have to migrate out either all of the constants or the functions related to creating the PipelineRuns.  I decided that rather than take on such a big refactor here I would file a follow-up story.  I confirmed manually that this works

dirgim (approved):
LGTM, nice work!

kasemAlem:
great work @14rcole 
/retest

PR #1096, opened 2025-04-10
feat(STONEINTG-852): integration service reports on push events

kasemAlem:
- test integration results will show in commits also on push event
Signed-off-by: Kasem Alem <kalem@redhat.com>

### DIFF
--- a/internal/controller/snapshot/snapshot_adapter.go
+++ b/internal/controller/snapshot/snapshot_adapter.go
@@ -567,12 +567,8 @@ func (a *Adapter) EnsureOverrideSnapshotValid() (controller.OperationResult, err
 }
 
 // EnsureGroupSnapshotExist is an operation that ensure the group snapshot is created for component snapshots
-// once a new component snapshot is created for an pull request and there are multiple existing PRs belonging to the same PR group
+// once a new component snapshot is created for an pull or push request and there are multiple existing PRs belonging to the same PR group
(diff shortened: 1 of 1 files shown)


internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -567,12 +567,8 @@ func (a *Adapter) EnsureOverrideSnapshotValid() (controller.OperationResult, err
 }
 
 // EnsureGroupSnapshotExist is an operation that ensure the group snapshot is created for component snapshots
-// once a new component snapshot is created for an pull request and there are multiple existing PRs belonging to the same PR group
+// once a new component snapshot is created for an pull or push request and there are multiple existing PRs belonging to the same PR group
 func (a *Adapter) EnsureGroupSnapshotExist() (controller.OperationResult, error) {
-	if gitops.IsSnapshotCreatedByPACPushEvent(a.snapshot) {
```
dirgim, line 572: It is not possible to create reliable group Snapshots for push events, so I don't think we should remove this condition.

dirgim (changes requested):
Ensuring that group Snapshot exist is a feature that's only meant to work for Pull requests, so I'm against this change.

internal/controller/snapshot/snapshot_adapter.go
hongweiliu17: +1, push event won't expect group snapshot.

PR #1136, opened 2025-05-19
fix(STONEINTG-1046): set the ITS PLR console url in PaC annotation

kasemAlem:
This PR is created to fix an issue we have : 
 Problem : 
     - IS copy the annotations and labels from snapshot to ITS PLR: 
          ITS PLR annotation  `pac.test.appstudio.openshift.io/log-url` get the log url of build PLR , we want it to have the 
              value of ITS PLR name .
  
  Solution: 
     -  When copying the annotations, if it's a PaC log-url  one , set the value to `NewIntegrationPipelineRun.Name`
     - Create a unit test to verify this function 
     
    more info are in Jira  [STONEINTG-1046](https://issues.redhat.com/browse/STONEINTG-1046) 


tekton/integration_pipeline.go
```diff
@@ -349,6 +349,11 @@ func (r *IntegrationPipelineRun) WithSnapshot(snapshot *applicationapiv1alpha1.S
 		_ = metadata.CopyLabelsByPrefix(&snapshot.ObjectMeta, &r.ObjectMeta, prefix)
 	}
 
+	// If the Annotation is PaC log-url, we need to set the value to the console URL of the PipelineRun
+	if metadata.HasAnnotation(r, PipelinesAsCodePrefix+"/log-url") {
+		pipelineRunConsoleURL := "https://console.redhat.com/openshift/pipelines/ns/" + r.Namespace + "/pipelineruns/" + r.Name
```
dirgim, line 354: The base URL is different for different kinds of clusters (staging, internal, user-provisioned, etc.), so I think we should use the [FormatPipelineRunURL](https://github.com/konflux-ci/integration-service/blob/1025b17955efd24a334e549bba327db7c72ad0e5/status/format.go#L248) function which accounts for these differences.

kasemAlem: when I tried to use it ,  I've added an import of  `"github.com/konflux-ci/integration-service/status"`
but I got some circle conflict with pkges 
maybe I'm using it incorrectly

PR #1147, opened 2025-05-22
fix(STONEINTG-1046): add PaC anntotation ot ITS PLR with console log url

kasemAlem:


internal/controller/integrationpipeline/integrationpipeline_adapter.go
```diff
@@ -90,6 +91,14 @@ func (a *Adapter) EnsureStatusReportedInSnapshot() (controller.OperationResult,
 			return err
 		}
 
+		integrationPipelineRunURL := status.FormatPipelineURL(a.pipelineRun.Name, a.pipelineRun.Namespace, a.logger.Logger)
+		if integrationPipelineRunURL == "" {
+			return fmt.Errorf("failed to get integration pipeline run console URL")
```
dirgim, line 96: I'm thinking that this might not be worth throwing an error for - maybe we can do something like this instead:
```
if integrationPipelineRunURL != "" {
    err = a.annotateIntegrationPipelineRunLogURL(a.context, a.client, a.pipelineRun, tekton.PipelinesAsCodePrefix+"/log-url", integrationPipelineRunURL)
	 if err != nil {
	     return err
	}
}
		
```

internal/controller/integrationpipeline/integrationpipeline_adapter.go
```diff
@@ -90,6 +91,13 @@ func (a *Adapter) EnsureStatusReportedInSnapshot() (controller.OperationResult,
 			return err
 		}
 
+		integrationPipelineRunURL := status.FormatPipelineURL(a.pipelineRun.Name, a.pipelineRun.Namespace, a.logger.Logger)
+		if integrationPipelineRunURL != "" {
+			err = a.annotateIntegrationPipelineRunLogURL(a.context, a.client, a.pipelineRun, tekton.PipelinesAsCodePrefix+"/log-url", integrationPipelineRunURL)
+			if err != nil {
+				return err
+			}
+		}
```
dirgim, line 100: I'm still thinking that this whole logic would probably make more sense in a separate `Ensure` function since the `EnsureStatusReportedInSnapshot` doesn't really have anything to do with updating the integration pipelineRun.

hongweiliu17: How about adding it after integration pipelinerun is created in https://github.com/konflux-ci/integration-service/blob/main/internal/controller/snapshot/snapshot_adapter.go#L788?

kasemAlem: thanks @dirgim for the important points you raised,  I've posted another comment with new `Ensure`  function 
as it was agreed , please check that 

@hongweiliu17 , that is another approach frankly I didn't thought of. We discussed it on the stand-up and Kruno made a fair point that this might be too risky, cause when PLR is created so it will be worked on by the tekton, thus we for safe side it be handled in the ITS PLR adapter as now.

kasemAlem:
@dirgim , still I'm refactoring the function `annotateIntegrationPipelineRunURL` as you requested in [comment](https://github.com/konflux-ci/integration-service/pull/1147#discussion_r2107474713)

PR #1151, opened 2025-05-28
fix: race condition in pipelinerun creation

14rcole:
When a snapshot is created, we go through the list of scenarios associated with that snapshot's application and create pipelineruns for any valid scenarios.  If a scenario is created at the same time as the snapshot then it may not be marked as valid in time for it to be run. Since the scenario is marked as neither valid nor invalid we were running into a nil point dereference from the missing status condition. This commit handles that edge case.


dirgim (approved):
LGTM

sonam1412 (approved):
LGTM

kasemAlem (approved):

PR #1197, opened 2025-07-02
fix: use a pipelinerun annotation to skip the snapshot release

### LINKED ISSUE #1192: How to adopt github merge queues without auto-releasing pre-merge builds?
Take a look at the documentation for adopting github merge queues here: https://konflux-ci.dev/docs/patterns/github-merge-queues/#modify-the-on-cel-expression-of-your-pull-request-pipeline

It recommends the following on-cel expressions for your pull request pipelinerun definition:

```
    pipelinesascode.tekton.dev/on-cel-expression: (event == "pull_request" && target_branch == "main") || (event == "push" && target_branch.startsWith("gh-readonly-queue/main/"))
```

The thing to notice is that a *push event* will trigger that pipeline, but only in a certain circumstance where the github merge queue branches are used. This _works_ in that it triggers those PR pipelineruns at the right time, triggers testing, and makes the merge queue proceed safely.

The unexpected side-effect, however, is that the build pipelineruns that complete as a part of that merge queue have the `pac.test.appstudio.openshift.io/event-type=push` label, and therefore their Snapshots bear that label, and therefore they are a candidate for auto-release even though they're not the resulted of *merged* PRs.

---

* Does that scenario/diagnosis sound plausible?
* How should we resolve this?

mmartinv:
Using the approach described at [1] has the side effect of releasing the associated snapshots before merging them into main.

This is a simple patch that allows the release of those snapshots by annotating the pipelinerun with:

`custom.appstudio.openshift.io/auto-release: "false"`

[1] https://konflux-ci.dev/docs/patterns/github-merge-queues/#modify-the-on-cel-expression-of-your-pull-request-pipeline

Resolves #1192


dirgim:
> > I think that the `custom.appstudio.openshift.io/auto-release` prefix makes sense since this annotation would concern user-side customization of the integration/release workflow and not the build part of it.
> 
> I tend to disagree here, I would think the main goal of custom labels/annotations is to cover usecases that are not related to Konflux workflow itself. Indeed, as a user, I would expect the label/annotation to be `release.appstudio.openshift.io/auto-release` (the same as the one that triggers auto releases) as a way offered to the user of overwriting the release behavior for specific push events.

I'm fine with using the `release.appstudio.openshift.io/auto-release` annotation in that case.

dirgim:
> > I think that the `custom.appstudio.openshift.io/auto-release` prefix makes sense since this annotation would concern user-side customization of the integration/release workflow and not the build part of it.
> 
> I tend to disagree here, I would think the main goal of custom labels/annotations is to cover usecases that are not related to Konflux workflow itself. Indeed, as a user, I would expect the label/annotation to be `release.appstudio.openshift.io/auto-release` (the same as the one that triggers auto releases) as a way offered to the user of overwriting the release behavior for specific push events.

The issue with using `release.appstudio.openshift.io/auto-release` is that it won't be automatically copied over to the Snapshot that's created for that pipelineRun (see [here](https://konflux-ci.dev/docs/testing/integration/creating/#data-injected-into-the-pipelinerun-of-the-integration-test)).

dirgim (approved):
LGTM

PR #1272, opened 2025-09-10
feat(konflux-8553): allow users to define auto-release logic

14rcole:
This change gives users more power over whether they auto-release push snapshots.  For example, if a user has an operator bundle they are now able to auto-release only if the bundle image is the updated component in their snapshot. The `release.appstudio.openshift.io/auto-release` label now accepts cel expressions.  A missing or empty label still evaluates to true.  Since 'true' and 'false' are valid cel expression this change is fully backward compatible.

### DIFF
--- a/gitops/release.go
+++ b/gitops/release.go
@@ -1,0 +1,144 @@
+/*
+Copyright 2024 Red Hat Inc.
+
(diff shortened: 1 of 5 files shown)


jencull (approved):
lgtm

Josh-Everett (approved):

PR #1290, opened 2025-10-03
fix(stoneintg-1352): fall back on target project updating PR status

14rcole:
This commit aligns the integration service gitlab PR status update strategy with that of PaC.  Comment from PaC:

By default, Pipelines-as-Code attempts to update the commit status through the GitLab API. It first tries the source project (fork), then falls back to the target project (upstream repository). The source project update succeeds when the configured token has access to the source repository and GitLab creates pipeline entries for external CI systems like Pipelines-as-Code. The target project fallback may fail if there's no CI pipeline running for that commit in the target repository, since GitLab only creates pipeline entries for commits that actually trigger CI in that specific project. If either status update succeeds, no comment is posted on the Merge Request.

### DIFF
--- a/status/reporter_gitlab.go
+++ b/status/reporter_gitlab.go
@@ -203,21 +203,30 @@ func (r *GitLabReporter) setCommitStatus(report TestReport) (int, error) {
 	r.logger.Info("creating commit status for scenario test status of snapshot",
 		"scenarioName", report.ScenarioName)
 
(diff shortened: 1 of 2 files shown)


dirgim:
@14rcole just realized that we only report the commit status if the source and project IDs are same (see [here](https://github.com/konflux-ci/integration-service/blob/f9ffbd484a619cd397ea4f863f434d803015b954/status/reporter_gitlab.go#L297-L300)) - should we change that behavior too?

dirgim (approved):
LGTM

PR #1299, opened 2025-10-19
fix(KONFLUX-10405): integration passed PLR was wrongly reported, fixed

kasemAlem:
Issue:

GitHub shows "failed" because the main PipelineRun is reporting an error. This error happens because it's trying to confirm the status of its child TaskRun , but the worker task has already been deleted from the cluster.

It's a race condition: the worker finishes successfully, gets cleaned up too quickly, and then the manager comes to check on it. Since the worker is gone, the manager panics and reports "I don't know what happened!", which GitHub sees as a failure.

Before Fix:
❌ Successful tests showed as failed on GitHub
😕 Developers wasted time investigating "fake failures"
🚫 Pull requests looked broken when they were actually fine

After Fix:
✅ Successful tests show as passed on GitHub
😊 Developers see correct status
🎉 Pull requests show accurate test results

[KONFLUX-10405](https://issues.redhat.com/browse/KONFLUX-10405)

### DIFF
--- a/status/reporter_gitlab.go
+++ b/status/reporter_gitlab.go
@@ -304,7 +304,9 @@ func (r *GitLabReporter) ReportStatus(ctx context.Context, report TestReport) (i
 	}
 
 	if statusCode, err := r.setCommitStatus(report); err != nil {
-		return statusCode, fmt.Errorf("failed to set gitlab commit status: %w", err)
+		r.logger.Error(err, "failed to set gitlab commit status, will attempt to leave a comment on the MR")
+	} else {
+		return statusCode, nil
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/integrationteststatus/integration_test_status.go
```diff
@@ -163,6 +166,12 @@ func (sits *SnapshotIntegrationTestStatuses) UpdateTestStatusIfChanged(scenarioN
 		sits.dirty = true
 	}
 
+	// CRITICAL FIX: Never overwrite final statuses
+	if detail.Status.IsFinal() {
+		// Do not update final statuses - they are immutable once set
+		return
+	}
```
14rcole: What is a final status and when would we set it?

PR #1370, opened 2025-11-28
Add test code coverage reporting using Coverport

msu8:
Add test code coverage reporting using Coverport, for more information
https://github.com/konflux-ci/coverport


jsztuka (approved):
lgtm

PR #1389, opened 2025-12-08
chore(STONEINTG-1189): annotate PLR when component is missing

jencull:
This fixes a silent failure where the Build PipelineRun would stall if
    the appstudio.openshift.io/component label was incorrect, not found,
    or missing entirely.
    
    Changes:
    * Added logic to catch "Component Not Found" errors during loader.
    * Added logic to catch cases where the component label is nil.
    * In both cases, the PipelineRun is now annotated with
      test.appstudio.openshift.io/create-snapshot-status="failed".
    * Ensured Finalizer is removed if component is not found, allowing
      resource deletion.
    
    Testing:
    * Added unit tests in buildpipeline_controller_test.go covering
      both scenarios.
    
    Generated-by: Cursor

### DIFF
--- a/internal/controller/buildpipeline/buildpipeline_controller.go
+++ b/internal/controller/buildpipeline/buildpipeline_controller.go
@@ -18,6 +18,7 @@ package buildpipeline
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/konflux-ci/integration-service/cache"
 	"k8s.io/client-go/util/retry"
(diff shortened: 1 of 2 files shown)


commit_msg.txt
```diff
@@ -0,0 +1,19 @@
+chore(STONEINTG-1189): annotate PLR when component is missing
```
dirgim: I don't think we should commit this file, but interesting to see it generated here.

PR #1455, opened 2026-02-05
chore: disable scenario controller

14rcole:
The only function in the scenario controller was EnsureScenarioContainsResourceKind, a function whose purpose was to add the `resourceKind` field  to all IntegrationTestScenarios. Now that the field exists for all scenarios on the cluster the function can be removed.  With that, we can also disable the scenario controller since it does not do any work. This PR disables the controller rather than deletes it entirely in case the controller is needed in the future.


dirgim (approved):
LGTM

PR #1461, opened 2026-02-09
feat(STONEINTG-1488): reduce git api call in snapshot/build plr controller

hongweiliu17:
* reduce git provier api call when preparing group snapshot in build plr and snapshot controller since component snapshot and build plr is created for open PR/MR, so we can consider its PR is opened, and also, monorepo components snapshot are created for the same PR/MR, so we can skip check PR/MR checking if the component snapshots from different components have the same git url and PR/MR number

Assisted-by: Claude Code AI
Signed-off-by: Hongwei Liu hongliu@redhat.com

### DIFF
--- a/gitops/snapshot.go
+++ b/gitops/snapshot.go
@@ -194,7 +194,7 @@ const (
 	PipelineAsCodePullRequestType = "pull_request"
 
 	// PipelineAsCodeMergeRequestType is the type of merge request event which triggered the pipelinerun in build service
-	PipelineAsCodeMergeRequestType = "merge request"
+	PipelineAsCodeMergeRequestType = "Merge Request"
 
 	// PipelineAsCodeGLMergeRequestType is the type of gitlab merge request event marked in label which triggered the pipelinerun in build service
 	PipelineAsCodeMergeUnderscoreRequestType = "Merge_Request"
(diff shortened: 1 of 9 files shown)


kasemAlem (approved):

gitops/snapshot.go
```diff
@@ -1592,3 +1592,23 @@ func IsCommentDisabled(ctx context.Context, adapterClient client.Client, compone
...
 	return false, nil
 }
+
+// HasSameGitSourceAndPRWithProcessedSnapshot checks if the given snapshot has the same git source and pull request number as the latest snapshot, which means the same PR, to avoid processing the same PR with multiple snapshots or pipelineruns
+func HasSameGitSourceAndPRWithProcessedSnapshot(latestSnapshot, snapshot *applicationapiv1alpha1.Snapshot) bool {
+	if latestSnapshot.GetName() == snapshot.GetName() {
+		return true
+	}
+
+	if !metadata.HasAnnotation(latestSnapshot, PipelineAsCodeRepoURLAnnotation) ||
```
jsztuka: Just to clarify, this only makes sure that latest snapshot and snapshot contains certain labels & annotations. What is the case when this situation occurs?

hongweiliu17: If any of the two snapshot have no these label/annotation, it means at least one of them are not from a pull-request, then `HasSameGitSourceAndPRWithProcessedSnapshot` get returned false, then we need to check if that snapshot is created for open pull/merge request.

PR #1534, opened 2026-04-17
fix(KONFLUX-13098): honour auto-release=false

NigelByrne1:
- set canBePromoted = false when release.appstudio.openshift.io/auto-release is false in CanSnapshotBePromoted()
 - Fix GetAutoReleasePlansForApplication() so CEL results append release plans when err == nil && canRelease
 - improve EvaluateSnapshotAutoReleaseAnnotation() 
 - unit tests added

asisted by AI - Cursor IDE ask mode

### DIFF
--- a/gitops/release_test.go
+++ b/gitops/release_test.go
@@ -119,6 +119,17 @@ var _ = Describe("Auto-release annotation evaluation", Ordered, func() {
 		Expect(allowed).To(BeFalse())
 	})
 
+	It("returns false when the snapshot has auto-release label false even if the expression is true", func() {
+		snapshotCopy := hasSnapshot.DeepCopy()
+		if snapshotCopy.Labels == nil {
+			snapshotCopy.Labels = make(map[string]string)
+		}
+		snapshotCopy.Labels[gitops.AutoReleaseLabel] = "false"
+		allowed, err := gitops.EvaluateSnapshotAutoReleaseAnnotation("true", snapshotCopy, true)
+		Expect(err).NotTo(HaveOccurred())
+		Expect(allowed).To(BeFalse())
+	})
+
 	Context("shouldRelease() CEL function", func() {
 		It("returns true when shouldRelease is true", func() {
 			snapshotCopy := hasSnapshot.DeepCopy()
--- a/gitops/snapshot_test.go
+++ b/gitops/snapshot_test.go
@@ -769,7 +769,35 @@ var _ = Describe("Gitops functions for managing Snapshots", Ordered, func() {
 		canBePromoted, reasons = gitops.CanSnapshotBePromoted(hasSnapshot)
 		Expect(canBePromoted).To(BeFalse())
 		Expect(reasons).To(HaveLen(3))
+	})
 
+	It("promotes the Snapshot when auto-release label is true", func() {
+		err := gitops.MarkSnapshotIntegrationStatusAsFinished(ctx, k8sClient, hasSnapshot, "Test message")
+		Expect(err).ToNot(HaveOccurred())
+		err = gitops.MarkSnapshotAsPassed(ctx, k8sClient, hasSnapshot, "Test message")
+		Expect(err).ToNot(HaveOccurred())
+
+		snapshotCopy := hasSnapshot.DeepCopy()
+		snapshotCopy.Labels[gitops.AutoReleaseLabel] = "true"
+		canBePromoted, reasons := gitops.CanSnapshotBePromoted(snapshotCopy)
+
+		Expect(canBePromoted).To(BeTrue())
+		Expect(reasons).To(BeEmpty())
+	})
+
+	It("does not promote the Snapshot when auto-release label is false even if otherwise eligible", func() {
+		err := gitops.MarkSnapshotIntegrationStatusAsFinished(ctx, k8sClient, hasSnapshot, "Test message")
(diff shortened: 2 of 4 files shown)


jsztuka:
I think we might need solution little more complex than that, for example the function that evaluates if the snapshot has auto-release annotation: https://github.com/konflux-ci/integration-service/blob/main/gitops/release.go#L33
for example, if the auto-release annotation contains following string: https://github.com/konflux-ci/integration-service/blob/main/gitops/release_test.go#L100 **no matter the duration** the snapshot is pronounced as auto-release == true. I would recommend updating  https://github.com/konflux-ci/integration-service/blob/main/gitops/release.go#L33 function logic. I'm honestly lost in the cel proccess of the function.

gitops/release_test.go
```diff
@@ -118,4 +118,15 @@ var _ = Describe("Auto-release annotation evaluation", Ordered, func() {
 		Expect(err).To(HaveOccurred())
 		Expect(allowed).To(BeFalse())
 	})
+
+	It("returns false when the snapshot has auto-release label false even if the expression is true", func() {
```
dirgim, line 122: I think we do need to support setting the label/annotation to "true" for backwards compatibility if I'm not missing something. @14rcole correct me if I'm wrong.

loader/loader.go
```diff
@@ -459,7 +459,7 @@ func (l *loader) GetAutoReleasePlansForApplication(ctx context.Context, c client
 		if ok || annotationValue != "" {
 			// If the annotation exists we evaluate the expression and return that
 			canRelease, err := gitops.EvaluateSnapshotAutoReleaseAnnotation(annotationValue, snapshot)
-			if err != nil || !canRelease {
+			if err == nil && canRelease {
```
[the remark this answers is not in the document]
14rcole: This  is  actually  still relevant. We  currently support an auto-release annotation and an auto-release label as a fallback. The label is legacy code that we can hopefully remove in https://redhat.atlassian.net/browse/STONEINTG-1607  but for now we need to support both

loader/loader.go
```diff
@@ -492,7 +492,7 @@ func (l *loader) GetAutoReleasePlansForComponentGroup(ctx context.Context, c cli
 		if annotationValue != "" {
 			// If the annotation exists we evaluate the expression and return that
 			canRelease, err := gitops.EvaluateSnapshotAutoReleaseAnnotation(annotationValue, snapshot)
-			if err != nil || !canRelease {
+			if err == nil && canRelease {
```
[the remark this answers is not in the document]
14rcole: See my comment above

gitops/release_test.go
14rcole: You're right. I added more info on one of the gemini comments

14rcole (changes requested):

jsztuka:
Changes looks ok to me, I see that there are some changes requested. Will approve once addressed.

gitops/snapshot_test.go
```diff
@@ -769,7 +769,18 @@ var _ = Describe("Gitops functions for managing Snapshots", Ordered, func() {
...
 		Expect(canBePromoted).To(BeFalse())
 		Expect(reasons).To(HaveLen(3))
+	})
 
+	It("does not promote the Snapshot when auto-release label is false even if otherwise eligible", func() {
+		err := gitops.MarkSnapshotIntegrationStatusAsFinished(ctx, k8sClient, hasSnapshot, "Test message")
+		Expect(err).ToNot(HaveOccurred())
+		err = gitops.MarkSnapshotAsPassed(ctx, k8sClient, hasSnapshot, "Test message")
+		Expect(err).ToNot(HaveOccurred())
+		hasSnapshot.Labels[gitops.AutoReleaseLabel] = "false"
```
kasemAlem: we better use `DeepCopy()` to avoid flakiness here  

see examples: 

```
  // release_test.go — correct                                                                                                                                                                                     
  snapshotCopy := hasSnapshot.DeepCopy()                                                                                                                                                                           
  snapshotCopy.Labels[gitops.AutoReleaseLabel] = "false"                                                                                                                                                           
                                                                                                                                                                                                                   
  // loader_test.go — correct                                                                                                                                                                                      
  snap := hasSnapshot.DeepCopy()                                                                                                                                                                                   
  snap.Labels[gitops.AutoReleaseLabel] = "false"   
```

kasemAlem:
wouldn't make better regression coverage if we add unit test 
testing `auto-release=true` on snapshot still allows release?

NigelByrne1:
@kasemAlem Thank you! Extra unit test added

kasemAlem (approved):

loader/loader.go
```diff
@@ -493,6 +498,11 @@ func (l *loader) GetAutoReleasePlansForComponentGroup(ctx context.Context, c cli
 		return nil, err
 	}
 
+	// snapshot level veto
+	if gitops.IsSnapshotAutoReleaseDisabled(snapshot) {
```
dirgim: I don't think that the loader function should know about the Snapshot auto release disabling logic.
I think this should be checked before the loader function is called in the first place.

dirgim (approved):
LGTM

kasemAlem (approved):

jsztuka (approved):
lgtm

14rcole (approved):

PR #1540, opened 2026-04-22
feat(STONEINTG-1548): gate auto-release on build pipeline SHOULD_RELEASE

Josh-Everett:
When a build plr sets its SHOULD_RELEASE result to false integration service now skips Release CR creation. A new shouldRelease() CEL function is available in ReleasePlan auto-release annotations.

### DIFF
--- a/gitops/release_test.go
+++ b/gitops/release_test.go
@@ -73,7 +73,7 @@ var _ = Describe("Auto-release annotation evaluation", Ordered, func() {
 	It("returns false when the auto-release annotation is missing", func() {
 		snapshotCopy := hasSnapshot.DeepCopy()
 		autoReleaseAnnotation := ""
-		allowed, err := gitops.EvaluateSnapshotAutoReleaseAnnotation(autoReleaseAnnotation, snapshotCopy)
+		allowed, err := gitops.EvaluateSnapshotAutoReleaseAnnotation(autoReleaseAnnotation, snapshotCopy, true)
 		Expect(err).NotTo(HaveOccurred())
 		Expect(allowed).To(BeFalse())
 	})
@@ -81,7 +81,7 @@ var _ = Describe("Auto-release annotation evaluation", Ordered, func() {
 	It("returns true when the auto-release annotation is 'true'", func() {
 		snapshotCopy := hasSnapshot.DeepCopy()
 		autoReleaseAnnotation := "true"
-		allowed, err := gitops.EvaluateSnapshotAutoReleaseAnnotation(autoReleaseAnnotation, snapshotCopy)
+		allowed, err := gitops.EvaluateSnapshotAutoReleaseAnnotation(autoReleaseAnnotation, snapshotCopy, true)
 		Expect(err).NotTo(HaveOccurred())
 		Expect(allowed).To(BeTrue())
 	})
@@ -89,7 +89,7 @@ var _ = Describe("Auto-release annotation evaluation", Ordered, func() {
 	It("returns false when the auto-release annotation is 'false'", func() {
 		snapshotCopy := hasSnapshot.DeepCopy()
 		autoReleaseAnnotation := "false"
-		allowed, err := gitops.EvaluateSnapshotAutoReleaseAnnotation(autoReleaseAnnotation, snapshotCopy)
+		allowed, err := gitops.EvaluateSnapshotAutoReleaseAnnotation(autoReleaseAnnotation, snapshotCopy, true)
 		Expect(err).NotTo(HaveOccurred())
 		Expect(allowed).To(BeFalse())
 	})
@@ -98,15 +98,15 @@ var _ = Describe("Auto-release annotation evaluation", Ordered, func() {
 		snapshotCopy := hasSnapshot.DeepCopy()
 		snapshotCopy.CreationTimestamp = metav1.NewTime(time.Date(2025, 1, 1, 0, 0, 0, 0, time.UTC))
 		autoReleaseAnnotation := "timestamp(snapshot.metadata.creationTimestamp) < (timestamp(now) - duration('168h'))"
-		allowed, err := gitops.EvaluateSnapshotAutoReleaseAnnotation(autoReleaseAnnotation, snapshotCopy)
+		allowed, err := gitops.EvaluateSnapshotAutoReleaseAnnotation(autoReleaseAnnotation, snapshotCopy, true)
 		Expect(err).NotTo(HaveOccurred())
 		Expect(allowed).To(BeTrue())
 	})
 
-	It("returns true  when the updated component is 'component-sample'", func() {
+	It("returns true when the updated component is 'component-sample'", func() {
 		snapshotCopy := hasSnapshot.DeepCopy()
 		autoReleaseAnnotation := "updatedComponentIs('component-sample')"
-		allowed, err := gitops.EvaluateSnapshotAutoReleaseAnnotation(autoReleaseAnnotation, snapshotCopy)
+		allowed, err := gitops.EvaluateSnapshotAutoReleaseAnnotation(autoReleaseAnnotation, snapshotCopy, true)
 		Expect(err).NotTo(HaveOccurred())
 		Expect(allowed).To(BeTrue())
 	})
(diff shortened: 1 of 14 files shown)


loader/loader.go
```diff
@@ -458,12 +458,15 @@ func (l *loader) GetAutoReleasePlansForApplication(ctx context.Context, c client
 		annotationValue, ok := rp.GetAnnotations()[gitops.AutoReleaseLabel] // label and annotation have same value
 		if ok || annotationValue != "" {
 			// If the annotation exists we evaluate the expression and return that
-			canRelease, err := gitops.EvaluateSnapshotAutoReleaseAnnotation(annotationValue, snapshot)
+			canRelease, err := gitops.EvaluateSnapshotAutoReleaseAnnotation(annotationValue, snapshot, shouldRelease)
 			if err != nil || !canRelease {
 				filteredReleasePlans.Items = append(filteredReleasePlans.Items, rp)
 			}
```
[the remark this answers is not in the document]
Josh-Everett: fixed

loader/loader.go
```diff
@@ -491,12 +494,15 @@ func (l *loader) GetAutoReleasePlansForComponentGroup(ctx context.Context, c cli
 		annotationValue := rp.GetAnnotations()[gitops.AutoReleaseLabel] // label and annotation have same value
 		if annotationValue != "" {
 			// If the annotation exists we evaluate the expression and return that
-			canRelease, err := gitops.EvaluateSnapshotAutoReleaseAnnotation(annotationValue, snapshot)
+			canRelease, err := gitops.EvaluateSnapshotAutoReleaseAnnotation(annotationValue, snapshot, shouldRelease)
 			if err != nil || !canRelease {
 				filteredReleasePlans.Items = append(filteredReleasePlans.Items, rp)
 			}
```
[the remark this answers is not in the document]
Josh-Everett: fixed

loader/loader.go
```diff
@@ -457,13 +457,16 @@ func (l *loader) GetAutoReleasePlansForApplication(ctx context.Context, c client
...
 		if ok || annotationValue != "" {
-			// If the annotation exists we evaluate the expression and return that
-			canRelease, err := gitops.EvaluateSnapshotAutoReleaseAnnotation(annotationValue, snapshot)
-			if err != nil || !canRelease {
+			// If the annotation exists we evaluate the CEL expression and include the plan when it allows release
+			canRelease, err := gitops.EvaluateSnapshotAutoReleaseAnnotation(annotationValue, snapshot, shouldRelease)
+			if err == nil && canRelease {
 				filteredReleasePlans.Items = append(filteredReleasePlans.Items, rp)
 			}
 		} else if !metadata.HasLabelWithValue(&rp, gitops.AutoReleaseLabel, "false") {
```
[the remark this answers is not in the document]
Josh-Everett: fixed

loader/loader.go
```diff
@@ -490,13 +493,16 @@ func (l *loader) GetAutoReleasePlansForComponentGroup(ctx context.Context, c cli
...
 		if annotationValue != "" {
-			// If the annotation exists we evaluate the expression and return that
-			canRelease, err := gitops.EvaluateSnapshotAutoReleaseAnnotation(annotationValue, snapshot)
-			if err != nil || !canRelease {
+			// If the annotation exists we evaluate the CEL expression and include the plan when it allows release
+			canRelease, err := gitops.EvaluateSnapshotAutoReleaseAnnotation(annotationValue, snapshot, shouldRelease)
+			if err == nil && canRelease {
 				filteredReleasePlans.Items = append(filteredReleasePlans.Items, rp)
 			}
 		} else if !metadata.HasLabelWithValue(&rp, gitops.AutoReleaseLabel, "false") {
```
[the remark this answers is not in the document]
Josh-Everett: fixed in latest update

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -573,7 +573,9 @@ func (a *Adapter) EnsureAllReleasesExist() (controller.OperationResult, error) {
 		return controller.ContinueProcessing()
 	}
 
-	releasePlans, err := a.getAutoReleasePlans()
+	shouldRelease := a.getShouldReleaseFromBuildPLR()
```
dirgim: Unfortunately this is not going to work reliably because we can't guarantee that the build pipelineRun won't be pruned by the time Snapshot has finished testing.

Because of the above, the plan is to first save the SHOULD_RELEASE result in an build pipeline's annotation and then copy it over to the Snapshot so it can be reliably checked at this point.

gitops/release_test.go
```diff
@@ -98,24 +98,60 @@ var _ = Describe("Auto-release annotation evaluation", Ordered, func() {
...
+
+	Context("shouldRelease() CEL function", func() {
+		It("returns true when shouldRelease is true", func() {
+			snapshotCopy := hasSnapshot.DeepCopy()
+			allowed, err := gitops.EvaluateSnapshotAutoReleaseAnnotation("shouldRelease()", snapshotCopy, true)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(allowed).To(BeTrue())
+		})
+
+		It("returns false when shouldRelease is false", func() {
```
jsztuka, line 130: What happens when shouldRelease does not exist, or is neither true or false?

gitops/release_test.go
```diff
@@ -98,24 +98,60 @@ var _ = Describe("Auto-release annotation evaluation", Ordered, func() {
...
+
+			allowed, err = gitops.EvaluateSnapshotAutoReleaseAnnotation(expr, snapshotCopy, false)
+			Expect(err).NotTo(HaveOccurred())
+			Expect(allowed).To(BeFalse())
+		})
+
+		It("can be bypassed with a CEL expression that doesn't call it", func() {
+			snapshotCopy := hasSnapshot.DeepCopy()
+			// Even with shouldRelease=false, a "true" expression bypasses gating
+			allowed, err := gitops.EvaluateSnapshotAutoReleaseAnnotation("true", snapshotCopy, false)
```
jsztuka, line 152: We might want to document this someplace to avoid confusion.

gitops/release_test.go
Josh-Everett: From the cel function's perspective, `shouldRelease()` always returns a boolean so

1. Annotation missing (no SHOULD_RELEASE result was set by the build pipeline): `getShouldRelease()` in the snapshot adapter returns true                                   
2. Annotation present but not "false" (e.g. "true", empty, any other string): returns true
3. Annotation is "false": returns false

The defaulting-to-true behavior is intentional so releases aren't blocked when the build pipeline doesn't produce a SHOULD_RELEASE result

I updated the comment`for `GetShouldRelease()` to better detail this.

gitops/release_test.go
Josh-Everett: Updated the comments at to give some better context at `EvaluateSnapshotAutoReleaseAnnotation()` function, let me know your thoughts.

jsztuka: thanks

loader/loader.go
```diff
@@ -490,13 +496,19 @@ func (l *loader) GetAutoReleasePlansForComponentGroup(ctx context.Context, c cli
...
-			if err != nil || !canRelease {
+			// If the annotation exists we evaluate the CEL expression and include the plan when it allows release
+			canRelease, err := gitops.EvaluateSnapshotAutoReleaseAnnotation(annotationValue, snapshot, shouldRelease)
+			if err != nil {
+				return nil, fmt.Errorf("failed to evaluate auto-release CEL expression for ReleasePlan %s: %w", rp.Name, err)
+			}
+			if canRelease {
 				filteredReleasePlans.Items = append(filteredReleasePlans.Items, rp)
 			}
 		} else if !metadata.HasLabelWithValue(&rp, gitops.AutoReleaseLabel, "false") {
```
jsztuka, line 507: I have troubles following this logic, first we get autoreleaseLabel, if it is non-empty we then evaluate if the content of annotation states should-release, in case it does append release plan item. Or we check if Autorelease label does not equal to "false" for the same release plan we checked Autorelease label earlier. 
Is there a possibility that release plan does not contain autorelease annotation but contains only autorelease label?

Josh-Everett: Yes sorry it does make it confusing to share the auto-release key for label and annotation but it's so we are backwards compatible: ReleasePlans that existed before this label was introduced still get auto-released without needing to be updated.
[Annotation path](https://github.com/konflux-ci/integration-service/pull/1540/changes#diff-8e9af1e483632eac17610237caaa5f879ea7bfb3e1e1ea228e45a17c770b9432R458) checks if the auto-release annotation has a non-empty value. If it does treat it as a CEL expression, evaluate it, and include it if the expression returns true.

The [label path](https://github.com/konflux-ci/integration-service/pull/1540/changes#diff-8e9af1e483632eac17610237caaa5f879ea7bfb3e1e1ea228e45a17c770b9432L465) runs only if the annotation is empty/missing, fall through to checking the auto-release label. Includes the plan unless the label is explicitly false.

So yes a ReleasePlan with no annotation and only the label auto-release: true hits the label path.

Following the chain makes it a little easier:

1. Build pipeline finishes and sets SHOULD_RELEASE in its Status.Results
2. CopyBuildPipelineRunResultsToSnapshot copies all results as Snapshot annotations (including result-should-release) 
3. Snapshot adapter's getShouldRelease() reads that annotation — returns false only if the value is "false", true otherwise...
4. That bool is passed to the loader's GetAutoReleasePlansForApplication, which uses it in both the label path (direct gating) and the annotation path (passed to the CEL evaluator as shouldRelease).

jsztuka (approved):
LGTM

jencull (approved):
lgtm

kasemAlem (approved):

PR #1565, opened 2026-05-13
feat: add snapshot options struct

14rcole:
Abstracts commonly-used function parameters for the snapshot package into a shared opts struct. This is a commonly-used pattern in kubernetes libraries and decreases the number of values passed to each function, making the headers and calls easier to understand

### DIFF
--- a/snapshot/create_test.go
+++ b/snapshot/create_test.go
@@ -50,6 +50,7 @@ var _ = Describe("Snapshot creation functions", Ordered, func() {
 		hasAppSample      *applicationapiv1alpha1.Application
 		hasCompSample     *applicationapiv1alpha1.Component
 		logger            logr.Logger
+		opts              SnapshotOpts
 	)
 	const (
 		SampleRepoLink           = "https://github.com/devfile-samples/devfile-sample-java-springboot-basic"
@@ -279,6 +280,7 @@ var _ = Describe("Snapshot creation functions", Ordered, func() {
 		Expect(k8sClient.Status().Update(ctx, successfulTaskRun)).Should(Succeed())
 
 		logger = log.FromContext(ctx)
+		opts = NewSnapshotOpts(ctx, k8sClient, helpers.IntegrationLogger{Logger: logger}, nil)
 	})
 
 	AfterAll(func() {
@@ -294,7 +296,7 @@ var _ = Describe("Snapshot creation functions", Ordered, func() {
 
 	Context("Testing PrepareSnapshotForPipelineRun()", func() {
 		It("ensures built component includes git context from Component CR", func() {
-			expectedSnapshot, err := PrepareSnapshotForPipelineRun(ctx, k8sClient, buildPipelineRun, componentName, hasCompGroup)
+			expectedSnapshot, err := PrepareSnapshotForPipelineRun(buildPipelineRun, componentName, hasCompGroup, opts)
 			Expect(err).ToNot(HaveOccurred())
 			Expect(expectedSnapshot).NotTo(BeNil())
 			var built *applicationapiv1alpha1.SnapshotComponent
@@ -310,7 +312,7 @@ var _ = Describe("Snapshot creation functions", Ordered, func() {
 		})
 
 		It("ensures that snapshot has label pointing to build pipelinerun", func() {
-			expectedSnapshot, err := PrepareSnapshotForPipelineRun(ctx, k8sClient, buildPipelineRun, componentName, hasCompGroup)
+			expectedSnapshot, err := PrepareSnapshotForPipelineRun(buildPipelineRun, componentName, hasCompGroup, opts)
 			Expect(err).ToNot(HaveOccurred())
 			Expect(expectedSnapshot).NotTo(BeNil())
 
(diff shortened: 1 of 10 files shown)


snapshot/opts.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"context"
+
+	"github.com/konflux-ci/integration-service/helpers"
+	"github.com/konflux-ci/integration-service/loader"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type SnapshotOpts struct {
```
14rcole, line 26: The struct fields don't really seem like options. I'm open to renaming it if anyone has a better name

jencull: SnapshotParams, SnapshotEnv, SnapshotSetup.....I agree, options isn't quite the right indicator

kasemAlem: Agree with Jenn's comment `SnapshotContext, SnapshotDeps`

snapshot/opts.go
```diff
@@ -0,0 +1,40 @@
...
+	ctx    context.Context
```
kasemAlem, line 27: Storing `context.Context` in a struct is an anti-pattern , following  [Go documentation](https://pkg.go.dev/context)  

  ▎ Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it.                                                                                                 
                  
  Storing ctx in SnapshotOpts creates several problems:                                                                                                                                                            
                  
  1. Stale context risk: The SnapshotOpts is created once (e.g., at the top of EnsureSnapshotExists) and then passed to multiple functions. If any of those functions runs across an async boundary or is called   
  later, the stored context may be canceled or carrying stale values.
  2. Convention violation: Every adapter in this codebase passes ctx as the first parameter following Go convention. This PR breaks that pattern specifically for the snapshot package functions while all other   
  packages (loader, status, gitops, tekton) continue to use ctx as a first-class parameter.                                                                                                                        
  3. Hidden dependency: When reading PrepareSnapshotForPipelineRun(pipelineRun, componentName, componentGroup, opts), it's not immediately clear that this function makes API calls — the ctx is buried inside
  opts.                                                                                                                                                                                                            
                  
  Suggestion: Keep ctx as an explicit first parameter on every function and only bundle client, logger, loader in the struct. This preserves the Go idiom and is what Kubernetes libraries like controller-runtime 
  themselves do (e.g., client.Client is passed alongside ctx, never wrapping it).

snapshot/create_test.go
```diff
@@ -279,6 +280,7 @@ var _ = Describe("Snapshot creation functions", Ordered, func() {
 		Expect(k8sClient.Status().Update(ctx, successfulTaskRun)).Should(Succeed())
 
 		logger = log.FromContext(ctx)
+		opts = NewSnapshotOpts(ctx, k8sClient, helpers.IntegrationLogger{Logger: logger}, nil)
```
kasemAlem, line 283: **loader field is nil in several test callsites — no nil guard**
                                                                                                                                                                                                                   
  Files: snapshot/create_test.go:283, snapshot/gcl_test.go:227, snapshot/validation_test.go:143_                                                                                                                    
                                                                                                                                                                                                                   
  Multiple test sites create opts with nil loader:                                                                                                                                                                 
  opts = NewSnapshotOpts(ctx, k8sClient, helpers.IntegrationLogger{Logger: logger}, nil)                                                                                                                           
                                                                                                                                                                                                                   
  This works today because the specific code paths tested don't call opts.loader, but there's no compile-time or runtime guard. If someone refactors a tested function to use the loader, they'll get a nil pointer
   panic with no helpful message. Either:                                                                                                                                                                          
                                                                                                                                                                                                                   
  - Add a nil check in NewSnapshotOpts (panic early with a clear message), or                                                                                                                                      
  - Use a no-op loader implementation in tests instead of nil, or
  - Document that loader is optional (and if so, why bundle it in a "required dependencies" struct?)                                                                                                               
                                                                                                                                                                                                                   
  The fact that loader can be nil undermines the design premise that this struct bundles "common required parameters."

snapshot/validation.go
```diff
@@ -17,23 +17,21 @@ limitations under the License.
...
 )
 
 // ValidateOverrideSnapshotComponents checks that every component in an override
 // snapshot is a known (name, version) member of the given ComponentGroup, has a valid image
 // digest, and has git source fields defined. All validation failures are collected and returned
 // as a single joined error so the caller receives a complete picture in one pass.
-func ValidateOverrideSnapshotComponents(ctx context.Context, snapshot *applicationapiv1alpha1.Snapshot, componentGroup *v1beta2.ComponentGroup) error {
-	log := log.FromContext(ctx)
+func ValidateOverrideSnapshotComponents(snapshot *applicationapiv1alpha1.Snapshot, componentGroup *v1beta2.ComponentGroup, opts SnapshotOpts) error {
+	log := opts.logger.Logger
```
kasemAlem, line 34: Before this PR, `ValidateOverrideSnapshotComponents` took (ctx, snapshot, componentGroup) — clean and minimal. After, it takes the full SnapshotOpts but only uses `opts.logger.Logger.` It doesn't use opts.client, 
  `opts.loader,` or opts.ctx. This widens the API surface unnecessarily and gives the impression the function does I/O when it's actually a pure validation function.                                                
                                                                                                                                                                                                                   
  **Suggestion**: Functions that only need a logger should just take a logger, not the full opts struct. The "opts for everything" approach trades parameter count for semantic clarity.

snapshot/create_test.go
14rcole: >  why bundle it in a "required dependencies" struct?

I wouldn't consider this a "required dependencies" struct but a "common dependencies" struct. Some of the functions don't use the loader at all so passing it shouldn't necessarily be required. That said, it (and the other three items from the adapter struct) are used commonly enough that passing four different parameters makes the code less readable

snapshot/validation.go
14rcole: That's a good point, I'll revert this header change

jsztuka:
Does this anyhow affect the way we approach snapshots?

jsztuka (approved):
lgtm

14rcole:
Given some of the feedback here I decided it was more work than it was worth to abstract just a few parameters

PR #1549, opened 2026-04-28
feat(STONEINTG-1519):  create PR group snapshots from ComponentGroups

dirgim:
* If an incoming build pipeline's component is part of a component group and if the build PLR belongs to a PR group create the group Snapshot based on the componentGroup(s) that the component belongs to
* Maintain feature parity with the current (application-based) PR group implementation
* Reduce code complexity for initial integration test status reporting in buildpipeline adapter's EnsureIntegrationTestReportedToGitProvider by extracting the logic into reportIntegrationStatusAndHandleGroups functions for, split by application and componentGroup mode
* Add a separate loader function for fetching build pipelineRuns for applications
* Add filterPipelineRunsForComponentGroups that ensures that build pipelineRuns for a PR group also match any of the componentGroups for that component
* Update unit tests to test for multiple component groups
* Add unit tests to cover the controller adapters expansion to cover PR groups for componentGroups

## Maintainers will complete the following section

- [ ] Commit messages are descriptive enough ([hints](https://www.freecodecamp.org/news/how-to-write-better-git-commit-messages/))
- [ ] Code coverage from testing does not decrease and new code is covered ([check the PR coverage on codecov](https://app.codecov.io/gh/konflux-ci/integration-service/pulls))
- [ ] [Controllers diagrams](https://github.com/konflux-ci/integration-service/tree/main/docs) are updated when PR changes controllers code  (if applicable)

PR #1587, opened 2026-06-04
feat(stoneintg-1350): add support for nested componentGroups

14rcole:
Allows users to add a ComponentGroup to another ComponentGroup's
    spec.Components field. The resulting data structure is a DAG in which
    all leaf nodes are Components and all non-leaves are ComponentGroups.
    
    When a Component is built, its ComponentGroups will be gathered and
    snapshots will be created for them. The Snapshot will contain the GCL
    and updated images for that ComponentGroup and all ComponentGroups
    nested within it. After that snapshot is created the snapshot controller
    will create snapshots for all ComponentGroups that contain the first
    ComponentGroup.
    
    This allows Konflux to support complex use-cases for ComponentGroups in
    which users have multi-tiered applications with subgroups that must be
    tested independently


snapshot/components.go
```diff
@@ -0,0 +1,29 @@
...
+		version := snapshot.Annotations[tektonconsts.PipelineRunComponentVersionAnnotation]
+		for _, snapshotComponent := range snapshot.Spec.Components {
+			if snapshotComponent.Name == component {
+				if version == "" || snapshotComponent.Version == version {
+					return []applicationapiv1alpha1.SnapshotComponent{snapshotComponent}
+				}
+			}
+		}
+	} else if gitops.IsGroupSnapshot(snapshot) {
+		// TODO: fill in code once PR  1549 is merged
```
14rcole: @dirgim FYI

loader/loader.go
```diff
@@ -941,3 +941,54 @@ func (l *loader) GetPRComponentSnapshotsForComponentApplication(ctx context.Cont
...
+func (l *loader) GetNestedComponentGroupsForComponentGroup(ctx context.Context, c client.Client, componentGroup *v1beta2.ComponentGroup) (*[]v1beta2.ComponentGroup, error) {
+	namespace := componentGroup.Namespace
+	var nestedComponentGroups []v1beta2.ComponentGroup
+	for _, component := range componentGroup.Spec.Components {
+		if strings.EqualFold(component.Kind, "componentgroup") {
+			nestedComponentGroup, err := l.GetComponentGroup(ctx, c, component.Name, namespace)
+			if err != nil {
+				return nil, fmt.Errorf("Could not get componentGroup %s nested in componentGroup %s: %+v", component.Name, componentGroup.Name, err)
+			}
+			nestedComponentGroups = append(nestedComponentGroups, *nestedComponentGroup)
```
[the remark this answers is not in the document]
14rcole: I'm not worried about this since the loader_mock package is only used for unit tests.  A nil pointer panic gets to the root of the problem faster in the rare case we accidentally write a unit test and do not pass the necessary mock values

snapshot/components.go
```diff
@@ -0,0 +1,34 @@
...
+func GetAllNewComponentsInSnapshot(snapshot *applicationapiv1alpha1.Snapshot) []applicationapiv1alpha1.SnapshotComponent {
+	if gitops.IsComponentSnapshot(snapshot) {
+		component := snapshot.Labels[gitops.SnapshotComponentLabel]
+		version := snapshot.Annotations[tektonconsts.PipelineRunComponentVersionAnnotation]
+		for _, snapshotComponent := range snapshot.Spec.Components {
+			if snapshotComponent.Name == component {
+				if version == "" || snapshotComponent.Version == version {
+					return []applicationapiv1alpha1.SnapshotComponent{snapshotComponent}
+				}
+			}
```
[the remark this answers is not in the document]
14rcole: Group snapshot support will be implemented in this ticket or #1549, whichever is merged last

internal/controller/snapshot/snapshot_adapter.go
```diff
@@ -97,6 +99,83 @@ func NewAdapter(context context.Context, snapshot *applicationapiv1alpha1.Snapsh
 	}
 }
 
+// EnsureParentSnapshotsExist is responsible for creating snapshots for ComponentGroups that contain the ComponentGroup for this snapshot
+// TODO: add unit tests
```
jsztuka: Is this planned in the future or is just leftover?

snapshot/create.go
```diff
@@ -39,12 +41,16 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/log"
 )
 
+const (
+	MaxDepth = 15
```
jsztuka, line 46: Is this a depth of nested componentgroups? How did you figure out this number?

snapshot/create.go
```diff
@@ -174,20 +186,106 @@ func PrepareSnapshot(ctx context.Context, adapterClient client.Client, component
...
+	newSnapshotComponents := GetAllNewComponentsInSnapshot(childSnapshot)
+	upsertMultipleComponentImages(snapshotComponentsMap, invalidComponents, newSnapshotComponents, logger)
+	snapshotComponents := flattenSnapshotComponentsMap(snapshotComponentsMap)
+
+	if len(snapshotComponents) == 0 {
+		return nil, helpers.NewMissingValidComponentError(joinInvalidComponentNamesAndVersions(invalidComponents))
+	}
+	snapshot := NewSnapshot(componentGroup, &snapshotComponents)
+
+	// expose the source repo URL and SHA in the snapshot as annotation do we don't have to do lookup in integration tests
```
jsztuka: do > so

snapshot/create.go
14rcole: This is just an arbitrary number. There's a risk of stack overflow if there are too many nested layers of recursion so this prevents that from happening. I can set it to another number but I don't think many users will have 15 levels of nested ComponentGroups

internal/controller/snapshot/snapshot_adapter.go
14rcole: Just leftover. I'll delete it

snapshot/create.go
```diff
@@ -174,20 +186,106 @@ func PrepareSnapshot(ctx context.Context, adapterClient client.Client, component
...
-	err := ctrl.SetControllerReference(componentGroup, snapshot, adapterClient.Scheme())
+	err = ctrl.SetControllerReference(componentGroup, snapshot, adapterClient.Scheme())
 	if err != nil {
 		return nil, err
 	}
 
 	return snapshot, nil
 }
 
+func getNestedSnapshotComponents(componentGroup *v1beta2.ComponentGroup, usedComponentGroups []string, depth int, loader loader.ObjectLoader, ctx context.Context, adapterClient client.Client, logger helpers.IntegrationLogger) (map[string]applicationapiv1alpha1.SnapshotComponent, map[v1beta2.ComponentState]InvalidComponentReason, error) {
```
jsztuka: We could add function description, even though the name is self-explanatory.

snapshot/create.go
```diff
@@ -245,47 +351,39 @@ func getSpecComponentsAndVersionsMap(componentGroup *v1beta2.ComponentGroup) map
...
-				// remove component from invalid list
-				*snapshotComponents = append(*snapshotComponents, updatedComponent)
-				// We don't care about the loop skipping problem here because we always exit
-				// immediately after deleting the component from the list
-				*invalidComponents = slices.Delete(*invalidComponents, i, i+1)
-
-				return
-			}
-		}
+// TODO: add unit tests
```
jsztuka: This line can be removed since Unit tests were created.

jsztuka (approved):
lgtm

snapshot/components.go
```diff
@@ -0,0 +1,87 @@
...
+			return snapshotComponents, err
+		}
+		snapshotComponents = append(snapshotComponents, updatedComponent)
+	} else if gitops.IsGroupSnapshot(snapshot) {
+		var componentSnapshotInfos []*gitops.ComponentSnapshotInfo
+		var err error
+		if componentSnapshotInfoString, ok := snapshot.Annotations[gitops.GroupSnapshotInfoAnnotation]; ok {
+			componentSnapshotInfos, err = gitops.UnmarshalJSON([]byte(componentSnapshotInfoString))
+			if err != nil {
+				return snapshotComponents, fmt.Errorf("failed to unmarshal JSON string: %w", err)
```
kasemAlem:
```suggestion
				return snapshotComponents, fmt.Errorf("failed to unmarshal JSON string: %+v", err)
```

kasemAlem:
beside the small nit, approved
well done @14rcole

snapshot/components.go
```diff
@@ -0,0 +1,87 @@
...
+			}
+		}
+		for _, info := range componentSnapshotInfos {
+			// Get ContainerImage, URL, and Revision from Snapshot
+
+			snapshot, err := loader.GetSnapshot(ctx, adapterClient, info.Namespace, info.Snapshot)
```
dirgim: Do we need to fetch the original component Snapshot here?
Shouldn't the updated component be already present in the group Snapshot?

snapshot/create.go
```diff
@@ -39,12 +41,17 @@ import (
...
+
 // PrepareSnapshotForPipelineRun prepares the Snapshot for a given PipelineRun,
 // component and application. In case the Snapshot can't be created, an error will be returned.
-func PrepareSnapshotForPipelineRun(ctx context.Context, adapterClient client.Client, pipelineRun *tektonv1.PipelineRun, componentName string, componentGroup *v1beta2.ComponentGroup) (*applicationapiv1alpha1.Snapshot, error) {
-	log := log.FromContext(ctx)
+func PrepareSnapshotForPipelineRun(ctx context.Context, adapterClient client.Client, pipelineRun *tektonv1.PipelineRun, componentName string, componentGroup *v1beta2.ComponentGroup, loader loader.ObjectLoader) (*applicationapiv1alpha1.Snapshot, error) {
+	logger := helpers.IntegrationLogger{Logger: log.FromContext(ctx)}
 
-	newSnapshotComponent, err := getSnapshotComponentFromBuildPLR(pipelineRun, componentName, log)
+	newSnapshotComponent, err := GetSnapshotComponentFromBuildPLR(pipelineRun, componentName, logger)
```
dirgim, line 54: Wondering what happens in the case where the parent componentGroup has multiple child componentGroups?

internal/controller/snapshot/snapshot_controller.go
```diff
@@ -183,6 +183,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		adapter.EnsureGlobalCandidateImageUpdated,
 		adapter.EnsureRerunPipelineRunsExist,
 		adapter.EnsureIntegrationPipelineRunsExist,
+		adapter.EnsureParentSnapshotsExist,
```
dirgim, line 186: I think `EnsureParentSnapshotsExist` should also be added to the AdapterInterface below.

snapshot/create.go
dirgim: Disregard this comment, GitHub threw me on a complerely different line than intended.
I realized that `getNestedSnapshotComponents` goes into nested ComponentGroups recursively in the meantime.

snapshot/create.go
```diff
@@ -174,20 +187,112 @@ func PrepareSnapshot(ctx context.Context, adapterClient client.Client, component
...
 	if err != nil {
 		return nil, err
 	}
 
 	return snapshot, nil
 }
 
+// getNestedSnapshotComponents gets the GCL SnapshotComponents for a ComponentGroup. It recurses into the nested ComponentGroups to get their SnapshotComponents then builds upward from there.
+// As a result, if a parent and child ComponentGroup share a ComponentVersion but have different images, the image in the parent will supersede the one in the child.
+func getNestedSnapshotComponents(componentGroup *v1beta2.ComponentGroup, usedComponentGroups []string, depth int, loader loader.ObjectLoader, ctx context.Context, adapterClient client.Client, logger helpers.IntegrationLogger) (map[string]applicationapiv1alpha1.SnapshotComponent, map[v1beta2.ComponentState]InvalidComponentReason, error) {
```
dirgim: The name of this function threw me for a loop initially where I assumed it would only fetch the SnapshotComponents from a single nested group.
Maybe a simple rename to `getAllNestedSnapshotComponents` could help with comprehension for this one?

snapshot/components.go
14rcole: You're right, that makes things much simpler.  I'll fix that

snapshot/create.go
14rcole: Here? Nothing. in `getNestedSnapshotComponents` though it'll just go through each of them in order.  It's doing a DFS though so it'll go through the entirety of child tree 1 before going to child tree 2

dirgim (approved):
LGTM

kasemAlem (approved):

PR #1591, opened 2026-06-07
feat(STONEINTG-1659): define NudgeConfig CRD with stateless CEL valid…

kasemAlem:
…ation

- Add NudgeConfig CRD types in api/v1beta2 (NudgeConfig, NudgeRelationship, NudgeConfigSpec, NudgeConfigStatus) with NudgeModeType enum (immediate/validated)
- Enforce singleton naming via CEL rule (must be named "nudge-config")
- Reject self-nudges (from == to) and duplicate (from, to) pairs via CEL
- Cap nudges at maxItems=256 and component names at maxLength=63 to stay within the API server's CEL cost budget for the O(n²) duplicate rule
- Default mode to "immediate" via kubebuilder schema defaulting
- Bump ENVTEST_K8S_VERSION 1.23 → 1.29 (CEL validation GA)
- Generate CRD YAML, deepcopy, RBAC roles, and sample manifests
- Add pure Go struct tests (7 cases) and envtest CEL validation tests (8 cases)

Assicted_By: Claude, Opus 4.6


api/v1beta2/integrationtestscenario_suite_test.go
```diff
@@ -18,16 +18,9 @@ package v1beta2
 
 import (
 	"context"
```
14rcole, line 20: Why were large parts of this file deleted?

kasemAlem: nice catch, 
by mistake , file reverted

kasemAlem: @14rcole 
 The webhook server setup in this suite was unused — no test in api/v1beta2/ registers webhook handlers. After reverting the deletion, CI confirmed it's actually broken: the suite installs webhook configs 
 from config/webhook/ but has no matching handlers, so the health check times out with "connection refused" and all tests are skipped. The file itself is still needed as the Ginkgo suite bootstrap (envtest, 
k8sClient, ctx) for the NudgeConfig CEL validation tests. The ENVTEST_K8S_VERSION bump to 1.29 is also required here — CEL validation is GA starting in K8s 1.29, and the NudgeConfig CRD relies entirely on 
CEL rules (no webhook). The old 1.23 envtest would silently ignore them.

cc @dirgim

dirgim: While I agree that the webhook server setup is unneeded, but since it's technically out of the scope of this change, @kasemAlem could you put it in a separate commit and add a quick explanation in the commit message?

dirgim (approved):
LGTM aside the nit about having the webhook setup in the test suite change as a separate commit.

api/v1beta2/integrationtestscenario_suite_test.go
```diff
@@ -72,23 +61,16 @@ var _ = BeforeSuite(func() {
 			),
```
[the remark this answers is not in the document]
kasemAlem: done,

api/v1beta2/nudgeconfig_validation_test.go
```diff
@@ -0,0 +1,199 @@
...
+				},
+			},
+		}
+		Expect(k8sClient.Create(ctx, nc)).To(Succeed())
+
+		created := &NudgeConfig{}
+		Expect(k8sClient.Get(ctx, types.NamespacedName{Name: NudgeConfigSingletonName, Namespace: "default"}, created)).To(Succeed())
+		Expect(created.Spec.Nudges[0].Mode).To(Equal(NudgeModeImmediate))
+	})
+
```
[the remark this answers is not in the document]
kasemAlem: nice catch 
added

Josh-Everett (approved):

api/v1beta2/nudgeconfig_types.go
```diff
@@ -0,0 +1,116 @@
...
+const NudgeConfigSingletonName = "nudge-config"
+
+// NudgeRelationship defines a single nudge from one component to another.
+type NudgeRelationship struct {
+	// From is the source component name that triggers the nudge when its build succeeds.
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=63
+	// +kubebuilder:validation:Pattern=`^[a-z0-9]([-a-z0-9]*[a-z0-9])?$`
+	// +required
+	From string `json:"from"`
```
14rcole, line 45: Shouldn't we be nudging between ComponentVersions?

If so, we may also want to support a wildcard for the version

docs/STONEINTG-1659-plan.md
```diff
@@ -0,0 +1,339 @@
+# STONEINTG-1659: Define NudgeConfig CRD schema with stateless CEL validation rules
```
14rcole, line 1: Is this document supposed to be included? It looks like an artifact from an Agent that assisted in generating the code

api/v1beta2/nudgeconfig_types.go
kasemAlem: Good point,  
Yet, per [ADR-0067](https://github.com/konflux-ci/architecture/blob/main/ADR/0067-nudging-relationship-singleton-crd.md), nudges are defined as component-to-component edges — from and to reference Component names, not `ComponentVersions`. If we want version-scoped nudging, that would need an ADR update, mean we go back and raise it in discussion with the design stakeholders.

docs/STONEINTG-1659-plan.md
kasemAlem: yes, 
my bad will be removed in the [follow up pr ](https://github.com/konflux-ci/integration-service/pull/1596)

PR #1594, opened 2026-06-14
feat(STONEINTG-1660): add ValidatingWebhook for NudgeConfig cycle detection

kasemAlem:
What
  
  Adds a ValidatingWebhook (failurePolicy: Fail) that rejects NudgeConfig CREATE/UPDATE operations when spec.nudges contains a cycle.

  Why

  CEL validation (from STONEINTG-1659) handles self-nudges and duplicate pairs, but cannot perform transitive graph traversal due to API server cost budget limits. A webhook is needed to catch cycles like
  A→B→C→A at admission time.

  How

  - Cycle detection (pkg/dag/verify.go): DFS with three-color marking on the nudge edge list. Reports the full cycle path in the error message (e.g. "cycle detected: a -> b -> c -> a"). Existing
  ValidateTestGraph/checkGraphForCycles are untouched.
  - Webhook (internal/webhook/v1beta2/nudgeconfig_webhook.go): Validates on CREATE and UPDATE (skips if nudges unchanged). Follows the same pattern as the ComponentGroup webhook.
  - Registration: Added to cmd/main.go and webhook test suite.

  Tests

  - 12 DAG unit tests: linear chain, diamond DAG, direct cycle, transitive cycle, self-loop, empty, disconnected graph
  - 12 webhook tests: valid/invalid creates, updates introducing cycles, unchanged updates, delete no-op, wrong type assertions

NOTE:  This PR can be merged only after [PR#1591](https://github.com/konflux-ci/integration-service/pull/1591)

### DIFF
--- a/internal/webhook/v1beta2/nudgeconfig_webhook.go
+++ b/internal/webhook/v1beta2/nudgeconfig_webhook.go
@@ -1,0 +1,96 @@
+/*
+Copyright 2026 Red Hat Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1beta2
+
+import (
+	"context"
+	"fmt"
+	"reflect"
+
+	"github.com/konflux-ci/integration-service/api/v1beta2"
+	"github.com/konflux-ci/integration-service/pkg/dag"
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	logf "sigs.k8s.io/controller-runtime/pkg/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+// nolint:unused
+var nudgeconfiglog = logf.Log.WithName("nudgeconfig-webhook")
+
+func SetupNudgeConfigWebhookWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewWebhookManagedBy(mgr).
+		For(&v1beta2.NudgeConfig{}).
+		WithValidator(&NudgeConfigCustomValidator{Client: mgr.GetClient()}).
+		Complete()
+}
+
+// NudgeConfigCustomValidator is a webhook handler and does not need deepcopy methods. (validator = validating webhook)
+// +k8s:deepcopy-gen=false
+type NudgeConfigCustomValidator struct {
+	Client client.Client
+}
+
+// +kubebuilder:webhook:path=/validate-appstudio-redhat-com-v1beta2-nudgeconfig,mutating=false,failurePolicy=fail,sideEffects=None,groups=appstudio.redhat.com,resources=nudgeconfigs,verbs=create;update,versions=v1beta2,name=vnudgeconfig.kb.io,admissionReviewVersions=v1
+
+var _ webhook.CustomValidator = &NudgeConfigCustomValidator{}
+
+func (v *NudgeConfigCustomValidator) ValidateCreate(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
+	nudgeConfig, ok := obj.(*v1beta2.NudgeConfig)
+	if !ok {
+		return nil, fmt.Errorf("expected a NudgeConfig object but got %T", obj)
+	}
+
+	nudgeconfiglog.Info("Validating NudgeConfig upon creation", "name", nudgeConfig.Name)
+
+	if len(nudgeConfig.Spec.Nudges) > 0 {
+		if err := dag.ValidateNudgeGraph(nudgeConfig.Spec.Nudges); err != nil {
+			return nil, fmt.Errorf("error validating nudge graph: %v", err)
+		}
+	}
+
+	return nil, nil
+}
+
+func (v *NudgeConfigCustomValidator) ValidateUpdate(ctx context.Context, oldObj, newObj runtime.Object) (admission.Warnings, error) {
+	oldNudgeConfig, ok := oldObj.(*v1beta2.NudgeConfig)
+	if !ok {
+		return nil, fmt.Errorf("expected a NudgeConfig for oldObj but got %T", oldObj)
+	}
+	newNudgeConfig, ok := newObj.(*v1beta2.NudgeConfig)
+	if !ok {
+		return nil, fmt.Errorf("expected a NudgeConfig for newObj but got %T", newObj)
+	}
+
+	nudgeconfiglog.Info("Validating NudgeConfig upon update", "name", newNudgeConfig.Name)
+
+	if !reflect.DeepEqual(oldNudgeConfig.Spec.Nudges, newNudgeConfig.Spec.Nudges) {
+		if len(newNudgeConfig.Spec.Nudges) > 0 {
+			if err := dag.ValidateNudgeGraph(newNudgeConfig.Spec.Nudges); err != nil {
+				return nil, fmt.Errorf("error validating nudge graph: %v", err)
+			}
+		}
+	}
+
+	return nil, nil
+}
+
+func (v *NudgeConfigCustomValidator) ValidateDelete(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
+	return nil, nil
+}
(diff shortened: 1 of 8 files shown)


kasemAlem:
/retest

This is a flaky infrastructure failure, not a code bug. The GitHub API returned a transient 502 at the exact moment PaC needed to verify ACL for the newly created PR.

kasemAlem:
/retest

failed to provision kind cluster

pkg/dag/verify.go
```diff
@@ -125,3 +128,90 @@ func checkGraphForCycles(testGraph map[string][]v1beta2.TestGraphNode) error {
...
+}
+
+const (
+	colorWhite = 0 // unvisited
+	colorGray  = 1 // in current DFS path
+	colorBlack = 2 // fully processed
+)
+
+// findCycleInAdjacencyList performs DFS to detect cycles and returns the cycle path.
+func findCycleInAdjacencyList(adj map[string][]string) ([]string, bool) {
```
14rcole, line 131: Would it be possible to modify and reuse the existing validation code? I think the types being compared are different so we may need to cast to a common struct, but it'll make the package far easier to maintain.

Note: the existing code uses BFS for cycle detection. I don't think it matters though

dirgim (approved):
LGTM

pkg/dag/verify.go
kasemAlem: Good call. 
 Unified both into a shared `checkGraphForCycles(adj map[string][]string) `that uses the DFS implementation. Both `ValidateTestGraph` and `ValidateNudgeGraph` now convert their typed inputs to `map[string][]string` and call it — one cycle detection algorithm for the package. The old Kahn's BFS is removed.

  As you noted, BFS vs DFS doesn't matter for correctness. DFS has the advantage of returning the actual cycle path in the error message instead of just "remaining indegrees", so I kept that one.

internal/webhook/v1beta2/nudgeconfig_webhook.go
```diff
@@ -0,0 +1,96 @@
...
+import (
+	"context"
+	"fmt"
+	"reflect"
+
+	"github.com/konflux-ci/integration-service/api/v1beta2"
+	"github.com/konflux-ci/integration-service/pkg/dag"
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
```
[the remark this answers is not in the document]
kasemAlem: Keeping the `Client` field for consistency with `ComponentGroupCustomValidator`, which follows the same pattern. Both validators receive the client from `mgr.GetClient()` at setup time — it's available if future validation logic needs to query the cluster (e.g., cross-resource cycle detection). Removing it from one but not the other would be inconsistent.

internal/webhook/v1beta2/nudgeconfig_webhook.go
```diff
@@ -0,0 +1,96 @@
+/*
+Copyright 2026 Red Hat Inc.
```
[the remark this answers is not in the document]
kasemAlem: 2026 is the correct copyright year for a newly created file. The older webhook files use 2023 because they were created in 2023. Backdating the copyright to match them would be incorrect.

internal/webhook/v1beta2/nudgeconfig_webhook.go
```diff
@@ -0,0 +1,96 @@
...
+	Client client.Client
+}
+
+// +kubebuilder:webhook:path=/validate-appstudio-redhat-com-v1beta2-nudgeconfig,mutating=false,failurePolicy=fail,sideEffects=None,groups=appstudio.redhat.com,resources=nudgeconfigs,verbs=create;update,versions=v1beta2,name=vnudgeconfig.kb.io,admissionReviewVersions=v1
+
+var _ webhook.CustomValidator = &NudgeConfigCustomValidator{}
+
+func (v *NudgeConfigCustomValidator) ValidateCreate(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
+	nudgeConfig, ok := obj.(*v1beta2.NudgeConfig)
+	if !ok || nudgeConfig == nil {
```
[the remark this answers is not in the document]
kasemAlem: Fixed — removed the `"non-nil"` qualifier to match the existing webhook error message pattern (`"expected a NudgeConfig object but got %T"`).

internal/webhook/v1beta2/nudgeconfig_webhook.go
```diff
@@ -0,0 +1,96 @@
...
+type NudgeConfigCustomValidator struct {
+	Client client.Client
+}
+
+// +kubebuilder:webhook:path=/validate-appstudio-redhat-com-v1beta2-nudgeconfig,mutating=false,failurePolicy=fail,sideEffects=None,groups=appstudio.redhat.com,resources=nudgeconfigs,verbs=create;update,versions=v1beta2,name=vnudgeconfig.kb.io,admissionReviewVersions=v1
+
+var _ webhook.CustomValidator = &NudgeConfigCustomValidator{}
+
+func (v *NudgeConfigCustomValidator) ValidateCreate(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
+	nudgeConfig, ok := obj.(*v1beta2.NudgeConfig)
```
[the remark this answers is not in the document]
kasemAlem: Fixed — removed the extra `|| nudgeConfig == nil` check to match the existing webhook validators which only check `!ok`.

Josh-Everett (approved):

internal/webhook/v1beta2/nudgeconfig_webhook.go
```diff
@@ -0,0 +1,96 @@
...
+			if err := dag.ValidateNudgeGraph(newNudgeConfig.Spec.Nudges); err != nil {
+				return nil, fmt.Errorf("error validating nudge graph: %v", err)
+			}
+		}
+	}
+
+	return nil, nil
+}
+
+func (v *NudgeConfigCustomValidator) ValidateDelete(ctx context.Context, obj runtime.Object) (admission.Warnings, error) {
```
jsztuka, line 94: I see that webhook has rbac set to create and update, is it expected that only users will be capable of deleting nudges?

internal/webhook/v1beta2/nudgeconfig_webhook_test.go
```diff
@@ -0,0 +1,178 @@
...
+
+		It("returns an error for wrong newObj type", func() {
+			wrongObj := &appstudiov1beta2.ComponentGroup{}
+			_, err := validator.ValidateUpdate(ctx, nudgeConfig, wrongObj)
+			Expect(err).To(HaveOccurred())
+			Expect(err.Error()).To(ContainSubstring("expected a NudgeConfig"))
+		})
+	})
+
+	When("a NudgeConfig is deleted", func() {
```
jsztuka, line 171: Who is deleting the nudgeconfig here?

pkg/dag/verify.go
```diff
@@ -73,55 +76,114 @@ func ValidateTestGraph(testGraph map[string][]v1beta2.TestGraphNode, scenarios .
...
-		if indegrees[node] == 0 {
-			queue = append(queue, node)
-			res = append(res, node)
+	if err := checkGraphForCycles(adj); err != nil {
+		return fmt.Errorf("invalid NudgeGraph - cycle detected: %w", err)
+	}
+	return nil
+}
+
+// checkGraphForCycles uses DFS to detect cycles in a directed graph represented
```
jsztuka: What does DFS stand for?

pkg/dag/verify_test.go
```diff
@@ -60,16 +59,102 @@ var _ = Describe("DAG library unittests", Ordered, func() {
...
+			})
+		})
+
+		When("the graph is empty", func() {
+			It("returns no error", func() {
+				Expect(checkGraphForCycles(map[string][]string{})).NotTo(HaveOccurred())
+			})
+		})
+
+		When("the graph is disconnected with a cycle in one component", func() {
```
jsztuka, line 148: Shouldn't this test also raise an error since it contains cycle?

jsztuka:
Just have a few questions, otherwise looks solid, good work!

internal/webhook/v1beta2/nudgeconfig_webhook.go
kasemAlem: Intentional — `ValidateDelete` is a no-op since deleting a `NudgeConfig` can't introduce cycles. 
Registering for **DELETE** would add a round-trip to the webhook on every deletion with no validation benefit, and since `failurePolicy=Fail`, it would also block deletions if the webhook is unavailable. The `ValidateDelete` method exists only because `webhook.CustomValidator` requires it

internal/webhook/v1beta2/nudgeconfig_webhook_test.go
kasemAlem: This test verifies the `ValidateDelete` method always returns nil (no-op). It calls the validator directly, not through the K8s API — it's a unit test of the interface method. 
Since we implement `webhook.CustomValidator`, all three methods (`Create/Update/Delete`) must exist, so we test that Delete is indeed a no-op.

pkg/dag/verify.go
kasemAlem: DFS = Depth-First Search. I'll spell it out in the comment.

pkg/dag/verify_test.go
kasemAlem: It does return an error (the `HaveOccurred()` assertion). 
But you're right that we should also verify the cycle path content like the other tests. I'll add assertions for the cycle nodes a, b,c.

jsztuka (approved):
lgtm

PR #1600, opened 2026-06-21
feat(STONEINTG-1668): create konflux-integration-runner SA if missing

kasemAlem:
Enable the scenario controller to ensure the integration pipeline ServiceAccount, components-namespace-pull Secret, and RoleBinding exist in the namespace when an IntegrationTestScenario is created.

- Re-enable the scenario controller with a create-only event predicate
- Create konflux-integration-runner SA if not present
- Create empty dockerconfigjson secret if not present and link it to the SA via ImagePullSecrets
- Create RoleBinding to the konflux-integration-runner ClusterRole
- Add RBAC permissions for serviceaccounts, secrets, and rolebindings
- Unit tests for all creation, linking, and idempotency scenarios

Assisteb-By: Claude Opus 4.6
For more info in [STONEINTG-1668](https://redhat.atlassian.net/browse/STONEINTG-1668)

[STONEINTG-1668]: https://redhat.atlassian.net/browse/STONEINTG-1668?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/scenario-controller.md
+++ b/docs/scenario-controller.md
@@ -9,34 +9,36 @@ flowchart TD
   classDef Amber fill:#FFDEAD;
   classDef Green fill:#BDFFA4;
 
-  predicate((PREDICATE: <br>Monitor IntegratonTestScenario <br>& filter created/updated/deleted <br>events for the resource ))
-  %%%%%%%%%%%%%%%%%%%%%%% Drawing EnsureCreatedScenarioIsValid() function
+  predicate((PREDICATE: <br>Monitor IntegrationTestScenario <br>& filter created events))
 
   %% Node definitions
-  
-  application_exists{"Application for scenario <br>was found?"}
-  set_owner_reference(Set owner reference to <br>IntegrationTestScenario <br> if not already existing)
-  status_check{"IntegrationTestScenario <br>has status<br>IntegrationTestScenarioValid<br>condition undefined<br>or false"}
-  update_scenario_status_valid(Update IntegrationTestScenario <br>status to valid)
-  update_scenario_status_invalid(Update IntegrationTestScenario <br>status to invalid)
-  remove_historical_finalizer(Check for and remove<br>historical IntegrationTestScenario <br> finalizer)
-  complete_reconciliation(Complete reconciliation for <br>IntegrationTestScenario)
-  continue_reconciliation(Continue with next reconciliation)
-
+  check_sa{"ServiceAccount<br>konflux-integration-runner<br>exists?"}
+  create_sa(Create ServiceAccount<br>with ImagePullSecrets)
+  check_secret{"Secret<br>components-namespace-pull<br>exists?"}
+  create_secret(Create empty<br>dockerconfigjson Secret)
+  check_linked{"Secret linked to<br>SA ImagePullSecrets?"}
+  link_secret(Update SA to link Secret)
+  check_rb{"RoleBinding<br>konflux-integration-runner<br>exists?"}
+  create_rb(Create RoleBinding to<br>ClusterRole konflux-integration-runner)
+  complete(Complete reconciliation)
 
   %% Node connections
-  predicate                        ---->    |"EnsureCreatedScenarioIsValid()"| remove_historical_finalizer
-  remove_historical_finalizer      -->      application_exists
-  application_exists               --No-->  update_scenario_status_invalid
-  application_exists               --Yes--> set_owner_reference
-  set_owner_reference              -->      status_check
-  status_check                     --No-->  update_scenario_status_invalid
-  status_check                     --Yes--> update_scenario_status_valid
-  update_scenario_status_valid     -->      complete_reconciliation
-  complete_reconciliation          -->      continue_reconciliation
-  update_scenario_status_invalid   -->      continue_reconciliation
+  predicate                  ---->  |"EnsureIntegrationPipelineServiceAccount()"| check_sa
+  check_sa                   --No-->  create_sa
+  check_sa                   --Yes--> check_secret
+  create_sa                  -->      check_secret
+  check_secret               --No-->  create_secret
+  check_secret               --Yes--> check_linked
+  create_secret              -->      check_linked
+  check_linked               --No-->  link_secret
+  check_linked               --Yes--> check_rb
+  link_secret                -->      check_rb
+  check_rb                   --No-->  create_rb
+  check_rb                   --Yes--> complete
+  create_rb                  -->      complete
 
-   %% Assigning styles to nodes
+  %% Assigning styles to nodes
   class predicate Amber;
+  class create_sa,create_secret,link_secret,create_rb Green;
 
   ```
--- a/internal/controller/scenario/predicates.go
+++ b/internal/controller/scenario/predicates.go
@@ -1,0 +1,25 @@
+package scenario
+
+import (
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+// ScenarioCreatedPredicate returns a predicate which filters out
+// only IntegrationTestScenarios that are the result of a create event.
+func ScenarioCreatedPredicate() predicate.Predicate {
+	return predicate.Funcs{
+		CreateFunc: func(createEvent event.CreateEvent) bool {
+			return true
+		},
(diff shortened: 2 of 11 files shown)


docs/scenario-controller.md
```diff
@@ -9,34 +9,36 @@ flowchart TD
...
-
+  check_sa{"ServiceAccount<br>konflux-integration-runner<br>exists?"}
+  create_sa(Create ServiceAccount<br>with ImagePullSecrets)
+  check_secret{"Secret<br>components-namespace-pull<br>exists?"}
+  create_secret(Create empty<br>dockerconfigjson Secret)
+  check_linked{"Secret linked to<br>SA ImagePullSecrets?"}
+  link_secret(Update SA to link Secret)
+  check_rb{"RoleBinding<br>konflux-integration-runner<br>exists?"}
+  create_rb(Create RoleBinding to<br>ClusterRole konflux-integration-runner)
+  complete(Complete reconciliation)
```
14rcole, line 23: Thanks for remembering to update this document!

internal/controller/scenario/predicates.go
```diff
@@ -0,0 +1,25 @@
...
+			return false
+		},
+		GenericFunc: func(genericEvent event.GenericEvent) bool {
+			return false
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return false
+		},
+	}
+}
```
14rcole, line 25: We have most of our predicates in `tekton/predicates.go`.  Should we put this predicate in the same spot?

If it's considered more idiomatic k8s code to keep the predicates in the controllers then I'm fine with leaving it here. If that's the case, can you create a ticket to move the other predicates into the controller packages?

internal/controller/scenario/scenario_adapter.go
```diff
@@ -47,6 +54,138 @@ func NewAdapter(context context.Context, scenario *v1beta2.IntegrationTestScenar
 	}
 }
 
-func (a *Adapter) EnsurePlaceholder() (controller.OperationResult, error) {
+// EnsureIntegrationPipelineServiceAccount ensures the integration pipeline service account,
+// image pull secret, and role binding exist in the IntegrationTestScenario's namespace.
+func (a *Adapter) EnsureIntegrationPipelineServiceAccount() (controller.OperationResult, error) {
```
14rcole: nit:
```suggestion
func (a *Adapter) EnsureIntegrationPipelineServiceAccountCreated() (controller.OperationResult, error) {
```

internal/controller/scenario/scenario_adapter_test.go
```diff
@@ -73,78 +73,284 @@ var _ = Describe("Scenario Adapter", Ordered, func() {
...
+			_ = k8sClient.Delete(ctx, rb)
+		})
+
+		It("creates the ServiceAccount, Secret, and RoleBinding", func() {
+			result, err := adapter.EnsureIntegrationPipelineServiceAccount()
+			Expect(err).ToNot(HaveOccurred())
+			Expect(result.CancelRequest).To(BeFalse())
+
+			sa := &corev1.ServiceAccount{}
+			err = k8sClient.Get(ctx, types.NamespacedName{
```
dirgim: We might want to wrap these checks which get the resource from the control plane into `Eventually` blocks to account for slower environments such as ones in GitHub actions.

internal/controller/scenario/predicates.go
```diff
@@ -0,0 +1,25 @@
+package scenario
+
+import (
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+)
+
+// ScenarioCreatedPredicate returns a predicate which filters out
+// only IntegrationTestScenarios that are the result of a create event.
+func ScenarioCreatedPredicate() predicate.Predicate {
```
dirgim, line 10: We might also need unit tests for this predicate if I didn't miss them somewhere else.

internal/controller/scenario/predicates.go
kasemAlem: Good point! I placed it here following the pattern already established by internal/controller/component/predicates.go which has ComponentCreatedPredicate and ComponentDeletedPredicate co-located with the component controller. The predicates in tekton/predicates.go are all PipelineRun-specific and depend on tekton package helpers (IsIntegrationPipelineRun, hasPipelineRunStateChangedTo*), so they made sense there at the time. 
 But keeping predicates next to their controllers is the more idiomatic k8s pattern — it improves discover-ability and keeps each controller self-contained.

created STONEINTG-1683

internal/controller/scenario/scenario_adapter.go
kasemAlem: you are right, 
renaming looks better one , 
done

internal/controller/scenario/scenario_adapter_test.go
kasemAlem: Good call — wrapped all Get + assertion blocks in Eventually(func(g Gomega) { ... }, "5s").Should(Succeed()) for the SA, Secret, and RoleBinding checks across all test cases. Also consolidated the RoleBinding test that had a redundant Eventually followed by a bare Get

internal/controller/scenario/predicates.go
kasemAlem: You're right, 
they were missing. Added predicates_test.go covering all four event functions: Create returns true, Update/Delete/Generic return false.

dirgim (approved):
LGTM

internal/controller/scenario/scenario_adapter.go
```diff
@@ -22,11 +22,18 @@ import (
...
+	corev1 "k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/util/retry"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-// Adapter holds the objects needed to reconcile a Release.
```
jsztuka, line 29: What is the reason behind moving logic from release to integrationTestScenario, was this adapter not used before? Is this being taken care of someplace else within those changes?

kasemAlem: This logic isn't moved from the release controller — it's new. 
The old docstring on line 29 **("Adapter holds the objects needed to reconcile a Release")** was a copy-paste error from when the adapter was originally scaffolded; this was always the scenario adapter. 
The scenario controller was disabled in [3ff066de](https://github.com/konflux-ci/integration-service/commit/3ff066de19f88e043825b638a5e45b61f38b2704) because its only function `EnsureScenarioContainsResourceKind` was no longer needed. This PR re-enables it with new SA/Secret/RoleBinding provisioning logic — the ITS create event is the natural trigger since these resources are needed to run the integration pipelines that `IntegrationTestScenarios` define.

internal/controller/scenario/scenario_adapter_test.go
```diff
@@ -29,21 +30,22 @@ import (
...
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 )
 
 var _ = Describe("Scenario Adapter", Ordered, func() {
 	var (
 		adapter                 *Adapter
 		hasApp                  *applicationapiv1alpha1.Application
 		integrationTestScenario *v1beta2.IntegrationTestScenario
-		invalidScenario         *v1beta2.IntegrationTestScenario
```
jsztuka, line 41: Why was this scenario removed?

kasemAlem: The `invalidScenario` fixture tested the old `EnsureCreatedScenarioIsValid `logic (validating the scenario's Application reference). That validation moved to the admission webhook [bb046904](https://github.com/konflux-ci/integration-service/commit/bb0469045e2e25c72538e10ef291c786a65a7802) and the adapter no longer does it — the new `EnsureIntegrationPipelineServiceAccountCreated` only provisions SA/Secret/RoleBinding based on the scenario's namespace, so the 
invalid-application test case no longer applies.

jsztuka (approved):
lgtm

PR #1604, opened 2026-06-25
feat(STONEINTG-1671): implement immediate-mode nudging orchestration

kasemAlem:
## Summary

  Implements Phase 1 of ADR-0067: migrating component nudging from build-service to integration-service.

  When a post-merge build PipelineRun succeeds, integration-service now reads the NudgeConfig singleton, finds immediate-mode edges where `from` matches the built component, and creates Renovate nudge
  PipelineRuns for each downstream target — producing the exact same PipelineRun shape that build-service creates today.

  ### What's included

  - **`EnsureNudgePipelineRunsExist`** in the build pipeline adapter — orchestrates the full nudge flow: push/success/idempotency guards → NudgeConfig lookup → edge filtering → credential
  resolution → Renovate PLR creation
  - **Renovate PipelineRun builder** (`tekton/nudge_pipeline.go`) — inline PipelineSpec with Renovate task, Secret, ConfigMap, optional CA bundle, owner references, security context — all matching
  build-service's output
  - **Credential resolution** (`tekton/nudge_credentials.go`) — basic auth via namespace-scoped SCM secrets + GitHub App path (stub, returns nil pending JWT implementation) + image registry credentials from SA
  dockercfg secrets
  - **Loader** — `GetNudgeConfig` added to `ObjectLoader` interface with mock support
  - **Constants** — ~40 build-service-compatible annotation/label names and Renovate defaults

  ### Design decisions

  - **No build-service dependency** — Renovate PipelineRun shape is replicated rather than imported, avoiding massive transitive deps
  - **No separate feature gates** — NudgeConfig existence is the only cutover signal. If no NudgeConfig exists in the namespace, nudging does not fire. The migration script controls when NudgeConfig is created per namespace.
  - **Fail-safe on permanent errors** — `ExtractNudgeBuildResult` and `GetImageRegistryCredentials` failures annotate the build PLR and stop (no infinite requeue), since these are permanent config errors that won't self-heal
  - **Idempotency** — `component-nudge-processed` annotation on the build PLR prevents duplicate nudge creation on re-reconciliation
  - **GitHub App auth** — function validates PaC secret structure but returns nil (TODO); basic auth covers the majority of real deployments today

  ### Deferred

  - GitHub App JWT + installation token flow (follow-up)
  - Distribution repository aliases from ReleasePlanAdmission
  - Metrics (`nudge_pipelinerun_created_total`, etc.)

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -7,6 +7,7 @@ rules:
 - apiGroups:
   - ""
   resources:
+  - configmaps
   - secrets
   verbs:
   - create
@@ -82,6 +83,7 @@ rules:
   - appstudio.redhat.com
   resources:
   - environments
+  - nudgeconfigs
   - releaseplans
   verbs:
   - get
--- a/internal/controller/buildpipeline/buildpipeline_adapter.go
+++ b/internal/controller/buildpipeline/buildpipeline_adapter.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"math/big"
 	"strconv"
+	"strings"
 	"time"
 
 	"k8s.io/client-go/util/retry"
@@ -39,6 +40,7 @@ import (
 	"github.com/konflux-ci/integration-service/status"
 	"github.com/konflux-ci/integration-service/tekton"
 	tektonconsts "github.com/konflux-ci/integration-service/tekton/consts"
+	"github.com/konflux-ci/integration-service/tekton/nudging"
 	"github.com/konflux-ci/operator-toolkit/controller"
 	"github.com/konflux-ci/operator-toolkit/metadata"
 	tektonv1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1"
@@ -143,6 +145,120 @@ func (a *Adapter) EnsureGlobalCandidateImageUpdated() (controller.OperationResul
 		a.logger.Error(err, "Failed to update the build pipelinerun's annotation to AddedToGlobalCandidateList")
 		return controller.RequeueWithError(err)
 	}
+
+	return controller.ContinueProcessing()
+}
+
+// EnsureNudgePipelineRunsExist creates Renovate nudge PipelineRuns for downstream
+// components defined in NudgeConfig when a push build succeeds.
+func (a *Adapter) EnsureNudgePipelineRunsExist() (controller.OperationResult, error) {
+	if !tekton.IsPLRCreatedByPACPushEvent(a.pipelineRun) {
+		return controller.ContinueProcessing()
+	}
+
+	if !h.HasPipelineRunSucceeded(a.pipelineRun) {
+		return controller.ContinueProcessing()
+	}
+
+	if metadata.HasAnnotation(a.pipelineRun, tektonconsts.NudgeProcessedAnnotation) {
+		return controller.ContinueProcessing()
+	}
+
+	nudgeConfig, err := a.loader.GetNudgeConfigForNamespace(a.context, a.client, a.pipelineRun.Namespace)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			return controller.ContinueProcessing()
+		}
+		a.logger.Error(err, "Failed to get NudgeConfig")
+		return controller.RequeueWithError(err)
+	}
+
+	componentName := a.component.Name
+	var targetNames []string
+	for _, nudge := range nudgeConfig.Spec.Nudges {
+		if nudge.From == componentName && (nudge.Mode == "" || nudge.Mode == v1beta2.NudgeModeImmediate) {
+			targetNames = append(targetNames, nudge.To)
+		}
+	}
+
+	if len(targetNames) == 0 {
+		_ = tekton.AnnotateBuildPipelineRun(a.context, a.pipelineRun, tektonconsts.NudgeProcessedAnnotation, "no-matching-edges", a.client)
+		return controller.ContinueProcessing()
+	}
+
+	var targetComponents []applicationapiv1alpha1.Component
+	for _, name := range targetNames {
+		comp, err := a.loader.GetComponent(a.context, a.client, name, a.pipelineRun.Namespace)
+		if err != nil {
+			if errors.IsNotFound(err) {
+				a.logger.Info("Nudge target component not found, skipping", "component.Name", name)
+				continue
+			}
+			a.logger.Error(err, "Failed to load nudge target component", "component.Name", name)
+			return controller.RequeueWithError(err)
+		}
+		targetComponents = append(targetComponents, *comp)
+	}
+
+	if len(targetComponents) == 0 {
+		a.logger.Info("No valid nudge target components found")
+		_ = tekton.AnnotateBuildPipelineRun(a.context, a.pipelineRun, tektonconsts.NudgeProcessedAnnotation, "no-valid-targets", a.client)
+		return controller.ContinueProcessing()
+	}
+
+	buildResult, err := nudging.ExtractBuildResultForNudging(a.pipelineRun, a.component)
+	if err != nil {
+		a.logger.Error(err, "Failed to extract build result for nudging, skipping")
+		_ = tekton.AnnotateBuildPipelineRun(a.context, a.pipelineRun, tektonconsts.NudgeProcessedAnnotation, "build-result-error", a.client)
+		return controller.ContinueProcessing()
+	}
+
+	simpleBranchName := a.component.Annotations != nil && a.component.Annotations[tektonconsts.NudgeSimpleBranchAnnotation] == "true"
+
+	saName := a.pipelineRun.Spec.TaskRunTemplate.ServiceAccountName
+	if saName == "" {
+		saName = tektonconsts.DefaultPipelineServiceAccount
+	}
+	imageRepoHost, imageRepoUser, imageRepoPwd, err := nudging.GetImageRegistryCredentials(a.context, a.client, a.component, saName)
+	if err != nil {
+		a.logger.Error(err, "Failed to get image registry credentials for nudging, skipping")
+		_ = tekton.AnnotateBuildPipelineRun(a.context, a.pipelineRun, tektonconsts.NudgeProcessedAnnotation, "credentials-error", a.client)
+		return controller.ContinueProcessing()
+	}
+
+	var targets []nudging.NudgeTarget
+
+	githubTargets := nudging.GetNudgeTargetsGithubApp(a.context, a.client, targetComponents, imageRepoHost, imageRepoUser, imageRepoPwd)
+	targets = append(targets, githubTargets...)
+
+	basicAuthTargets := nudging.GetNudgeTargetsBasicAuth(a.context, a.client, targetComponents, imageRepoHost, imageRepoUser, imageRepoPwd)
+	targets = append(targets, basicAuthTargets...)
+
+	if len(targets) == 0 {
+		a.logger.Info("No nudge targets with resolved credentials found")
+		_ = tekton.AnnotateBuildPipelineRun(a.context, a.pipelineRun, tektonconsts.NudgeProcessedAnnotation, "no-credentials", a.client)
+		return controller.ContinueProcessing()
+	}
+
+	err = nudging.CreateNudgePipelineRun(a.context, a.client, a.pipelineRun, targets, buildResult, simpleBranchName)
+	if err != nil {
+		a.logger.Error(err, "Failed to create nudge PipelineRun")
+		return controller.RequeueWithError(err)
+	}
+
+	names := make([]string, len(targets))
+	for i, t := range targets {
+		names[i] = t.ComponentName
+	}
+	processedValue := strings.Join(names, ",")
+	err = tekton.AnnotateBuildPipelineRun(a.context, a.pipelineRun, tektonconsts.NudgeProcessedAnnotation, processedValue, a.client)
+	if err != nil {
+		a.logger.Error(err, "Failed to annotate build PLR as nudge-processed")
+		return controller.RequeueWithError(err)
+	}
+
+	a.logger.LogAuditEvent("Created nudge PipelineRun for downstream components", a.pipelineRun, h.LogActionAdd,
+		"targets", processedValue)
 
 	return controller.ContinueProcessing()
 }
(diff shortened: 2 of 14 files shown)


tekton/nudge_pipeline.go
```diff
@@ -0,0 +1,661 @@
...
+	caConfigMapName := fmt.Sprintf("renovate-ca-%s", nameSuffix)
+
+	secretTokens := map[string]string{}
+	configmaps := map[string]string{}
+	var renovateCmds []string
+
+	for i, target := range targets {
+		configKey := fmt.Sprintf("%s-%d.json", target.ComponentName, i)
+		tokenKey := fmt.Sprintf("token-%d", i)
+		imageKey := fmt.Sprintf("image-%d", i)
```
jencull: Bash variable names cannot contain hyphens, causes expansion errors. Use underscores instead.

internal/controller/buildpipeline/buildpipeline_adapter.go
```diff
@@ -147,6 +148,118 @@ func (a *Adapter) EnsureGlobalCandidateImageUpdated() (controller.OperationResul
 	return controller.ContinueProcessing()
 }
 
+// EnsureNudgePipelineRunsExist creates Renovate nudge PipelineRuns for downstream
```
jencull, line 152: Will UT to cover this be in another ticket?

internal/controller/buildpipeline/buildpipeline_adapter.go
```diff
@@ -147,6 +148,118 @@ func (a *Adapter) EnsureGlobalCandidateImageUpdated() (controller.OperationResul
...
+	simpleBranchName := a.component.Annotations != nil && a.component.Annotations[tektonconsts.NudgeSimpleBranchAnnotation] == "true"
+
+	saName := a.pipelineRun.Spec.TaskRunTemplate.ServiceAccountName
+	if saName == "" {
+		saName = "appstudio-pipeline"
+	}
+	imageRepoHost, imageRepoUser, imageRepoPwd, err := tekton.GetImageRegistryCredentials(a.context, a.client, a.component, saName)
+	if err != nil {
+		a.logger.Error(err, "Failed to get image registry credentials for nudging")
+		return controller.RequeueWithError(err)
```
jencull: is there a risk of an infinite retry loop here? if, for any reason, the registry credentials can't be resolved. Possibly an edge case but its the kind of thing that happens.

jencull:
When you mention "feature gates" in the flow description, are you referring to the push/success/idempotency guards above, or is there a separate opt-in gate intended here? The description also mentions "feature gate keys" under Constants but I don't see those added — just want to make sure nothing got lost.

tekton/nudge_pipeline.go
kasemAlem: Good catch, 
fixed — switched to underscores (token_0, image_0)

internal/controller/buildpipeline/buildpipeline_adapter.go
kasemAlem: Added unit tests for EnsureNudgePipelineRunsExist in this PR — covers guard clauses (non-push, not succeeded, already processed, NudgeConfig missing), error paths (transient loader failures, credential resolution failures), edge filtering (validated-mode skip), and annotation paths (no-matching-edges, no-valid-targets)

internal/controller/buildpipeline/buildpipeline_adapter.go
kasemAlem: Good point — changed `ExtractNudgeBuildResult` and `GetImageRegistryCredentials` failures to annotate-and-stop (build-result-error / credentials-error) instead of requeueing, since these are permanent config errors that won't self-heal

kasemAlem:
> When you mention "feature gates" in the flow description, are you referring to the push/success/idempotency guards above, or is there a separate opt-in gate intended here? The description also mentions "feature gate keys" under Constants but I don't see those added — just want to make sure nothing got lost.

The "feature gates" language in the description is stale — we originally planned a global ConfigMap toggle + owner label double gate, but dropped it. NudgeConfig existence is the only cutover signal now. The "guards" in the code are just the push/success/idempotency checks. I'll update the PR description to clarify.

tekton/nudge_credentials.go
```diff
@@ -0,0 +1,622 @@
...
+// GET /app/installations to find the installation for each repo, then
+// POST /app/installations/{id}/access_tokens to get scoped tokens.
+// For now this returns nil and logs a message. Basic auth covers the majority of use cases.
+func GetNudgeTargetsGithubApp(ctx context.Context, c client.Client, targetComponents []applicationapiv1alpha1.Component, imageRepoHost, imageRepoUser, imageRepoPwd string) []NudgeTarget {
+	log := ctrllog.FromContext(ctx)
+
+	// Read the global PaC secret to check if GitHub App is configured
+	pacSecret := corev1.Secret{}
+	globalPaCSecretKey := types.NamespacedName{
+		Namespace: tektonconsts.BuildServiceNamespaceName,
```
dirgim: If I'm not mistaken, I think we'll want to use the GitHub App PaC secret from the integration service's namespace instead of depending on build-service's setup.

See a similar implementation in the [GetAppCredentials](https://github.com/konflux-ci/integration-service/blob/84b3f45d9af1ae3e2f8fe66422f43995c2c08af9/status/reporter_github.go#L85-L147) function for example of this.

tekton/nudging/nudge_pipeline.go
```diff
@@ -0,0 +1,661 @@
...
+		)
+	}
+	if len(renovateCmds) == 0 {
+		return nil
+	}
+
+	// Look for a CA bundle ConfigMap in the build-service namespace.
+	allCaConfigMaps := &corev1.ConfigMapList{}
+	if err := c.List(ctx, allCaConfigMaps,
+		client.InNamespace(tektonconsts.BuildServiceNamespaceName),
```
dirgim, line 354: note: We might need to consider migrating the config maps to the integration service's namespace. That could be done in a follow-up story/task.

kasemAlem: good point, 
@dirgim , should we add it as an AC to the migration script ? or better separate story/task ?

internal/controller/buildpipeline/buildpipeline_adapter.go
```diff
@@ -147,6 +148,120 @@ func (a *Adapter) EnsureGlobalCandidateImageUpdated() (controller.OperationResul
...
+		a.logger.Error(err, "Failed to extract build result for nudging, skipping")
+		_ = tekton.AnnotateBuildPipelineRun(a.context, a.pipelineRun, tektonconsts.NudgeProcessedAnnotation, "build-result-error", a.client)
+		return controller.ContinueProcessing()
+	}
+
+	simpleBranchName := a.component.Annotations != nil && a.component.Annotations[tektonconsts.NudgeSimpleBranchAnnotation] == "true"
+
+	saName := a.pipelineRun.Spec.TaskRunTemplate.ServiceAccountName
+	if saName == "" {
+		saName = "appstudio-pipeline"
```
dirgim: nit: we might want to extract this as a constant too.

tekton/nudge_credentials.go
kasemAlem: Good catch — switched `GetNudgeTargetsGithubApp` to read the PaC secret from the integration-service namespace using the `INTEGRATION_NS` env var (defaulting to `"integration-service"`), matching the pattern in `GetAppCredentials` from `status/reporter_github.go`. Removed the cross-namespace dependency on build-service.

internal/controller/buildpipeline/buildpipeline_adapter.go
kasemAlem: Done — added `DefaultPipelineServiceAccount` constant in `tekton/consts/consts.go` and updated both usages (adapter and `nudge_pipeline.go`).

tekton/nudging/nudge_pipeline.go
dirgim: The new separate story looks good to me, thanks!

kasemAlem: created [STONEINTG-1710](https://redhat.atlassian.net/browse/STONEINTG-1710)

dirgim (approved):
LGTM

internal/controller/buildpipeline/buildpipeline_adapter.go
```diff
@@ -147,6 +148,120 @@ func (a *Adapter) EnsureGlobalCandidateImageUpdated() (controller.OperationResul
...
+			}
+			a.logger.Error(err, "Failed to load nudge target component", "component.Name", name)
+			return controller.RequeueWithError(err)
+		}
+		targetComponents = append(targetComponents, *comp)
+	}
+
+	if len(targetComponents) == 0 {
+		a.logger.Info("No valid nudge target components found")
+		_ = tekton.AnnotateBuildPipelineRun(a.context, a.pipelineRun, tektonconsts.NudgeProcessedAnnotation, "no-valid-targets", a.client)
```
14rcole, line 205: Would it be better to set this as a status condition than annotation?

tekton/nudge_pipeline.go
```diff
@@ -0,0 +1,661 @@
...
+	AutomergeSchedule     []string             `json:"automergeSchedule,omitempty"`
+}
+
+// DisableAllPackageRules is the catch-all rule that disables all packages; the
+// subsequent enable rule overrides it for the specific built image only.
+var DisableAllPackageRules = PackageRule{MatchPackagePatterns: []string{"*"}, Enabled: false}
+
+// ExtractNudgeBuildResult reads the build pipeline results and annotations to
+// produce the data needed for nudge Renovate config generation.
+func ExtractNudgeBuildResult(plr *tektonv1.PipelineRun, component *applicationapiv1alpha1.Component) (*NudgeBuildResult, error) {
```
14rcole: Is this extracting data from a build in order to initiate nudging or is it extracting data from a build that was triggered by a nudge?

loader/loader.go
```diff
@@ -1105,6 +1106,19 @@ func (l *loader) GetAllComponentGroupsInNamespace(ctx context.Context, c client.
 	return componentGroupList.Items, err
 }
 
+// GetNudgeConfig returns the NudgeConfig singleton from the given namespace.
+func (l *loader) GetNudgeConfig(ctx context.Context, c client.Client, namespace string) (*v1beta2.NudgeConfig, error) {
```
14rcole: nit: I think this function should either take the name as a parameter (the caller can pass in v1beta2.NudgeConfigSingletonName) or be renamed to something like `GetNudgeConfigForNamespace/GetNudgeConfigSingletonForNamespace`. It makes it clearer to someone reading the code where this function is called to understand what nudge config is being gotten.

14rcole:
Can you move the new nudging files in `tekton` into its own nudging package? It could be a subpackage of tekton, I just think it's distinct enough from the other code that it should be separate.

internal/controller/buildpipeline/buildpipeline_adapter.go
kasemAlem: I would agree with your good point, it could be better choice . 
But since this is on the `BuildPipelineRun`, which is not on our ownership 
using the status condition would require CRD change and requires build team approvals 
and this can be more complex, I believe Annotation can be less complex in these situation 
@14rcole , wdyt?

tekton/nudge_pipeline.go
kasemAlem: ah, 
nice catch this name does make a confusion , 

renamed to `ExtractBuildResultForNudging` — it extracts data from the source build `PipelineRun` (`IMAGE_URL`, `IMAGE_DIGEST`, file matches) in order to generate the Renovate config that initiates nudging 
downstream. The previous name was ambiguous about the direction.

loader/loader.go
kasemAlem: Renamed to `GetNudgeConfigForNamespace` — it always fetches the singleton (`NudgeConfigSingletonName`) for the given namespace, so the new name makes that clear without needing to read the implementation.

kasemAlem:
> Can you move the new nudging files in `tekton` into its own nudging package? It could be a subpackage of tekton, I just think it's distinct enough from the other code that it should be separate.

 Good call — the nudge code is self-contained enough to warrant its own package. I'd prefer to do this as a follow-up rather than expand this already-XXL PR. The cross-package surface is small (just 
 `GetOutputImage`/`GetOutputImageDigest`), so the move should be clean. I'll create a ticket for it.
 @14rcole wdyt?

internal/controller/buildpipeline/buildpipeline_adapter.go
14rcole: That's a fair point. Let's leave this as-is

14rcole:
> Good call — the nudge code is self-contained enough to warrant its own package. I'd prefer to do this as a follow-up rather than expand this already-XXL PR. The cross-package surface is small (just `GetOutputImage`/`GetOutputImageDigest`), so the move should be clean. I'll create a ticket for it. @14rcole wdyt?

I'd rather see it done in this PR. Since those four files are entirely new anyway it shouldn't change the overall size of the PR

kasemAlem:
> > Good call — the nudge code is self-contained enough to warrant its own package. I'd prefer to do this as a follow-up rather than expand this already-XXL PR. The cross-package surface is small (just `GetOutputImage`/`GetOutputImageDigest`), so the move should be clean. I'll create a ticket for it. @14rcole wdyt?
> 
> I'd rather see it done in this PR. Since those four files are entirely new anyway it shouldn't change the overall size of the PR

after double check, your point is correct and less dependencies 
I have moved to a sub package

14rcole (approved):
Great work on this!

config/rbac/role.yaml
```diff
@@ -7,6 +7,7 @@ rules:
 - apiGroups:
   - ""
   resources:
+  - configmaps
```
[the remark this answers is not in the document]
kasemAlem: It's a valid point though — the nudge code only does Get and Create on ConfigMaps (for the Renovate config), so list and watch were likely added
  automatically by the cached client requirement. Since controller-runtime's cached client needs list+watch for any resource you Get, the current RBAC is correct and necessary

internal/controller/buildpipeline/buildpipeline_controller.go
```diff
@@ -162,6 +165,7 @@ func (r *Reconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 		adapter.EnsurePRGroupAnnotated,
 		adapter.EnsureIntegrationTestReportedToGitProvider,
 		adapter.EnsureGlobalCandidateImageUpdated,
+		adapter.EnsureNudgePipelineRunsExist,
```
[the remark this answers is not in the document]
kasemAlem: nice catch, 
Fixed

PR #1614, opened 2026-07-02
feat(STONEINTG-1677): consolidated Gitlab status reporting

jencull:
- add HTTP 422 as unrecoverable across all three forge reporters
- add ReportConsolidatedStatus to reporter.go  GitLab posts a single "X/Y passed" status when scenario count exceeds the per-pipeline job threshold; GitHub and Forgejo are no-ops
- propagate IsOptional through TestReport so optional failures do not set the consolidated status to failed
- add MaxIndividualStatuses constant (50) to control the threshold at which GitLab switches to consolidated status reporting
- unit tests for consolidated status, transition error suppression, and threshold routing

Assisted by: Cursor AI


internal/controller/statusreport/statusreport_adapter.go
```diff
@@ -554,38 +557,71 @@ func (a *Adapter) iterateIntegrationTestStatusDetailsInStatusReport(reporter sta
...
+
+			if writeErr := status.WriteSnapshotReportStatus(a.context, a.client, testedSnapshot, srs); writeErr != nil {
+				return fmt.Errorf("failed to report status AND write snapshot report status metadata: %w", writeErr)
+			}
+
+			a.logger.Info("Successfully report integration test status for snapshot",
+				"testedSnapshot.Name", testedSnapshot.Name,
+				"destinationSnapshot.Name", destinationSnapshot.Name,
+				"testStatus", integrationTestStatusDetail.Status)
+			srs.SetLastUpdateTime(integrationTestStatusDetail.ScenarioName, destinationSnapshot.Name, integrationTestStatusDetail.LastUpdateTime)
```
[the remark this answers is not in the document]
jencull: No changes in this file introduced in this PR.

internal/controller/statusreport/statusreport_adapter.go
```diff
@@ -554,38 +557,71 @@ func (a *Adapter) iterateIntegrationTestStatusDetailsInStatusReport(reporter sta
...
-			//integration test contains no changes
-			a.logger.Info("Integration Test doen't contain new status updates", "scenario.Name", integrationTestStatusDetail.ScenarioName)
-			continue
-		}
+	// When the number of scenarios exceeds the configured threshold for GitLab or
+	// Forgejo reporters, post a single consolidated commit status instead of one
+	// per scenario to avoid exhausting the provider's per-pipeline job limit.
+	maxIndividual := status.GetMaxIndividualStatuses()
+	useConsolidated := (reporter.GetReporterName() == status.GitLabProvider || reporter.GetReporterName() == status.ForgejoProvider) &&
+		len(integrationTestStatusDetails) > maxIndividual
```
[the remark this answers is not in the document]
jencull: actioned and change made.

status/reporter_gitlab.go
```diff
@@ -515,6 +510,102 @@ func (r *GitLabReporter) ReportStatus(ctx context.Context, report TestReport) (i
...
+		case intgteststat.IntegrationTestStatusTestFail,
+			intgteststat.IntegrationTestStatusTestInvalid,
+			intgteststat.IntegrationTestStatusEnvironmentProvisionError_Deprecated,
+			intgteststat.IntegrationTestStatusDeploymentError_Deprecated,
+			intgteststat.BuildPLRFailed,
+			intgteststat.SnapshotCreationFailed,
+			intgteststat.GroupSnapshotCreationFailed:
+			if !report.IsOptional {
+				overallState = gitlab.Failed
+			}
```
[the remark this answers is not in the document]
jencull: Deleted maps to Canceled in the existing per-scenario reporter, not Failed. The consolidated status intentionally treats it the same way by not counting it as a failure.

status/reporter_gitlab.go
```diff
@@ -515,6 +510,102 @@ func (r *GitLabReporter) ReportStatus(ctx context.Context, report TestReport) (i
...
+	return srcStatusCode, srcErr
+}
+
+// isGitLabRedundantTransitionError returns true when GitLab rejects a commit
+// status update because the status is already in the requested state. These
+// errors are safe to ignore — the desired state is already reflected.
+func isGitLabRedundantTransitionError(err error) bool {
+	return strings.Contains(err.Error(), "Cannot transition status via :enqueue from :pending") ||
+		strings.Contains(err.Error(), "Cannot transition status via :run from :running")
+}
```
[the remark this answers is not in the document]
jencull: actioned.

jencull:
Actions taken on Fullsend recommendations: 

[logic-error] Fixed — removed ForgejoProvider from the useConsolidated condition. Forgejo always uses the per-scenario path.

[missing-retry] Fixed — extracted setConsolidatedCommitStatus helper and wrapped ReportConsolidatedStatus with retry.OnError, consistent with ReportStatus.

[missing-case] Fixed — added explicit IntegrationTestStatusDeleted case with a comment explaining it intentionally does not affect the consolidated state (maps to Canceled in per-scenario reporting).

[edge-case] Moot — GetMaxIndividualStatuses env var replaced with const MaxIndividualStatuses = 50. Hardcoded the number, as done in GC etc. 

[variable-naming] Moot — env var removed, constant has no provider prefix. Forgejo also removed from routing.

[typo] Not actioned — pre-existing typo, not introduced by this PR. Fixing it here would add noise to the diff.

[comment-style-inconsistency] Already fixed via gemini code review message — target project path in setCommitStatus was updated to use isGitLabRedundantTransitionError during a rebase cleanup.

jencull:
there is a flaky test that fails about 40% of the time when run on local. Ticket created to address it, 1694. https://github.com/konflux-ci/integration-service/actions/runs/28600938718/job/84808668695 Same test is failing on PR 1607

status/reporter_gitlab.go
```diff
@@ -515,6 +507,132 @@ func (r *GitLabReporter) ReportStatus(ctx context.Context, report TestReport) (i
...
+			// Deleted maps to Canceled in per-scenario reporting and does not
+			// affect the overall pass/fail outcome of the consolidated status.
+		}
+	}
+
+	consoleName := getConsoleName()
+	opt := gitlab.SetCommitStatusOptions{
+		State:       gitlab.BuildStateValue(overallState),
+		Name:        gitlab.Ptr(fmt.Sprintf("%s / Integration Tests", consoleName)),
+		Description: gitlab.Ptr(fmt.Sprintf("%d/%d passed", passed, total)),
```
kasemAlem, line 555: as a UX improvement I think it would make more sense if we add a clickable way uses can move towards see what failed by linking the consolidated commit status to the snapshot overview page in the Konflux console, so users can click "650/700 passed" and see which tests failed.

  The way to do it: add a `CONSOLE_URL_SNAPSHOT `env var (same pattern as the existing `CONSOLE_URL `for PipelineRuns), with `{{NAMESPACE}}` and `{{SNAPSHOT_NAME}}` placeholders, and set `opt.TargetURL` in
  `ReportConsolidatedStatus` using it.

if it's not for this pr scope we may take in separate issue

status/reporter_gitlab.go
```diff
@@ -515,6 +507,132 @@ func (r *GitLabReporter) ReportStatus(ctx context.Context, report TestReport) (i
...
+	overallState := gitlab.Success
+
+	for _, report := range reports {
+		switch report.Status {
+		case intgteststat.IntegrationTestStatusTestPassed, intgteststat.IntegrationTestStatusTestWarning:
+			passed++
+		case intgteststat.IntegrationTestStatusInProgress:
+			if overallState == gitlab.Success {
+				overallState = gitlab.Running
+			}
```
kasemAlem, line 530: I think here we have a bug in logic : 
  The code loops through all test reports to figure out one overall GitLab status. It starts as Success and updates as it encounters different test states. The problem is with two specific states — Pending and
  Running (InProgress).

  Both have the same guard: "only update if the current state is still Success." So whichever one the loop hits first sets the state, and the other one can never override it.

  Example: You have 3 tests. Test A is pending (hasn't started), Test B is running (actively executing), Test C passed.

  - If the loop processes them in order A, B, C: A sets state to Pending. B tries to set Running but the guard says "only if Success" — it's Pending now, so B is ignored. Final state: Pending.
  - If the loop processes them in order B, A, C: B sets state to Running. A tries to set Pending but the guard blocks it. Final state: Running.

  Same data, different result — just because of iteration order.

If at least one test is actively running, the overall state should be Running. Pending means "nothing has started yet," which is wrong when work is in progress.

fix proposed : 

```
  case intgteststat.IntegrationTestStatusInProgress:
      if overallState == gitlab.Success || overallState == gitlab.Pending {
          overallState = gitlab.Running
      }

```

status/status.go
```diff
@@ -459,6 +460,12 @@ func getConsoleName() string {
 	return consoleName
 }
 
+// MaxIndividualStatuses is the maximum number of per-scenario commit statuses
+// that will be posted to GitLab before switching to a single consolidated
+// status. Above this threshold, one "X/Y passed" status is posted instead to
+// avoid exhausting GitLab's max_jobs_per_pipeline limit.
+const MaxIndividualStatuses = 50
```
kasemAlem, line 467: should we consider this be an env variable so it be controlled by operators/users without the need to build a new image ?
if this is not in this or scope we can open a new issue for it

kasemAlem:
adding the new field [isOptional](https://github.com/jencull/integration-service/blob/28047cf05876c72aa414b997d26d540cd9aab442/status/reporter.go#L64)  to the status/reporter is a good way
but it's only used in the [consolidated path](https://github.com/jencull/integration-service/blob/28047cf05876c72aa414b997d26d540cd9aab442/status/reporter_gitlab.go#L542) , meanwhile in the [scenario path](https://github.com/jencull/integration-service/blob/28047cf05876c72aa414b997d26d540cd9aab442/status/reporter_gitlab.go#L170-L176) we kept using the k8s lookup, and this way is using another redundant K8s Get, we can use the isOptional now and save this redundant API call.

status/reporter_gitlab.go
```diff
@@ -515,6 +507,132 @@ func (r *GitLabReporter) ReportStatus(ctx context.Context, report TestReport) (i
...
+		return http.StatusOK, nil
+	}
+
+	r.logger.Error(errors.Join(err, srcErr), "failed to set consolidated commit status on both projects",
+		"targetProjectID", r.targetProjectID, "sourceProjectID", r.sourceProjectID, "sha", r.sha)
+	if !r.ReturnCodeIsUnrecoverable(statusCode) {
+		return statusCode, err
+	}
+	return srcStatusCode, srcErr
+}
```
kasemAlem, line 629: When both the target project and the source project fail to set a commit status, the code needs to decide which error to return. The caller uses the returned status code to decide whether to retry or give up.

  The existing [setCommitStatus](https://github.com/konflux-ci/integration-service/blob/3e5f7eb3051c8c044f8f617156edbd906ab32ded/status/reporter_gitlab.go#L309-L317)  has a 3-step fallback:

  1. If target error is recoverable → return target error (caller retries against target)
  2. If source error is recoverable → return source error (caller retries against source)
  3. If both are unrecoverable → return target error (caller gives up)

  The new [setConsolidatedCommitStatus](https://github.com/jencull/integration-service/blob/28047cf05876c72aa414b997d26d540cd9aab442/status/reporter_gitlab.go#L619-L622) has a 2-step fallback:

  1. If target error is recoverable → return target error (same)
  2. Otherwise → return source error (regardless of whether source is recoverable or not)

  The difference shows up when both are unrecoverable: the original returns the target error, the new code returns the source error. The end result is the same — the caller gives up either way because the error
  is unrecoverable. But someone reading the two functions side by side would wonder why they behave differently in that edge case.

  It's not a bug — just an inconsistency that could confuse a future reader. The suggestion is simply to match the original 3-step pattern so both functions work the same way.

kasemAlem:
Test gaps

  1. Threshold routing in the adapter: There's no test in `statusreport_adapter_test.go` that verifies the consolidated path is actually triggered when the scenario count exceeds `MaxIndividualStatuses`. The unit
  tests cover `ReportConsolidatedStatus` in isolation, but the branching logic in the adapter — which is the most critical integration point — is untested.
  2. Mixed non-terminal states: There's no test for `ReportConsolidatedStatus` receiving a mix of `Pending` and `InProgress` reports in different iteration orders. This would catch the state priority bug (`Running`
  should override `Pending` regardless of order) and serve as a regression test for the fix.

status/reporter_gitlab.go
jencull: TY, good catch. Makes sense to do it here.

status/reporter_gitlab.go
jencull: Done. ty.

status/status.go
jencull: Not yet, potentially in the future. The follow up ticket is the observability aspect where we will make further decisions based on how this plays out for our users when in use.

jencull:
> adding the new field [isOptional](https://github.com/jencull/integration-service/blob/28047cf05876c72aa414b997d26d540cd9aab442/status/reporter.go#L64) to the status/reporter is a good way
> but it's only used in the [consolidated path](https://github.com/jencull/integration-service/blob/28047cf05876c72aa414b997d26d540cd9aab442/status/reporter_gitlab.go#L542) , meanwhile in the [scenario path](https://github.com/jencull/integration-service/blob/28047cf05876c72aa414b997d26d540cd9aab442/status/reporter_gitlab.go#L170-L176) we kept using the k8s lookup, and this way is using another redundant K8s Get, we can use the isOptional now and save this redundant API call.

@kasemAlem agreed. I will create a ticket for it in our backlog. TY

status/reporter_gitlab.go
jencull: made the change to keep it consistent. TY.

jencull:
> Test gaps
> 
> 1. Threshold routing in the adapter: There's no test in `statusreport_adapter_test.go` that verifies the consolidated path is actually triggered when the scenario count exceeds `MaxIndividualStatuses`. The unit
>    tests cover `ReportConsolidatedStatus` in isolation, but the branching logic in the adapter — which is the most critical integration point — is untested.
> 2. Mixed non-terminal states: There's no test for `ReportConsolidatedStatus` receiving a mix of `Pending` and `InProgress` reports in different iteration orders. This would catch the state priority bug (`Running`
>    should override `Pending` regardless of order) and serve as a regression test for the fix.

tests written. tyvm.

kasemAlem:
/retest

The etcd timeout + leader election loss + pod restart created a 44-minute gap during which the completed PLR was not processed. A retest should pass.

internal/controller/statusreport/statusreport_adapter.go
```diff
@@ -554,38 +557,70 @@ func (a *Adapter) iterateIntegrationTestStatusDetailsInStatusReport(reporter sta
...
-		if writeErr := status.WriteSnapshotReportStatus(a.context, a.client, testedSnapshot, srs); writeErr != nil { // try to write what was already written
-			return fmt.Errorf("failed to report status AND write snapshot report status metadata: %w", writeErr)
+		// Mark all scenarios as reported so they are not re-reported on the next reconcile
+		for _, detail := range integrationTestStatusDetails {
+			srs.SetLastUpdateTime(detail.ScenarioName, destinationSnapshot.Name, detail.LastUpdateTime)
+		}
+		if writeErr := status.WriteSnapshotReportStatus(a.context, a.client, testedSnapshot, srs); writeErr != nil {
+			return fmt.Errorf("failed to write snapshot report status after consolidated status: %w", writeErr)
 		}
+	} else {
```
dirgim, line 579: suggestion: This additional conditional layer (`if useConsolidated {...} else {...}`)  increases the code complexity by quite a bit. Maybe extracting this branch into a new function could help here?

internal/controller/statusreport/statusreport_adapter.go
```diff
@@ -525,23 +525,26 @@ func (a *Adapter) iterateIntegrationTestStatusDetailsInStatusReport(reporter sta
...
+			if writeErr := status.WriteSnapshotReportStatus(a.context, a.client, testedSnapshot, srs); writeErr != nil {
 				return fmt.Errorf("failed to generate test report AND write snapshot report status metadata: %w", e.Join(reportErr, writeErr))
 			}
 			return fmt.Errorf("failed to generate test report: %w", reportErr)
 		}
+		allReports = append(allReports, *testReport)
+
 		// split passed and failing integration test report in gitlab comment to show failing tests on top
 		if integrationTestStatusDetail.Status == intgteststat.IntegrationTestStatusTestPassed {
-			// generate comment for passed integration test with short text which has pipelinerun link but without task run details
```
dirgim, line 544: nit: probably best to leave this comment here just in case.

internal/controller/statusreport/statusreport_adapter.go
```diff
@@ -525,23 +525,26 @@ func (a *Adapter) iterateIntegrationTestStatusDetailsInStatusReport(reporter sta
 		return nil
 	}
 
+	// Build the per-scenario test reports. These are always needed — for comment
+	// text in both paths and for per-scenario commit statuses in the individual path.
+	var allReports []status.TestReport
 	for _, integrationTestStatusDetail := range integrationTestStatusDetails {
-		// set isFinalStatus to true if there is at least one integration test status is in final status, which means the comment for integration test might not be updated again to have only one comment for each component
```
dirgim, line 529: nit: probably best to leave this comment here just in case.

internal/controller/statusreport/statusreport_adapter.go
jencull: done, slightly edited version included

internal/controller/statusreport/statusreport_adapter.go
jencull: done

internal/controller/statusreport/statusreport_adapter.go
jencull: done, ty :)

dirgim (approved):
LGTM

kasemAlem (approved):
great work , thx @jencull

PR #1646, opened 2026-07-12
feat(STONEINTG-1682): add one-time migration script for build-nudges-ef to NudgeConfig

kasemAlem:
…

  Implement a standalone binary that reads Component.spec.build-nudges-ref
  data across tenant namespaces and creates or merges NudgeConfig CRDs as
  part of ADR-0067 Phase 2 migration from build-service to integration-service
  owned nudging.

The script supports:
- Automatic tenant namespace discovery via label selectors
- Targeted namespace migration via positional arguments
- Dry-run mode (--dry-run) for safe pre-migration validation
- Merge-not-replace semantics preserving user-added NudgeConfig entries
- Filtering of dangling references, self-nudges, and duplicates
- Client-side DAG cycle detection before API writes
- Idempotent execution (second run is a no-op)
- Migration metadata via labels and annotations for traceability

  Test Plan

  - 25 unit tests covering merge logic, filtering, cycle detection, dry-run, and idempotency
  - Manual dry-run validated against stage and production clusters

  Assisted-By: Claude (Anthropic)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,7 +26,8 @@ RUN if [ "$ENABLE_COVERAGE" = "true" ]; then \
         echo "Building production binary..."; \
         CGO_ENABLED=0 go build -a -o manager ./cmd; \
     fi \
- && CGO_ENABLED=0 go build -a -o snapshotgc cmd/snapshotgc/snapshotgc.go
+ && CGO_ENABLED=0 go build -a -o snapshotgc cmd/snapshotgc/snapshotgc.go \
+ && CGO_ENABLED=0 go build -a -o nudge-migrate cmd/nudge-migrate/nudge_migrate.go
 
 ARG ENABLE_WEBHOOKS=true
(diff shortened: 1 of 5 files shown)


dirgim:
Hi @kasemAlem, since this is meant to be a single-use (or a couple of uses) script, I don't think it should be added to the controller's image.
Wdyt about having it in the https://github.com/konflux-ci/integration-service-utils repository instead?

kasemAlem:
> Hi @kasemAlem, since this is meant to be a single-use (or a couple of uses) script, I don't think it should be added to the controller's image. Wdyt about having it in the https://github.com/konflux-ci/integration-service-utils repository instead?

you are right, 
I'll create on on utils then will close this one , 
thanks @dirgim

kasemAlem:
closing this one , a new one is created https://github.com/konflux-ci/integration-service-utils/pull/13

PR #1649, opened 2026-07-16
feat(STONEINTG-1681): detect stale nudge references

dirgim:
* Add stale reference handling - detect orphaned entries in NudgeConfig after Component deletion
* Consolildate the logic for this so it can be used both in webhooks and controllers

Assisted-by: Cursor

### DIFF
--- a/helpers/nudges.go
+++ b/helpers/nudges.go
@@ -1,0 +1,73 @@
+/*
+Copyright 2026 Red Hat Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package helpers
+
+import (
+	"fmt"
+	"strings"
+
+	applicationapiv1alpha1 "github.com/konflux-ci/application-api/api/v1alpha1"
+	"github.com/konflux-ci/integration-service/api/v1beta2"
+)
+
+const (
+	// StaleReferencesStatusCondition is the status condition indicating if a given NudgeConfig has stale Component references
+	StaleReferencesStatusCondition = "StaleReferences"
+	// StaleReferencesDetectedReason is the status condition reason indicating that stale references were detected in the NudgeConfig
+	StaleReferencesDetectedReason = "StaleReferencesDetected"
+	// NoStaleReferencesReason is the status condition reason indicating that no stale references were detected in the NudgeConfig
+	NoStaleReferencesReason = "NoStaleReferences"
+)
+
+// FindMissingNudgeConfigReferences returns a boolean indicating any missing nudgeConfig references and an accompanying message
+// for a given list of components and nudge relationships
+func FindMissingNudgeConfigReferences(components []applicationapiv1alpha1.Component, nudges []v1beta2.NudgeRelationship, namespace string) (bool, string) {
+	existing := make(map[string]struct{}, len(components))
+	for i := range components {
+		existing[components[i].Name] = struct{}{}
+	}
+
+	var missingFrom, missingTo []string
+	seenFrom, seenTo := map[string]struct{}{}, map[string]struct{}{}
+	for _, n := range nudges {
+		if _, ok := existing[n.From]; !ok {
+			if _, dup := seenFrom[n.From]; !dup {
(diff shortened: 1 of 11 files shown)


loader/loader.go
```diff
@@ -130,6 +131,21 @@ func (l *loader) GetAllApplicationComponents(ctx context.Context, c client.Clien
...
+func (l *loader) GetAllComponentsInNamespace(ctx context.Context, c client.Client, namespace string) (*[]applicationapiv1alpha1.Component, error) {
+	namespaceComponents := &applicationapiv1alpha1.ComponentList{}
+	opts := []client.ListOption{
+		client.InNamespace(namespace),
+	}
+
+	err := c.List(ctx, namespaceComponents, opts...)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
dirgim: The context of the call site gets added in calling functions, the loader functions are not meant to have this context so they do not wrap the errors for their calls.

jsztuka (approved):
lgtm

internal/controller/buildpipeline/buildpipeline_adapter.go
```diff
@@ -1605,3 +1612,44 @@ func (a *Adapter) emitBuildTimingSpans() {
...
+		meta.SetStatusCondition(&nudgeConfig.Status.Conditions, metav1.Condition{
+			Type:    h.StaleReferencesStatusCondition,
+			Status:  newConditionStatus,
+			Reason:  newConditionReason,
+			Message: msg,
+		})
+
+		err = a.client.Status().Patch(a.context, nudgeConfig, patch)
+		if err != nil {
+			return fmt.Errorf("failed to update nudge config: %w", err)
```
kasemAlem:
```suggestion
			return fmt.Errorf("failed to patch nudge config status: %w", err)
```

helpers/nudges_test.go
```diff
@@ -0,0 +1,138 @@
...
+
+func newComponent(name string) applicationapiv1alpha1.Component {
+	return applicationapiv1alpha1.Component{
+		ObjectMeta: metav1.ObjectMeta{Name: name},
+	}
+}
+
+var _ = Describe("Helpers for nudges", func() {
+
+	Context("FindMissingNudgesConfigReferences", func() {
```
kasemAlem:
```suggestion
	Context("FindMissingNudgeConfigReferences", func() {
```

internal/controller/buildpipeline/buildpipeline_adapter.go
```diff
@@ -1605,3 +1612,44 @@ func (a *Adapter) emitBuildTimingSpans() {
 		a.pipelineRun = patched
 	}
 }
+
+func (a *Adapter) checkNudgeConfigForStaleReferences(nudgeConfig *v1beta2.NudgeConfig, namespace string) error {
+	nudges := nudgeConfig.Spec.Nudges
+	existingStatusCondition := meta.FindStatusCondition(nudgeConfig.Status.Conditions, h.StaleReferencesStatusCondition)
+	if len(nudges) == 0 && (existingStatusCondition == nil || existingStatusCondition.Status == metav1.ConditionFalse) {
```
kasemAlem:
```suggestion
  // Short-circuit only when there is genuinely nothing to do: no nudges to check
  // and no prior True condition to clean up. We intentionally fall through when
  // existingStatusCondition is nil so that a fresh NudgeConfig gets an explicit
  // StaleReferences=False condition rather than leaving the field absent.
  // Absent is ambiguous (unknown vs healthy); False is an affirmative health signal.
  if len(nudges) == 0 && existingStatusCondition != nil && existingStatusCondition.Status == metav1.ConditionFalse {
      return nil
  }
```

  The only logic change is flipping existingStatusCondition == nil || existingStatusCondition.Status == metav1.ConditionFalse to existingStatusCondition != nil && existingStatusCondition.Status == metav1.ConditionFalse. A fresh object (nil condition) now falls through and gets the
  condition set to False. An object that already has False still short-circuits. An object with True still falls through (unchanged behavior).

kasemAlem:
great work @dirgim , thank you !
added a couple of minor comments , beside approved

internal/controller/buildpipeline/buildpipeline_adapter.go
dirgim: Ack, updated.

helpers/nudges_test.go
dirgim: Ack, updated.

internal/controller/buildpipeline/buildpipeline_adapter.go
dirgim: Good catch, updated.

kasemAlem (approved):

helpers/nudges.go
```diff
@@ -0,0 +1,73 @@
+/*
```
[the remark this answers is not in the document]
dirgim: The header is there.

PR #1650, opened 2026-07-16
feat(stoneintg-1692): support basic e2e tests for ComponentGroup

14rcole:
Duplicates the functionality of the integration tests in integration.go to support componentGroups. Does so in a way that allows easy removal of application-speficic code when that code is deprecated. Also updates several helper functions to have the same dual support

### DIFF
--- a/e2e-tests/pkg/clients/has/componentgroups.go
+++ b/e2e-tests/pkg/clients/has/componentgroups.go
@@ -1,0 +1,104 @@
+package has
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/konflux-ci/integration-service/api/v1beta2"
+	"github.com/konflux-ci/integration-service/e2e-tests/pkg/utils"
+	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	rclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// GetComponentGroup returns a component group given a name and namespace from kubernetes cluster.
+func (h *HasController) GetComponentGroup(name string, namespace string) (*v1beta2.ComponentGroup, error) {
+	componentGroup := v1beta2.ComponentGroup{
+		Spec: v1beta2.ComponentGroupSpec{},
+	}
+	if err := h.KubeRest().Get(context.Background(), types.NamespacedName{Name: name, Namespace: namespace}, &componentGroup); err != nil {
+		return nil, err
+	}
+
+	return &componentGroup, nil
+}
+
+// CreateComponentGroup creates a ComponentGroup object with a default timeout of
+// 10 minutes. The timeout covers only the API write; controller reconciliation
+// (status.globalCandidateList population) happens asynchronously and must be waited on separately.
+func (h *HasController) CreateComponentGroup(name string, namespace string, components []v1beta2.ComponentReference) (*v1beta2.ComponentGroup, error) {
+	return h.CreateComponentGroupWithTimeout(name, namespace, components, time.Minute*10)
+}
+
+// CreateComponentGroupWithTimeout creates a component group in the kubernetes cluster with a custom default time for creation.
+func (h *HasController) CreateComponentGroupWithTimeout(name string, namespace string, components []v1beta2.ComponentReference, timeout time.Duration) (*v1beta2.ComponentGroup, error) {
+	componentGroup := &v1beta2.ComponentGroup{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+		},
+		Spec: v1beta2.ComponentGroupSpec{
+			Components: components,
+		},
+	}
+
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
+	defer cancel()
+	if err := h.KubeRest().Create(ctx, componentGroup); err != nil {
+		return nil, err
+	}
+
+	return componentGroup, nil
+}
+
+// DeleteComponentGroup delete a ComponentGroup resource from the namespace.
+// Optionally, it can avoid returning an error if the resource did not exist:
+// - specify 'false', if it's likely the ComponentGroup has already been deleted (for example, because the Namespace was deleted)
+func (h *HasController) DeleteComponentGroup(name string, namespace string, reportErrorOnNotFound bool) error {
+	componentGroup := v1beta2.ComponentGroup{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      name,
+			Namespace: namespace,
+		},
+	}
+	if err := h.KubeRest().Delete(context.Background(), &componentGroup); err != nil {
+		if !k8sErrors.IsNotFound(err) || (k8sErrors.IsNotFound(err) && reportErrorOnNotFound) {
+			return fmt.Errorf("error deleting a component group: %+v", err)
+		}
+	}
+	return utils.WaitUntil(h.ComponentGroupDeleted(&componentGroup), 1*time.Minute)
+}
+
+// ComponentGroupDeleted check if a given componentGroup object was deleted successfully from the kubernetes cluster.
+func (h *HasController) ComponentGroupDeleted(componentGroup *v1beta2.ComponentGroup) wait.ConditionFunc {
+	return func() (bool, error) {
+		_, err := h.GetComponentGroup(componentGroup.Name, componentGroup.Namespace)
+		return err != nil && k8sErrors.IsNotFound(err), nil
+	}
+}
+
+// DeleteAllComponentGroupsInASpecificNamespace removes all componentGroup CRs from a specific namespace. Useful when creating a lot of resources and want to remove all of them
+func (h *HasController) DeleteAllComponentGroupsInASpecificNamespace(namespace string, timeout time.Duration) error {
+	if err := h.KubeRest().DeleteAllOf(context.Background(), &v1beta2.ComponentGroup{}, rclient.InNamespace(namespace)); err != nil {
+		return fmt.Errorf("error deleting componentGroups from the namespace %s: %+v", namespace, err)
+	}
+
+	return utils.WaitUntil(func() (done bool, err error) {
+		componentGroupList, err := h.ListAllComponentGroups(namespace)
+		if err != nil {
+			return false, nil
+		}
+		return len(componentGroupList.Items) == 0, nil
+	}, timeout)
+}
+
+// ListAllComponentGroups returns a list of all ComponentGroups in a given namespace.
+func (h *HasController) ListAllComponentGroups(namespace string) (*v1beta2.ComponentGroupList, error) {
+	componentGroupList := &v1beta2.ComponentGroupList{}
+	err := h.KubeRest().List(context.Background(), componentGroupList, &rclient.ListOptions{Namespace: namespace})
+
+	return componentGroupList, err
+}
--- a/e2e-tests/pkg/clients/has/components.go
+++ b/e2e-tests/pkg/clients/has/components.go
@@ -59,13 +59,13 @@ func (h *HasController) GetComponentByApplicationName(applicationName string, na
 }
 
 // GetComponentPipeline returns first pipeline run for a given component labels
-func (h *HasController) GetComponentPipelineRun(componentName, applicationName, namespace, sha string) (*pipeline.PipelineRun, error) {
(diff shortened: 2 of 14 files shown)


e2e-tests/tests/integration-service/helpers.go
```diff
@@ -0,0 +1,114 @@
...
+	"github.com/konflux-ci/integration-service/e2e-tests/pkg/framework"
+	"github.com/konflux-ci/integration-service/e2e-tests/pkg/utils"
+	"github.com/konflux-ci/integration-service/e2e-tests/pkg/utils/build"
+	ginkgo "github.com/onsi/ginkgo/v2"
+	gomega "github.com/onsi/gomega"
+)
+
+func cleanup(f framework.Framework, testNamespace, parentName, componentName string, snapshot *appstudioApi.Snapshot, usingComponentGroups bool) {
+	if !ginkgo.CurrentSpecReport().Failed() {
+		gomega.Expect(f.AsKubeAdmin.IntegrationController.DeleteSnapshot(snapshot, testNamespace)).To(gomega.Succeed())
```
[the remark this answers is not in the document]
14rcole: We can ignore this, snapshot will never be nil

dirgim (approved):
LGTM

jsztuka (approved):
lgtm

e2e-tests/pkg/clients/has/components.go
```diff
@@ -75,8 +75,8 @@ func (h *HasController) GetComponentPipelineRunWithType(componentName string, ap
 }
 
 // GetComponentPipelineRunsWithType returns all pipeline runs for a given component labels with pipeline type within label "pipelines.appstudio.openshift.io/type" ("build", "test")
-func (h *HasController) GetComponentPipelineRunsWithType(componentName string, applicationName string, namespace, pipelineType string, sha string, eventType string) (*[]pipeline.PipelineRun, error) {
-	pipelineRunLabels := map[string]string{tektonconsts.PipelineRunComponentLabel: componentName, tektonconsts.PipelineRunApplicationLabel: applicationName}
+func (h *HasController) GetComponentPipelineRunsWithType(componentName string, namespace, pipelineType string, sha string, eventType string) (*[]pipeline.PipelineRun, error) {
```
kasemAlem, line 78: Removing `applicationName` from the signature and `PipelineRunApplicationLabel` from the label map is correct for `ComponentGroup` callers — `ComponentGroup` pipeline runs aren't labeled by application. But Application-mode callers in `integration-application.go` now filter by component name only. In a namespace where two applications each own a component with the same name, this can return pipeline runs that belong to the wrong application, causing flaky or incorrect test assertions.

 The pattern already established in this PR by `CreateComponent (setting the field conditionally when applicationName != "")` applies here too:

```
// TODO: remove when we remove support for applications  
 func (h *HasController) GetComponentPipelineRunsWithType(
    componentName string, applicationName string, namespace, pipelineType string, sha string, eventType string,
  ) (*[]pipeline.PipelineRun, error) {
      pipelineRunLabels := map[string]string{
         tektonconsts.PipelineRunComponentLabel: componentName,
     }
     if applicationName != "" {
         pipelineRunLabels[tektonconsts.PipelineRunApplicationLabel] = applicationName
     }
     // ... rest unchanged
  }
```
 
Then update the callers in integration-application.go to pass the application name as before, and the callers in `integration.go` (`ComponentGroup` path) to pass "".
 
This also restores the `applicationName` parameter to `GetComponentPipelineRun` and `GetComponentPipelineRunWithType` so they can thread the value through.

gitops/snapshot.go
```diff
@@ -1343,7 +1343,8 @@ func SetOwnerReference(ctx context.Context, adapterClient client.Client, snapsho
 // IsContextValidForSnapshot checks the context and compares it against the Snapshot to determine if it applies
 func IsContextValidForSnapshot(scenarioContextName string, snapshot *applicationapiv1alpha1.Snapshot) bool {
 	// `application` context is supported for backwards-compatibility and considered the same as `all`
-	if scenarioContextName == "application" || scenarioContextName == "all" {
+	// TODO: remove "application" context after migration to componentGroups (ITS's with this context should be switched to "componentGroup")
+	if scenarioContextName == "application" || scenarioContextName == "componentgroup" || scenarioContextName == "all" {
```
kasemAlem, line 1347: Missing test coverage for the new "`componentgroup`" context
 
This adds `componentgroup`as a valid context in production service code, but `gitops/snapshot_test.go` has no scenario covering it. 
The existing `FilterIntegrationTestScenariosWithContext` test exercises 
 `application`, `group`, `push`, `override`, etc. — but not `componentgroup`. A regression here would silently cause integration tests to be skipped for all `ComponentGroup` snapshots.
 
 The fix is straightforward — add one scenario variant to the existing It block and bump the expected counts by 1 everywhere (since `componentgroup` behaves identically to `application`, matching all snapshot types)


```
// add alongside the other scenario variants
   componentGroupScenario := integrationTestScenario.DeepCopy()
   componentGroupScenario.Spec.Contexts = []v1beta2.TestContext{
       {Name: "componentgroup", Description: "ComponentGroup Testing"},
   }
  
   allScenarios := []v1beta2.IntegrationTestScenario{
       *applicationScenario,
       *componentGroupScenario,   // <-- new
       *componentScenario,
       *componentSampleScenario,
       *componentSample2Scenario,
       *pullRequestScenario,
       *pushScenario,
       *groupScenario,
       *overrideScenario,
       *componentAndGroupScenario,
       *unsupportedScenario,
   }
  
   Then update the expected counts:
  
   // push component snapshot
   Expect(*filteredScenarios).To(HaveLen(7)) // was 6
  
   // pull request component snapshot
   Expect(*filteredScenarios).To(HaveLen(7)) // was 6
  
   // group snapshot
   Expect(*filteredScenarios).To(HaveLen(6)) // was 5
  
   // override snapshot
   Expect(*filteredScenarios).To(HaveLen(4)) // was 3
```

kasemAlem:
Root cause of `context deadline exceeded:` component has no application, so PaC never triggers a build
  
 The `createComponent` helper calls `CreateComponentCheckImageRepository` with `applicationName=""`, which (correctly, by this PR's design) leaves `component.Spec.Application` unset. However, the Pipelines as Code
   controller in the e2e cluster still gates build PipelineRun creation on `component.Spec.Application` being set — with it empty, PaC never configures a webhook and no build is triggered. `GetBuildPipelineRun` then
   polls until context deadline exceeded.
  
   This is a pre-condition problem, not a test logic bug: the `ComponentGroup` build path requires the build system to set `PipelineRunComponentVersionAnnotation` on `PipelineRuns` (tracked in **STONEBLD-3604**), but that
   change isn't deployed in the e2e cluster yet. Until it is, no build PipelineRun will ever appear for a component without `Spec.Application`, so this test cannot pass in the current environment.
 
  Options:
   1. Block merge until `STONEBLD-3604` is deployed in the e2e cluster. This is the cleanest path — the test is correct, the environment isn't ready.
   2. Temporarily wire the test component to a dummy Application (create one in BeforeAll, pass its name to `createComponent`), so PaC triggers a build. The integration service will fall back to the Application
   path, which defeats the purpose of the test.
   3. Mark the test with a skip condition (ginkgo.Skip("requires STONEBLD-3604 in cluster")) until the build system is ready, and merge the infrastructure changes now.
  
   Option 1 or 3 is recommended. The repeated /retest without a fix will keep failing for the same reason.

kasemAlem:
what is the reason of renaming the `const.go` to `const-application.go` ?

e2e-tests/pkg/clients/integration/pipelineruns.go
```diff
@@ -32,11 +32,16 @@ var (
 
 // GetComponentPipeline returns the pipeline for a given component labels.
 // In case of failure, this function retries till it gets timed out.
-func (i *IntegrationController) GetBuildPipelineRun(componentName, applicationName, namespace string, pacBuild bool, sha string) (*tektonv1.PipelineRun, error) {
+// TODO: when we remove the old application-specific code, remove the usingComponentGroups, applicationName, and pacBuild parameters
```
kasemAlem: Inconsistent parameter naming: `applicationName` should be `parentName` throughout this file
  
   Every other helper updated in this PR that gained a `usingComponentGroups` bool parameter was also renamed from `applicationName` to `parentName` — see `CreateIntegrationTestScenario`, `GetIntegrationTestScenarios`, 
   `CreateSnapshotWithComponents`, `CreateReleasePlan`. This file missed that rename across all 6 functions:
  
   - GetBuildPipelineRun(componentName, applicationName, ...)
   - WaitForAllIntegrationPipelinesToBeFinished(testNamespace, applicationName, ...)
   - WaitForFinalizerToGetRemovedFromAllIntegrationPipelineRuns(testNamespace, applicationName, ...)
   - GetAnnotationIfExists(testNamespace, applicationName, componentName, ...)
   - WaitForBuildPipelineRunToGetAnnotated(testNamespace, applicationName, componentName, ...)
   - WaitForBuildPipelineToBeFinished(testNamespace, applicationName, componentName, ...)
  
   When `usingComponentGroups=true`, the caller passes a `ComponentGroup` name into a parameter called `applicationName`. That mismatch is confusing on its own, but it's especially sharp next to the 
   `usingComponentGroups` bool sitting right beside it in the same signature — a reader can see both "this is a component group name" and "this parameter is called `applicationName`" in one glance.
  
   Rename all six to `parentName` to match the rest of the PR. Each has a TODO comment already in place, so the cleanup path is clear.

e2e-tests/tests/integration-service/helpers.go
kasemAlem, line 26: snapshot can be nil here — the current CI failure proves it
  
   The author replied to the bot saying "snapshot will never be nil," but the existing CI failure shows exactly the scenario where it is: the test times out waiting for a build PipelineRun, which means snapshot 
   is never assigned. When AfterAll then calls cleanup, snapshot is still its zero value (nil), and DeleteSnapshot(nil, ...) panics — masking the original timeout error with a nil dereference instead.
  
   The fix is one line:
  
```
   func cleanup(f framework.Framework, testNamespace, parentName, componentName string, snapshot *appstudioApi.Snapshot, usingComponentGroups bool) {
       if !ginkgo.CurrentSpecReport().Failed() {
           if snapshot != nil {
               gomega.Expect(f.AsKubeAdmin.IntegrationController.DeleteSnapshot(snapshot, testNamespace)).To(gomega.Succeed())
           }
           // rest unchanged
       }
   }
```
  
   This is especially important to have before merging since the test is currently timing out in CI — without this guard every failed run will produce a misleading panic on top of the real error.

e2e-tests/pkg/clients/has/componentgroups.go
```diff
@@ -0,0 +1,102 @@
...
+	}
+	if err := h.KubeRest().Get(context.Background(), types.NamespacedName{Name: name, Namespace: namespace}, &componentGroup); err != nil {
+		return nil, err
+	}
+
+	return &componentGroup, nil
+}
+
+// CreateComponentGroup creates a componentGroup in the kubernetes cluster with 10 minutes default time for creation.
+func (h *HasController) CreateComponentGroup(name string, namespace string, components []v1beta2.ComponentReference) (*v1beta2.ComponentGroup, error) {
```
kasemAlem, line 32: Doc comment overstates what the timeout controls
  
   "10 minutes default timeout" implies the function waits up to 10 minutes for the `ComponentGroup` to be ready. It doesn't — it passes the timeout to `context.WithTimeout` wrapping only the Kubernetes Create API 
   call, which returns as soon as the object is written to etcd (typically milliseconds). The controller reconciliation that populates `status.globalCandidateList` happens asynchronously after Create returns.
  
   Any test that calls CreateComponentGroup and then immediately reads `status.globalCandidateList` will find it empty regardless of this timeout. The timeout only guards against a hung API server.
  
   Suggested doc fix:
  
```
   // CreateComponentGroup creates a ComponentGroup object. The timeout covers only
   // the API write; controller reconciliation (status.globalCandidateList population)
   // happens asynchronously and must be waited on separately.
   func (h *HasController) CreateComponentGroup(...) (*v1beta2.ComponentGroup, error) {
       return h.CreateComponentGroupWithTimeout(name, namespace, components, time.Minute*10)
   }
  
```
   Same clarification applies to `CreateComponentGroupWithTimeout's` doc comment.

gitops/snapshot.go
14rcole: We don't have unit tests for any other contexts so I don't think we really need to add one for this one. We could open up a follow-up ticket to add unit tests for all of the contexts

e2e-tests/pkg/clients/has/components.go
14rcole: @kasemAlem Practically speaking, does this matter? The Components and Applications in the e2e tests are all given a random suffix so there will never be a collision. Can we safely ignore this warning? Doing so would decrease the code changes required to this PR and the code changes required to the future PR that removes the application-specific code.

PR #1652, opened 2026-07-20
fix(STONEINTG-1695): updatedComponentIs checks component label

ainephelan365:
## Summary
Fixes a bug in the `updatedComponentIs()` custom CEL function used in ReleasePlan auto-release annotation expressions.

## Root Cause
The function was iterating over `spec.components[]` to check if a component with the given name existed in the snapshot. 

Since every snapshot contains all components in the application (the newly built component plus all others from the Global Candidate List), the function returned` true` for any component name, regardless of which component actually triggered the build.

## Fix
Updated `updatedComponentIs()` to read the `appstudio.openshift.io/component` label on the snapshot instead. This label is set at snapshot creation time and identifies the specific component that triggered the build, making it the correct source of truth.

*Also added a comment noting the function does not apply to group snapshots, as they are always pull request snapshots where auto-release is not applicable.*

## Tests
Updated existing tests to set the `appstudio.openshift.io/component` label on snapshot copies. Added a new `Context` block with three test cases covering:
* Label matches requested component $\rightarrow$ `true`
* Label points to a different component (even though both exist in `spec.components`) $\rightarrow$ `false`
* Component label absent but other labels present $\rightarrow$ `false`

## Links
* **Jira:** https://redhat.atlassian.net/jira/software/c/projects/BMPTEMP/boards/9424?selectedIssue=STONEINTG-1695.
*https://onsi.github.io/ginkgo/

---

### Maintainers will complete the following section

- [ ] Commit messages are descriptive enough (hints)
- [ ] Code coverage from testing does not decrease and new code is covered (check the PR coverage on codecov)
- [ ] Controllers diagrams are updated when PR changes controllers code (if applicable)

### DIFF
--- a/gitops/release.go
+++ b/gitops/release.go
@@ -123,29 +123,23 @@ func (sf snapshotCELFunctions) shouldReleaseFn(args ...ref.Val) ref.Val {
 }
 
 func (sf snapshotCELFunctions) updatedComponentIs(arg ref.Val) ref.Val {
-	name, ok := arg.Value().(string)
+	// Note: does not work for group snapshots, they are always PR snapshots so auto-release does not apply.
+	name := arg.Value().(string)
+
+	// Check the snapshot.metadata.labels for updated component
+	meta, ok := sf.snapshot["metadata"].(map[string]any)
 	if !ok {
 		return types.Bool(false)
 	}
-	// Navigate snapshot.spec.components[].name
-	spec, ok := sf.snapshot["spec"].(map[string]any)
+	labels, ok := meta["labels"].(map[string]any)
 	if !ok {
 		return types.Bool(false)
 	}
-	components, ok := spec["components"].([]any)
+	compName, ok := labels[SnapshotComponentLabel].(string)
 	if !ok {
 		return types.Bool(false)
 	}
-	for _, c := range components {
-		cm, ok := c.(map[string]any)
-		if !ok {
-			continue
-		}
-		if compName, ok := cm["name"].(string); ok && compName == name {
-			return types.Bool(true)
-		}
-	}
-	return types.Bool(false)
+	return types.Bool(compName == name)
 }
 
 // convertToCELObjectMap marshals the typed object to JSON and back to a
--- a/gitops/release_test.go
+++ b/gitops/release_test.go
@@ -105,10 +105,57 @@ var _ = Describe("Auto-release annotation evaluation", Ordered, func() {
 
 	It("returns true when the updated component is 'component-sample'", func() {
 		snapshotCopy := hasSnapshot.DeepCopy()
+		if snapshotCopy.Labels == nil {
+			snapshotCopy.Labels = make(map[string]string)
+		}
+		snapshotCopy.Labels["appstudio.openshift.io/component"] = "component-sample" //updating label to trigger the CEL expression
 		autoReleaseAnnotation := "updatedComponentIs('component-sample')"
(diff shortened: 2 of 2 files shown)


dirgim (approved):
LGTM

gitops/release_test.go
```diff
@@ -105,12 +105,59 @@ var _ = Describe("Auto-release annotation evaluation", Ordered, func() {
...
 		allowed, err := gitops.EvaluateSnapshotAutoReleaseAnnotation(autoReleaseAnnotation, snapshotCopy, true)
 		Expect(err).NotTo(HaveOccurred())
 		Expect(allowed).To(BeTrue())
 	})
 
+	// Verifies that updatedComponentIs() uses the appstudio.openshift.io/component label
+	// to identify the triggering component, not spec.components which contains all components.
+	Context("updatedComponentIs() with multi-component snapshots", func() {
+
+		It("returns true when the label matches the requested component in the snapshot", func() {
```
kasemAlem, line 122: This test sets the label to `frontend` and appends `backend` to spec.components, but `frontend` is never added to `spec.components.` The test passes and the intent is correct, but the scenario is slightly weaker than it could be — it doesn't prove that the function ignores a 
  component that is present in `spec.components` when it's not the label value.
  
  Consider adding `frontend` to spec.components as well. That way the test clearly demonstrates: even though frontend exists in the component list, the function correctly returns false for `updatedComponentIs('backend') `because the label is the only source of truth.

gitops/release.go
```diff
@@ -123,29 +123,23 @@ func (sf snapshotCELFunctions) shouldReleaseFn(args ...ref.Val) ref.Val {
 }
 
 func (sf snapshotCELFunctions) updatedComponentIs(arg ref.Val) ref.Val {
-	name, ok := arg.Value().(string)
+	// Note: does not work for group snapshots, they are always PR snapshots so auto-release does not apply.
+	name := arg.Value().(string)
```
kasemAlem, line 127: This is an unsafe type assertion. It tries the same conversion but throws away the ok check. If `arg.Value()` is not a string, Go does not return false — it panics (crashes the program).

```suggestion
     name, ok := arg.Value().(string)
  if !ok {
      return types.Bool(false)
  }
```

  This is a safe type assertion in Go. It tries to convert `arg.Value()` to a string, and puts the result into two variables:
  - name — the string value if it succeeded
  - ok — a boolean: true if the conversion worked, false if it didn't

  If it fails (e.g. someone passed a number instead of a string), `ok` is false, and the function just returns false safely — no crash.

kasemAlem:
Great work on this fix, @ainephelan365 ! The core logic change is spot on — switching from iterating over spec.components to reading the metadata label is exactly the right approach, and the new test cases do a solid job of covering the key scenarios, especially the case where a component is present in the list but wasn't the one that triggered the build.

a nit comment:  It would be more useful if you add more information to the pr description that would safe time to you and other users: 

  - What is the issue this pr is fixing, in short summary 
  - Root cause 
  - how it's being fixed 
  - useful links, jira ..etc 

Overall this is a clean, well-scoped fix — nice job!

ainephelan365:
Thank you for the feedback and for the review @kasemAlem ! You've given some great advice that I will definitely carry over to my next task!

gitops/release.go
ainephelan365: Noted the suggestion to add` frontend` to `spec.components` in second test of release_test.go , will carry that pattern forward in future tests.

jencull (approved):
lgtm

PR #1659, opened 2026-07-26
fix: grant bind verb on konflux-integration-runner CRole

kasemAlem:
The scenario controller creates a RoleBinding for the konflux-integration-runner ClusterRole when an IntegrationTestScenario is created. This was failing with a privilege escalation error because Kubernetes prevents granting permissions the creator doesn't already hold.

Added the bind verb scoped to the konflux-integration-runner ClusterRole in the RBAC marker, which is the correct Kubernetes mechanism for this case — it allows creating the RoleBinding without requiring the IS controller manager to hold all the permissions the ClusterRole grants.

Bug discovered while working on [STONEINTG-1722](https://redhat.atlassian.net/browse/STONEINTG-1722)

[STONEINTG-1722]: https://redhat.atlassian.net/browse/STONEINTG-1722?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -106,6 +106,14 @@ rules:
   - watch
 - apiGroups:
   - rbac.authorization.k8s.io
+  resourceNames:
+  - konflux-integration-runner
+  resources:
+  - clusterroles
+  verbs:
+  - bind
+- apiGroups:
+  - rbac.authorization.k8s.io
   resources:
   - rolebindings
   verbs:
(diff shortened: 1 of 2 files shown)


Josh-Everett (approved):

jsztuka (approved):
lgtm

dirgim (approved):

PR #1664, opened 2026-07-30
fix(STONEINTG-1671): use IS-owned annotation and finalizer for nudge processing

kasemAlem:
build-service's removePipelineFinalizer() unconditionally sets build.appstudio.openshift.io/component-nudge-processed="true" when skipping nudging for a NudgeConfig namespace. IS bailed out on any non-empty value of that annotation, so Renovate PipelineRuns were never created when NudgeConfig was present.

Fix:
- Rename NudgeProcessedAnnotation from build.appstudio.openshift.io/component-nudge-processed to test.appstudio.openshift.io/component-nudge-processed so IS owns its marker independently of build-service's annotation (consistent with the existing IS annotation/finalizer domain)
- Add NudgePipelineRunFinalizer (test.appstudio.openshift.io/nudge-pipelinerun) to hold the build PLR open between nudge PLR creation and the annotation write, preventing premature GC on a succeeded PLR; finalizer is added just before CreateNudgePipelineRun and removed after the annotation lands successfully; the already-annotated early-exit path and the deleting-PLR path both clean it up on crash-recovery

Note: PLRs annotated with the old key by IS before this change are not recognized by the idempotency guard during rollout, but any resulting duplicate nudge PLR creation is safe because Renovate is idempotent.

Assisted-By: Claude-Code opus 4.6

### DIFF
--- a/helpers/finalizers.go
+++ b/helpers/finalizers.go
@@ -20,3 +20,7 @@ package helpers
 const IntegrationPipelineRunFinalizer string = "test.appstudio.openshift.io/pipelinerun"
 const IntegrationTestScenarioFinalizer string = "test.appstudio.openshift.io/scenario"
 const ComponentFinalizer string = "test.appstudio.openshift.io/component"
+
+// NudgePipelineRunFinalizer is the finalizer name added to build PipelineRuns while IS is
+// actively creating a nudge PipelineRun, preventing premature GC before nudging completes.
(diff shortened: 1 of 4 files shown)


jsztuka (approved):
lgtm

dirgim (approved):
LGTM

Josh-Everett (approved):

PR #1691, opened 2026-08-31
fix(config): remove invalid excluded_attributes, add central adr_source

msu8:
## Summary

Removes 3 invalid `excluded_attributes` IDs (`test_coverage`, `semantic_naming`, `branch_protection`) that no longer exist in the current `agentready` schema and were causing every assessment run to hard-fail. Also un-excludes `architecture_decisions` and adds `adr_source` pointing at the shared `konflux-ci/architecture` ADR repo, since this repo has no ADR docs of its own — without it, `architecture_decisions` would just score 0 forever.

## Testing

Ran the `agentready` container directly against the updated config (mounting the central ADR repo the same way the CI runner does): no schema error, `architecture_decisions: pass (100)` via the central repo.

**Note on agentready version bump:** 2.31.2 → 2.52.2 - this is required - `adr_source` isn't recognized by the older version and causes a hard failure (Unknown config keys: adr_source). The score/grade will drop because new assessors (adr unrelated) will run and fail. The check still passes and doesn't gate on score.

- [ ] [Controllers diagrams](https://github.com/konflux-ci/integration-service/tree/main/docs) are updated when PR changes controllers code (if applicable)


kasemAlem:
@msu8 , this PR requires a rebase I think

msu8:
@kasemAlem Sure, done.

msu8:
@kasemAlem Hi Kasem, can we try to get this merged?

PR #1714, opened 2026-09-17
feat(STONEINTG-1812): add konflux-ci.dev nudge PLR label with legacy …

kasemAlem:
Context:
 Infra uses Kueue to assign konflux-nudge (highest priority) to nudge PipelineRuns via a CEL
  rule matching the label build.appstudio.redhat.com/type: nudge. As IS migrates labels to the konflux-ci.dev prefix, this label needs renaming — otherwise nudge PLRs would silently lose priority and risk the 1h GitHub App token expiring before the nudge completes.
  
  
  Fix:
    IS now sets both build.konflux-ci.dev/type: nudge (new) and build.appstudio.redhat.com/type: nudge (legacy) on nudge PLRs during the transition. The infra-deployments Kueue CEL rules are updated to accept either label. This decouples the two deployments — no timing coordination required — and makes rollback safe regardless of which side needs reverting. The legacy label and old CEL check will be removed in a follow-up cleanup PR once both sides are stable.
    
    
[STONEINTG-1812](https://redhat.atlassian.net/browse/STONEINTG-1812)

[STONEINTG-1812]: https://redhat.atlassian.net/browse/STONEINTG-1812?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tekton/nudging/nudge_pipeline.go
```diff
@@ -414,7 +414,8 @@ func CreateNudgePipelineRun(ctx context.Context, c client.Client, nudgingPLR *te
 			Name:      name,
 			Namespace: namespace,
 			Labels: map[string]string{
-				tektonconsts.NudgeTypeLabel: tektonconsts.NudgePipelineRunTypeValue,
+				tektonconsts.NudgeTypeLabel:       tektonconsts.NudgePipelineRunTypeValue,
+				tektonconsts.NudgeTypeLabelLegacy: tektonconsts.NudgePipelineRunTypeValue,
```
dirgim, line 418: Do we need to also set the legacy label if we're going to have kueue CEL expression support both labels?

I would expect us to set only the new label, and the build service to set the old label, while kueue supports both until the migration is fully over.

kasemAlem: yep, 
I've removed the legacy one 
thanks for the comment

